Issue #2, opened 2025-03-22
[BUG] [Windows] Setup. Not getting the MCP server to run on Claude Desktop


ivberg:
## Bug Description
I am fairly new to MCP so maybe doing something obviously wrong. But perhaps you could offer debug tips or add better docs.

So far, I have not gotten the MCP server to work on Windows with Claude Desktop.

## Steps To Reproduce
1. Clone this repo and read the docs. 
2. Read some basic MCP docs for Claude Desktop - https://modelcontextprotocol.io/quickstart/user
3. Look at env setup. There seems to be env variables required but it's unclear how to get them to Claude Desktop. I can hack these in say with command line Claude launch but right now saving these for later until I get an error message related to this
4. pip install -e . // Seems successful 
`Successfully built davinci-resolve-mcp
Installing collected packages: pluggy, packaging, iniconfig, colorama, pytest, davinci-resolve-mcp
....
Successfully installed colorama-0.4.6 davinci-resolve-mcp-1.1.0 iniconfig-2.1.0 packaging-24.2 pluggy-1.5.0 pytest-8.3.5`
6. I see no other setup / install instructions so winging it
### Winging it
7. Claud Desktop - File -> Settings -> Developer -> Edit Config (opens folder with claude_desktop_config.json in it) // Not documented
8. See mcp.json in this repo and copy to claude_desktop_config.json
9. Close claude Desktop GUI, but it's clear MCP server was not picked up. On Windows, need to exit with tray icon
10. Claude Desktop restarts and shows error with links to log file
11. Log is something like `python: can't open file '%localappdata%\\AnthropicClaude\\app-0.8.1\\src\\resolve_mcp.py': [Errno 2] No such file or directory`
12. From this, it seems we are missing full path to repo. Edit claude_desktop_config.json and set
`    "davinci-resolve-mcp": {
      "command": "python",
      "args": [
        "S:/src/davinci-resolve-mcp/src/resolve_mcp.py"
      ],
`
13. Fully restart Claude Desktop again. Now error is
`2025-03-22T19:39:28.537Z [davinci-resolve-mcp] [info] Server started and connected successfully
2025-03-22T19:39:28.545Z [davinci-resolve-mcp] [info] Message from client: {"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"claude-ai","version":"0.1.0"}},"jsonrpc":"2.0","id":0}
Traceback (most recent call last):
[...]

PR #4, opened 2025-03-27
Test Windows and fix launcher, config templates, and docs

ivberg:


ivberg:
Together with @samuelgursky initial support - addresses #2

PR #25, opened 2026-04-20
fix: normalize SetCDL arrays to space-separated strings

RubenFelix:
## Summary

`TimelineItem.SetCDL()` in the Resolve scripting API expects string values (`"1.0 1.0 1.0"`) for `Slope`/`Offset`/`Power`, per the [official docs](https://github.com/samuelgursky/davinci-resolve-mcp/blob/main/docs/resolve_scripting_api.txt#L456):

```python
SetCDL({"NodeIndex": "1", "Slope": "0.5 0.4 0.2", ...})
```

Both servers were forwarding the caller's CDL dict unchanged, so any client passing arrays (very natural for MCP clients — JSON arrays are the obvious shape) got silently corrupted grades: `SetCDL` returned `True`, but the resulting node had garbage values in some channels.

Concretely, calling `set_cdl` with `Slope: [1, 1, 1], Offset: [0, 0, 0], Power: [1, 1, 1], Saturation: 1` (a pass-through CDL) produced a heavy red cast on the viewer because the Blue channel on the Gain wheel ended up at `0.01` instead of `1.00`.

## Fix

Adds `_normalize_cdl()` helper in both servers. It converts list/tuple/number values in the CDL dict to the space-separated-string form the API actually expects. Clients can now pass either form — arrays or pre-formatted strings.

- `src/server.py` — compound `timeline_item_color.set_cdl` action
- `src/resolve_mcp_server.py` — granular `timeline_item_set_cdl` tool

No version bump included — leaving that to the maintainer's release cycle.

## Test plan

- [ ] Call `set_cdl` with array form `{"Slope": [1,1,1], "Offset": [0,0,0], "Power": [1,1,1], "Saturation": 1}` — verify Gain/Offset/Gamma/Sat on the node show the expected values in Resolve (not `0.01` residuals)
- [ ] Call `set_cdl` with the documented string form `{"Slope": "1 1 1", ...}` — verify backwards compat still works
- [ ] Apply a non-neutral grade (e.g., `Power: [1.2, 1.2, 1.2]`) and confirm the image updates correctly with no spurious color cast

### DIFF
--- a/src/resolve_mcp_server.py
+++ b/src/resolve_mcp_server.py
@@ -5349,6 +5349,24 @@ def _get_timeline_item(track_type="video", track_index=1, item_index=0):
     return items[item_index], None
 
 
+def _normalize_cdl(cdl):
+    """Resolve's SetCDL expects string values ("1.0 1.0 1.0"), not arrays.
+    Accepts either and converts to the string form the API requires."""
+    if not isinstance(cdl, dict):
+        return cdl
+    out = {}
+    for k, v in cdl.items():
+        if isinstance(v, (list, tuple)):
+            out[k] = " ".join(str(x) for x in v)
+        elif isinstance(v, bool):
+            out[k] = str(v)
+        elif isinstance(v, (int, float)):
+            out[k] = str(v)
+        else:
+            out[k] = v
+    return out
+
+
 # ------------------
 # MediaPool Tools (remaining)
 # ------------------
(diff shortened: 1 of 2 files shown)


samuelgursky:
Thanks for the clean repro and fix direction. I included this behavior in v2.2.0 via commit c109119 rather than merging the PR verbatim, because the server layout changed during the same release work.

CDL normalization now lives in a shared helper and is used by both compound and granular `set_cdl` codepaths. Existing string values still work, and array/tuple/numeric values are now normalized into the space-separated CDL format expected by Resolve, with regression coverage.

samuelgursky:
Closing this PR because the behavior has been implemented in v2.2.0 via c109119 rather than merged verbatim. Thanks again for the contribution; the fix direction was incorporated into the release.

PR #29, opened 2026-04-28
feat: enable fusion_comp to target timeline item compositions

RaincloudTheDragon:
Refactor the `fusion_comp` tool to support targeting Fusion compositions associated with specific timeline items, using `clip_id`, `timeline_item_id`, or a `timeline_item` dictionary.

Previously, `fusion_comp` was restricted to operating on the composition currently active on the Fusion page via `Fusion().GetCurrentComp()`. This enhancement allows for programmatic manipulation of Fusion effects on any clip on the timeline, greatly improving capabilities for automation and bulk operations across a project.

Introduces the `bulk_set_inputs` action, which leverages this new targeting mechanism to apply multiple input changes across various timeline-scoped compositions in a single, efficient operation.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -223,6 +223,112 @@ def _get_item(p):
         return tl, None, _err(f"No item at index {item_index} on {track_type} track {track_index}")
     return tl, items[item_index], None
 
+
+def _find_timeline_item_by_clip_id(tl, clip_id) -> Optional[Any]:
+    """Find a video or subtitle timeline item by GetUniqueId()."""
+    if not clip_id:
+        return None
+    want = str(clip_id)
+    for tt in ("video", "subtitle"):
+        try:
+            n = tl.GetTrackCount(tt)
+        except Exception:
+            continue
+        if not n:
+            continue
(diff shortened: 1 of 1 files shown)


samuelgursky:
Thanks for the feature direction. I implemented this locally in v2.2.0 via commit c109119 rather than merging verbatim, because the branch now includes the v2.2 refactor and installer/stdio hardening work.

`fusion_comp` can now target timeline-item Fusion comps via `clip_id`, `timeline_item_id`, or `timeline_item`, with optional `comp_name` / `comp_index`. Legacy active Fusion page behavior is preserved. I also added `bulk_set_inputs`, but it requires explicit timeline scope for every op so it cannot accidentally mutate the currently open Fusion comp.

This has regression coverage and was live-tested in a disposable Resolve project using a generated Fusion Composition only.

samuelgursky:
Closing this PR because the behavior has been implemented in v2.2.0 via c109119 rather than merged verbatim. Thanks again for the contribution; the feature direction was incorporated into the release.

PR #31, opened 2026-05-01
feat(media_pool): support positioned AppendToTimeline via clip_infos

RaincloudTheDragon:
Expose MediaPool.AppendToTimeline([{clipInfo}, ...]) by accepting
params.clip_infos alongside the existing clip_ids simple-append path.
Each clip info entry resolves clip_id/media_pool_item_id through the
media pool tree and maps snake_case JSON keys to Resolve's camelCase
clipInfo fields (start/end frame, record frame, track index, optional
media type). Positioned returns include timeline_item_id per appended
item for follow-up Fusion ops.

Update SKILL.md with an example for rebuilding subtitle rows after
delete_clips.

Made-with: Cursor
Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # DaVinci Resolve MCP Server
 
-[![Version](https://img.shields.io/badge/version-2.3.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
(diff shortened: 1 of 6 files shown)


samuelgursky (approved):
Approved after live Resolve validation. Verified positioned clip_infos append with synthetic media, confirmed invalid append cases now return errors, and added the 2.3.1 follow-up fix for invalid returned timeline items.

PR #40, opened 2026-05-26
Add Fusion group settings management and bulk expressions

RaincloudTheDragon:
Implements new actions within the `fusion_comp` kernel to provide advanced programmatic control over Fusion GroupOperators and to facilitate batch application of expressions:

- **Group Settings Management:**
    -   `group_settings_export`: Saves `GroupOperator` settings to a `.setting` file and provides a summary of its published inputs.
    -   `group_settings_patch_controls`: Modifies `.setting` files to standardize published input order, crucial for maintaining consistent UI layouts (e.g., thought-bubble templates).
    -   `group_settings_load`: Backs up an existing group's settings and then applies new settings from a `.setting` file, ensuring the group itself is never deleted.
    -   `probe_group_published_inputs`: Enables inspection of both live and file-based published inputs of a group.

- **Bulk Expression Application:**
    -   `bulk_set_expressions`: Offers a robust way to apply multiple expressions to various tool inputs across different timeline-scoped Fusion comps in a single batch operation, including undo/redo support.

- **Guidance and Guardrails:**
    -   Integrates `fusion_commit_hint` and `FUSION_GROUP_GUARDRAILS` to provide essential post-mutation checklists and warnings regarding programmatic Fusion composition edits, especially concerning `InstanceInput` remapping complexities.

A new utility module, `src/utils/fusion_group_settings.py`, was added to encapsulate the parsing, patching, and template rendering logic for `.setting` files. This significantly enhances the ability to automate Fusion workflow and standardize composite elements.


samuelgursky:
Thanks for this @RaincloudTheDragon — the underlying capabilities are genuinely useful and several of them landed nearly verbatim. I'm going to close this PR in favor of [#43](https://github.com/samuelgursky/davinci-resolve-mcp/pull/43), which adapts your work with attribution (Co-Authored-By in the commit), but I want to be transparent about the changes I made and why:

**What I kept ~as-is**
- `bulk_set_expressions` — solid implementation, mirrors `bulk_set_inputs` cleanly. The undo wrap + comp.Lock + per-op error contract are all correct.
- `group_settings_export` — useful. The `SaveSettings` call works exactly as written.
- `probe_group_published_inputs` — clean read-only diagnostic.

**What I changed**

1. **`INSTANCE_INPUT_RE` is broken on nested braces.** Any real-world macro export contains `UserControls = ordered() { Custom = { ... } }` tables, and the flat regex truncates the `InstanceInput` body at the first inner `}`. The patch operation survives this because `_group_inputs_span` uses balanced-brace matching for the outer block — but `parse_setting_file` reports wrong `input_count`, wrong field values, and the patch `diff` is wrong. I replaced the regex with a balanced-brace walker (`_iter_instance_input_blocks` + `_shallow_fields`) and added a nested-`UserControls` fixture to lock the fix in. See [`tests/test_fusion_group_settings.py`](https://github.com/samuelgursky/davinci-resolve-mcp/blob/feat/fusion-group-settings-helpers/tests/test_fusion_group_settings.py).

2. **Dropped the hardcoded thought-bubble template.** `THOUGHT_BUBBLE_SPEECH_ORDER_TEMPLATE` is a very specific 14-input AMZ SpeechBubble layout — useful in your workflow, but it shouldn't be baked into a general-purpose Resolve MCP kernel where nobody else can use it. I renamed the action to `group_settings_splice_inputs(source_path, template_path, dest_path)` and made the template a user-supplied `.setting` file. You can export a known-good group with `group_settings_export` to produce that template — the splice then works for any macro shape.

3. **`group_settings_load` had no undo wrap or comp lock.** Other mutation actions in this kernel (`bulk_set_inputs`, `add_tool`, etc.) all wrap in `comp.StartUndo`/`EndUndo` + `comp.Lock`. Without it, Fusion's Ctrl+Z can't reverse the load. I added the wrap and now return the backup path in error responses too, so callers always have a recovery handle.

4. **Dropped `fusion_commit_hint` as a standalone action**, since `_fusion_group_guardrails_note()` was already injecting the same content into every other action's response (~600 bytes of static advisory per call — no other kernel does this). The constants `FUSION_COMMIT_CHECKLIST` and `FUSION_GROUP_GUARDRAILS` still live in `src/utils/fusion_group_settings.py`, and the actions can mix them into responses on demand with `include_advisory=true`.

The commit on #43 credits you via `Co-Authored-By: Raincloud <raincloudthedragon@gmail.com>`. The core design — exposing `SaveSettings` / `LoadSettings`, batching expressions, a structured published-input probe — is all yours; the parser hardening, template generalization, and undo plumbing are the refinements that let it land. Really appreciate the contribution.

Closing in favor of #43.

RaincloudTheDragon:
> Dropped the hardcoded thought-bubble template.

I didn't mean to commit that, I'm sorry, that was clumsy of me.

PR #60, opened 2026-06-09
Normalize render formats and add proxy UI fallback

### LINKED ISSUE #59: Render helpers query QuickTime codecs with display name instead of format id
## Summary

The render helpers treat the display name returned by `Project.GetRenderFormats()` as the argument for codec/resolution calls. On Resolve, `GetRenderFormats()` can return a mapping like:

```python
{"QuickTime": "mov"}
```

For codec and resolution APIs, Resolve expects the format id/extension (`mov`), not the UI/display label (`QuickTime`). Querying `GetRenderCodecs("QuickTime")` can return an empty dict even though `GetRenderCodecs("mov")` returns the expected QuickTime codecs, including `Apple ProRes 422 LT` / `ProRes422LT`.

## Impact

This breaks native Resolve render planning for QuickTime codecs, especially proxy workflows that need ProRes LT:

```text
format/display: QuickTime
format id:      mov
codec label:    Apple ProRes 422 LT
codec id:       ProRes422LT
```

In the MCP, `probe_render_matrix(formats=["QuickTime"])` reports no codecs, and `prepare_render_job(format="QuickTime", codec="ProRes422LT", ...)` passes `QuickTime` into `SetCurrentRenderFormatAndCodec()` instead of `mov`.

## Proposed Fix

Normalize render format input through the `GetRenderFormats()` mapping before calling:

- `GetRenderCodecs(format)`
- `GetRenderResolutions(format, codec)`
- `SetCurrentRenderFormatAndCodec(format, codec)`
- render job preparation helpers

The fix should accept both `QuickTime` and `mov` as user/tool input while using `mov` for Resolve API calls.

## Validation From Local Fix

I added regression coverage around this behavior and verified:

```bash
python -m pytest tests/test_import.py tests/test_render_deliver_probe.py
```

Result: `21 passed`.

diesdaas:
Fixes #59.

This PR now covers:
- Normalizes Resolve render format display names such as QuickTime to the API format id such as mov before codec probing and job setup, which restores ProRes LT discovery and setup.
- Adds media_pool.generate_proxy_media_ui_capabilities and media_pool.generate_proxy_media_ui as a guarded macOS System Events fallback for Resolve's native Generate Proxy Media command.
- Keeps the documented API boundary explicit: Resolve exposes LinkProxyMedia, not a documented Generate Proxy Media scripting call.
- Makes the proxy UI capability probe usable without a live Resolve connection, so clients can discover the fallback before Resolve/project preflight.
- Cleans up legacy live-test harness collection so ordinary pytest runs skip Resolve-dependent scripts instead of exiting during collection.

Tests:
- python -m pytest
- python -m pytest tests/test_media_pool_ingest_probe.py tests/test_batch_cli.py::BatchCliRunIntegrationTests::test_run_drives_synthetic_job_to_completion tests/test_resolve20_api.py
- python -m py_compile src/server.py tests/test_media_pool_ingest_probe.py tests/test_batch_cli.py tests/test_resolve20_api.py tests/test_all_tools.py tests/test_phase2_skipped.py tests/test_phase3_final.py tests/test_phase4_remaining.py tests/test_phase5_remaining.py
- git diff --check

Current full-suite result: 839 passed, 18 skipped, 8 warnings.

### DIFF
--- a/.claude/skills/davinci-resolve-mcp.md
+++ b/.claude/skills/davinci-resolve-mcp.md
@@ -1,0 +1,21 @@
+---
+name: davinci-resolve-mcp
+description: Work with DaVinci Resolve projects through the local davinci-resolve MCP server, including project inspection, timeline editing, media pool management, Output Blanking UI fallback, offline media diagnosis, and Resolve MCP troubleshooting.
+---
+
+# DaVinci Resolve MCP — Claude Skill Wrapper
+
+This is a Claude-compatible wrapper for the canonical repo skill.
+
+Read and follow:
+
+`skills/davinci-resolve-mcp/SKILL.md`
+
+Use the helper scripts from:
+
+`skills/davinci-resolve-mcp/scripts/`
+
+Do not duplicate local machine paths into this wrapper. The canonical skill uses
+environment variables and repo-relative paths so it can be installed into Codex,
+Claude, OpenCode, OpenClaude, or another MCP client without user-specific path
+leaks.
--- a/docs/SKILL.md
+++ b/docs/SKILL.md
@@ -738,9 +738,18 @@ structure, interchange, comparison, missing-media, and relink-planning helpers:
 - `import_timeline_checked(path, options?, timeline_name?, import_source_clips?, require_temp_path?, dry_run?)`
 - `compare_timelines(right_timeline_id?|right_timeline_index?|left_snapshot?, right_snapshot?)`
 - `probe_interchange_roundtrip(format?, output_dir?, cleanup_imported?)`
-- `detect_missing_media`
-- `build_relink_plan(search_roots)`
+- `detect_missing_media(sanitized?|sanitize_paths?, omit_raw_paths?)`
+- `build_relink_plan(search_roots, max_depth?, max_seconds?, max_files_scanned?, skip_search_when_volume_missing?, sanitized?)`
 - `conform_boundary_report`
+
+For offline media, prefer the sanitized readback first:
+`timeline(action="detect_missing_media", params={"sanitized": true})`. The
+response includes a `diagnosis` block with deduplicated Media Pool items,
+missing volume roots, sample basenames, and a recommended next step. If a source
+volume such as a camera card is not mounted, `build_relink_plan` skips broad
+search by default and reports `skip_reason="missing_source_volume_not_mounted"`;
+mount the volume or pass `skip_search_when_volume_missing=false` only when a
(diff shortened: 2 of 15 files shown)


samuelgursky:
Thanks for the updates and the continued effort here. Status on my end:

**Taken into main** (adapted by hand, since the branch is based on a ~v2.28 tree and doesn't merge against current main): the render format-id normalization (fixes #59), the confirm-token preference-lookup fix you caught, the missing-media diagnosis + bounded relink planning, the thumbnail and vision-shorthand fixes, and an adapted version of your doctor script. Good finds — the format-id bug and the confirm-token bug were both real, and both are now covered by regression tests. All of this just shipped in [v2.37.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.37.0), live-validated against Resolve Studio 21 (raw `GetRenderCodecs("QuickTime")` reproduces the empty result; the normalized path returns 77 codecs including `ProRes422LT`).

**Not taking the UI-automation pieces** (Output Blanking and Generate Proxy Media), and I want to be clear about why, because it's a design decision rather than a quality critique — the guards on the proxy action (dry_run, allow_generate, before/after readback) are genuinely well done.

This server's safety model is a typed, guarded tool surface: confirm tokens, dry-runs, path sandboxing, versioned destructive edits. A UI-automation action is an escape hatch around all of that — an agent that can click menus in Resolve can reach anything in the app with none of those guards in the way, and the Accessibility grant required goes to the whole MCP host process. On top of that, clicks have no transactional semantics (a click "succeeding" doesn't mean the operation did, especially for async jobs like proxy generation), and it breaks in headless/batch contexts. Those constraints don't improve with better engineering, so UI automation is out of scope for the MCP tool surface by design.

For the proxy workflow specifically, there's a fully documented API path to the same outcome: generate the proxies outside the UI — ffmpeg to ProRes LT, or Resolve's own render queue via the render API this PR just fixed — then attach them with `MediaPoolItem.LinkProxyMedia` (which the server already wraps as `media_pool.link_proxy_checked`). You're right that Resolve exposes no Generate Proxy Media scripting call; that's worth a feature request to Blackmagic, but the workflow itself doesn't need to wait on it.

**Also not taking** the `skills/` bundle, `.claude` wrapper, and `openai.yaml` — that's client-specific setup that duplicates `docs/SKILL.md` and pins stale v2.28.0 content.

Closing this PR in favor of the adapted commits. Thanks again — this PR materially improved the project.

PR #61, opened 2026-06-09
Normalize render formats and add proxy UI fallback

### LINKED ISSUE #59: Render helpers query QuickTime codecs with display name instead of format id

diesdaas:
Fixes #59.

This PR covers:
- Normalizes Resolve render format display names such as QuickTime to the API format id such as mov before codec probing and job setup, which restores ProRes LT discovery and setup.
- Adds media_pool.generate_proxy_media_ui_capabilities and media_pool.generate_proxy_media_ui as a guarded macOS System Events fallback for Resolve's native Generate Proxy Media command.
- Keeps the documented API boundary explicit: Resolve exposes LinkProxyMedia, not a documented Generate Proxy Media scripting call.
- Makes the proxy UI capability probe usable without a live Resolve connection, so clients can discover the fallback before Resolve/project preflight.
- Cleans up legacy live-test harness collection so ordinary pytest runs skip Resolve-dependent scripts instead of exiting during collection.

Tests:
- python -m pytest
- python -m pytest tests/test_media_pool_ingest_probe.py tests/test_batch_cli.py::BatchCliRunIntegrationTests::test_run_drives_synthetic_job_to_completion tests/test_resolve20_api.py
- python -m py_compile src/server.py tests/test_media_pool_ingest_probe.py tests/test_batch_cli.py tests/test_resolve20_api.py tests/test_all_tools.py tests/test_phase2_skipped.py tests/test_phase3_final.py tests/test_phase4_remaining.py tests/test_phase5_remaining.py
- git diff --check

Current full-suite result: 839 passed, 18 skipped, 8 warnings.

### DIFF
--- a/.claude/skills/davinci-resolve-mcp.md
+++ b/.claude/skills/davinci-resolve-mcp.md
@@ -1,0 +1,21 @@
+---
+name: davinci-resolve-mcp
+description: Work with DaVinci Resolve projects through the local davinci-resolve MCP server, including project inspection, timeline editing, media pool management, Output Blanking UI fallback, offline media diagnosis, and Resolve MCP troubleshooting.
+---
+
+# DaVinci Resolve MCP — Claude Skill Wrapper
+
+This is a Claude-compatible wrapper for the canonical repo skill.
+
+Read and follow:
+
+`skills/davinci-resolve-mcp/SKILL.md`
+
+Use the helper scripts from:
+
+`skills/davinci-resolve-mcp/scripts/`
+
+Do not duplicate local machine paths into this wrapper. The canonical skill uses
+environment variables and repo-relative paths so it can be installed into Codex,
+Claude, OpenCode, OpenClaude, or another MCP client without user-specific path
+leaks.
--- a/docs/SKILL.md
+++ b/docs/SKILL.md
@@ -738,9 +738,18 @@ structure, interchange, comparison, missing-media, and relink-planning helpers:
 - `import_timeline_checked(path, options?, timeline_name?, import_source_clips?, require_temp_path?, dry_run?)`
 - `compare_timelines(right_timeline_id?|right_timeline_index?|left_snapshot?, right_snapshot?)`
 - `probe_interchange_roundtrip(format?, output_dir?, cleanup_imported?)`
-- `detect_missing_media`
-- `build_relink_plan(search_roots)`
+- `detect_missing_media(sanitized?|sanitize_paths?, omit_raw_paths?)`
+- `build_relink_plan(search_roots, max_depth?, max_seconds?, max_files_scanned?, skip_search_when_volume_missing?, sanitized?)`
 - `conform_boundary_report`
+
+For offline media, prefer the sanitized readback first:
+`timeline(action="detect_missing_media", params={"sanitized": true})`. The
+response includes a `diagnosis` block with deduplicated Media Pool items,
+missing volume roots, sample basenames, and a recommended next step. If a source
+volume such as a camera card is not mounted, `build_relink_plan` skips broad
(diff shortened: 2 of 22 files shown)


samuelgursky:
Thanks for the follow-up. The new commit here (`67a70c7b`) contained genuinely useful test hygiene, and it's now on main **with your authorship preserved** — 10de676 is committed under your name (adapted by hand since the branch doesn't merge against current main):

- the `_skip_or_exit` preflight so the legacy live harnesses (`test_all_tools`, `test_phase2`–`5`) skip under pytest instead of dying at collection, plus your missing-project/timeline/clip guards
- the `test_resolve20_api.py` helper rename so pytest stops collecting `test()` as a test function
- the hermetic transcription-capability stub in the batch CLI integration test

We extended your skip helper in a follow-up commit (0757ab3) to also raise `unittest.SkipTest` when imported outside pytest — this repo's canonical runner is unittest, where the modules previously still exited at import. With both pieces the offline suite is now fully clean under both runners for the first time (unittest: 1019 tests, 0 errors, 5 clean skips; pytest: the legacy modules skip at collection). Standalone `python3 tests/test_*.py` behavior is unchanged.

The remainder of this PR is the same material as #60: the render format-id fix, confirm-token fix, missing-media diagnosis, and doctor script already shipped in [v2.37.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.37.0), and the UI-automation actions (Output Blanking, Generate Proxy Media) and the `skills/` bundle remain declined for the reasons in the [#60 close comment](https://github.com/samuelgursky/davinci-resolve-mcp/pull/60#issuecomment-4664570881) — in short, UI automation is out of scope for the MCP tool surface by design (it bypasses the server's guard model and needs host-wide Accessibility grants), and the proxy workflow has a documented API path via ffmpeg / the render queue + `LinkProxyMedia`. Reopening the same branch won't change that decision, so I'm closing this one too.

If you contribute again — and I hope you do, your bug reports and test fixes have been consistently on point — please rebase on current main and scope each PR to one concern. Small, focused PRs from an up-to-date branch can be merged directly with your commits intact, no adaptation needed.

PR #69, opened 2026-06-17
feat: added option to add excluded folders and change inventory limit

rgxdev:
image

Added a possibility to **exclude custom folders** from being indexed during inventory walk, **increased maximum clips** to 10000 possible and added option to set a custom value for maximum clips being indexed(default is 500)

### DIFF
--- a/src/analysis_dashboard.py
+++ b/src/analysis_dashboard.py
@@ -4790,6 +4790,12 @@ HTML = r"""<!doctype html>
               <button type="button" class="secondary path-browse" data-browse-target="prefPreferredGeneratedMediaFolder">Browse…</button>
             </div>
           </label>
+          <label>Inventory limit
+            <input id="prefInventoryLimit" type="number" min="1" max="10000" placeholder="Default 500">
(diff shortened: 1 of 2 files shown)


samuelgursky:
Thanks @rgxdev! Adapted and shipped in v2.53.0.

Both controls landed: `media_analysis.inventory_exclude_bins` (comma-separated folder names skipped recursively during the inventory walk) and `media_analysis.inventory_limit` (configurable, ceiling raised 2000 → 10000), wired through `setup` and the control panel.

One change from the PR as submitted: the default now excludes **nothing** rather than defaulting to excluding an `assets` bin. Defaulting to exclusion would have silently stopped indexing existing `assets` folders for everyone on upgrade — so exclusion is now strictly opt-in. Added unit coverage for the exclusion walk and the preference round-trip.

Closing since it landed on main as an adaptation rather than a direct merge. Credited in the changelog and release notes.

Release: https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.53.0

PR #78, opened 2026-07-02
fix: initialize Windows Resolve scripting env before bridge import

POLEPALLIANVESH:
## Summary
Fixes a Windows startup crash in network transports by initializing the Resolve/Python runtime environment before importing `DaVinciResolveScript`.

## Cause
On Windows, importing the Resolve bridge too early can trigger a native `fusionscript` access violation before server bind.

## Change
Before bridge import on Windows, set `PYTHONHOME`, add Resolve install dir to `PATH`, and call `os.add_dll_directory()`.

## Validation
`python -m py_compile src/server.py` passes, and `--transport streamable-http` stays running without pre-bind crash.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -526,6 +526,32 @@ resolve = None
 dvr_script = None
 _resolve_lock = threading.RLock()
 
+# On Windows the fusionscript native bridge DLL must be locatable before the
+# Python import machinery attempts to load it.  Setting PYTHONHOME, prepending
+# the Resolve install directory to PATH, and registering it with
+# os.add_dll_directory() ensures the dynamic loader can find fusionscript.dll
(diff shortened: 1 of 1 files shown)


samuelgursky:
Thanks @POLEPALLIANVESH! This branch was based on v2.57.5 and the current build is v2.59.0, so rather than merge the branch I integrated your change directly onto `main` with you credited as co-author.

Landed in 277ccf6 (`fix: initialize Windows Resolve scripting env before bridge import (#78)`), verbatim: PYTHONHOME + PATH prepend + `os.add_dll_directory()` ahead of the bridge import, fully guarded by `sys.platform.startswith('win')` so it's a no-op on macOS/Linux. `py_compile` clean; full offline suite (1343 tests) green. Closing as integrated — appreciate the fix.

Issue #74, opened 2026-06-19
[Feature Request] Support for Source Track Selector control


swayll:
**Description:**

## Problem Statement

As example, when agent adding Fusion Text+ clips from templates (via the Titles category in the Effects Library), they are currently inserted only onto the video track specified in the **Source Track Selector** patching panel. There is no programmatic way via the DaVinci Resolve API—and consequently no tool in the current MCP server—to change which track the Text+ clip lands on when inserted programmatically.

|img1|img2|img3|
|---|---|---|
|<img width="234" height="206" alt="Image" src="https://github.com/user-attachments/assets/68b6adb4-5812-46d7-bce2-46cb0be16a9e" />|<img width="686" height="402" alt="Image" src="https://github.com/user-attachments/assets/8014dae5-8530-4000-986f-9493f1dbdfd1" />|<img width="760" height="388" alt="Image" src="https://github.com/user-attachments/assets/346b1103-88a3-40a6-9ed9-bde9e06ac959" />|

## Expected Behavior

I want to be able to control the destination track. Specifically:
- Query the current Source Track Selector target track
- Set the Source Track Selector to target a specific video track (e.g., V2, V3)

## Current Workarounds

Currently, users must manually:
1. Click the Source Track Selector in the Edit page UI to change the target track before inserting 
2. Insert the Text+ clip manually from the Effects Library > Titles category 

This breaks automation workflows where AI assistants (via MCP) need to place titles on specific tracks.

## Questions

1. **Does the DaVinci Resolve Scripting API expose any method to get or set the Source Track Selector destination?**

2. **If the API does support this functionality**, I propose adding a new tool to the MCP server—perhaps `set_source_track_selector` or `set_insert_target_track`—that would allow AI assistants to control the destination track before inserting Text+ clips or other generators.

## Proposed Solution

If the Resolve API supports this, add to the MCP `ResolveAPI` class:

```python
def set_source_track_selector(self, track_index: int) -> bool:
    """Set the Source Track Selector to target the specified video track.
    
    Args:
        track_index: 1-based index of the video track (1 = V1, 2 = V2, etc.)
    
    Returns:
        bool: True if successful, False otherwise.
    """
    # Implementation using Resolve API
    pass

def get_source_track_selector(self) -> int:
    """Get the currently selected video track index for the Source Track Selector.
    
    Returns:
        int: 1-based track index, or -1 if query fails.
    """
    pass
```

## Additional Context

[...]

PR #80, opened 2026-07-03
api-truth: record AppendToTimeline clipInfo endFrame exclusive-bound semantics

chenyuxiaojin:
## What

Adds one entry to the `api_truth` ledger: **`MediaPool.AppendToTimeline` clipInfo `endFrame` is an exclusive bound** — an appended item's duration is `endFrame - startFrame` frames, not `end - start + 1`. Internal quirk entry (no `submit` key), so `docs/reference/api-limitations.md` is unaffected.

## How it was verified

Live readback probe on Resolve Studio 21.0 (macOS, davinci-resolve-mcp v2.51+), while batch-assembling a second video layer mirrored from an `execute_tighten` plan's keep_ranges:

- Appending `{startFrame: 6254, endFrame: 6348, recordFrame: 0, mediaType: 1}` yields an item of **94 frames** (`GetDuration()` readback), matching `end - start` exactly.
- 130 consecutive appends whose absolute `recordFrame`s advance by exactly `(endFrame - startFrame)` assemble a **gap-free** track (every item's record end == next item's record start), and the track's total length equals `Σ(end - start)`.
- `_build_append_clip_info_dict` (server.py) passes `startFrame`/`endFrame` through verbatim, so the probe measures the raw API bound, not an MCP-side adjustment.

I queried `resolve_control.api_truth` for this before probing and got 0 facts — exactly the situation the ledger exists for, hence this PR.

## Why it matters

Two in-repo call sites currently assume the inclusive bound (filed separately as an issue with measurements): a record cursor advanced by `end - start + 1` leaves a 1-frame hole between consecutive clips, and a keep-range converted with `end - 1` shaves the last frame of every kept segment.

## Test

`tests/test_api_truth.py::test_append_endframe_exclusive_bound_recorded` — lookup regression in the style of the issue #74 entry test. `python -m unittest tests.test_api_truth tests.test_api_truth_mitigations`: 11 tests, all pass.

### DIFF
--- a/src/utils/api_truth.py
+++ b/src/utils/api_truth.py
@@ -487,6 +487,30 @@ API_TRUTH: List[Dict[str, Any]] = [
         "issue": 77,
         "mitigation": ["_verify_clip_property_writeback", "_verify_writeback"],
     },
+    {
+        "symbol": "MediaPool.AppendToTimeline clipInfo endFrame (exclusive bound)",
+        "object": "MediaPool",
+        "signature": "([{mediaPoolItem, startFrame, endFrame, recordFrame, "
+                     "trackIndex, mediaType}]) -> [TimelineItem]",
+        "reality": "clipInfo endFrame is an EXCLUSIVE bound: the appended item's "
+                   "duration is endFrame - startFrame frames, not "
+                   "endFrame - startFrame + 1. Verified by readback probe on "
+                   "Resolve Studio 21.0 — appending {startFrame: 6254, "
+                   "endFrame: 6348} yields a 94-frame item, and 130 such "
+                   "appends whose absolute recordFrames advance by exactly "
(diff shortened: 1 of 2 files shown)


samuelgursky:
Thanks @chenyuxiaojin! Integrated onto `main` with you as co-author (the branch forked at v2.57.5; v2.58.0's grading catalog added ledger entries near the same spot, so I placed your entry at the end of `API_TRUTH` rather than merging the conflicting hunk).

Landed in 50cfe3d (`docs(api_truth): record AppendToTimeline endFrame exclusive-bound semantics (#80)`), with your lookup regression test. No `submit` key, so `api-limitations.md` is unaffected (`gen_api_limitations.py --check` clean). This is now the documented basis for the #82 fix. Closing as integrated.

PR #81, opened 2026-07-03
edit_engine: dedupe identical plan_tighten skipped rows into one entry with a count

chenyuxiaojin:
## What

`plan_tighten` appends one `{item, reason}` row to `skipped` per timeline item. When a whole layer's source has no analysis, that repeats the identical row once per segment.

Real session that motivated this: a 25-minute dual-layer talking-head timeline, each layer cut into 87 segments, second layer (screen recording) intentionally unanalyzed → the plan payload and the MCP response each carried **87 identical** `{"item": "recording_….mp4", "reason": "no analysis for source media (db_ingest or analyze first)"}` rows. Zero added signal, meaningful added tokens for MCP clients.

## Change

`_dedupe_skipped()` collapses identical `(item, reason)` rows into a single entry with a `count` field, preserving first-occurrence order. Applied once at the end of the scan loop, so both the `No dead-air lifts found` early return and the saved/echoed plan use the deduped list.

Backward compatibility: the list shape and the `item`/`reason` keys are unchanged; existing consumers only see fewer duplicate rows plus a new `count` key (`count: 1` for singletons).

## Test

`tests/test_edit_engine.py::test_skipped_rows_deduped_with_count` — one unanalyzed source cut into 3 segments reports a single row with `count: 3`. Full module: `python -m unittest tests.test_edit_engine` → 31 tests, all pass (existing `test_unanalyzed_media_skipped_with_reason` unchanged and green).

### DIFF
--- a/src/utils/edit_engine.py
+++ b/src/utils/edit_engine.py
@@ -339,6 +339,29 @@ def _gaps_in_range(
     return gaps
 
 
+def _dedupe_skipped(skipped: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """Collapse identical (item, reason) skip rows into one entry with a count.
+
+    A timeline layer cut into many segments from one unanalyzed source clip
+    otherwise reports the same skip once per segment (87 identical rows in a
+    real two-layer session), which bloats the plan payload without adding
+    signal. First occurrence order is preserved.
+    """
+    out: List[Dict[str, Any]] = []
+    index: Dict[Tuple[Any, Any], Dict[str, Any]] = {}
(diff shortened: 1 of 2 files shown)


samuelgursky:
Thanks @chenyuxiaojin! Integrated onto `main` with you as co-author (rebased past the v2.57.5→v2.59.0 drift; applied cleanly).

Landed in 7ccf751 (`feat(edit_engine): dedupe identical plan_tighten skipped rows with a count (#81)`), including your `test_skipped_rows_deduped_with_count`. Full offline suite green. Closing as integrated.

PR #85, opened 2026-07-10
docs: add Autohand Code MCP setup

igorcosta:
## Summary

Add an Autohand Code command to the installation guide after the supported-client table. It reuses the package's managed `server` launcher, so users keep the existing setup and Resolve path detection flow.

## Verification

- `node bin/davinci-resolve-mcp.mjs --help` confirms the documented `server` launcher
- `git diff --check`


samuelgursky:
Thanks for the PR! I pushed a small fix (05510ab) before merging: `autohand mcp add` rejects the bare `-y` flag — commander errors with `unknown option '-y'` since the `add` subcommand doesn't allow unknown options. Autohand's npx normalization inserts `-y` into the stored config automatically, so the documented command now omits it. Also moved the block below the multi-client sentence so that sentence stays attached to the installer table it describes.

PR #87, opened 2026-07-13
Timeline tooling fixes, transport-wedge fix, and background jobs for long ops

### LINKED ISSUE #88: Server wedges on long ops; several timeline-tool ergonomics & correctness issues
While driving the MCP against DaVinci Resolve Studio I ran into a cluster of issues. Grouping them here for context; a proposed fix is in #87.

**Server stability**
- Long operations (transcription, subtitle generation, scene-cut/Dolby analysis, timeline export/import) run synchronously on the event loop, so a multi-minute call freezes the whole server — including the stdio transport — until it returns, and the client has to be restarted to recover.
- Those same long operations also exceed the client's tool-window timeout, so the call is reported as failed even though the work completes in Resolve.

**`timeline.create_variant_from_ranges`**
- The returned `items` echo the requested frame numbers rather than where clips were actually placed, which hides gaps/shrink (e.g. when source and timeline frame rates differ).
- `dry_run` returns before resolving clip ids / validating frames, so a dry run can succeed while the real call fails.
- No opt-in way to butt ranges together gap-free; the assembled timeline is silently video-only unless audio ranges are supplied; the help advertises a `source_clip_id` key and timeline-frame numbers that don't match what the code accepts (media-pool id + source frames).

**Discoverability / error quality**
- `timeline set_current` only accepts the 1-based index (which shifts as timelines are added/archived), not a stable id/name.
- `timeline get_items` raises a bare Python `KeyError` on a missing param and has no `action_help` entry.
- `timeline apply_cuts` silently drops non-applicable cuts with no reason.
- `action_help()` lists only the documented subset of actions, and returns `HELP_NOT_REGISTERED` for a non-existent action (indistinguishable from a valid-but-undocumented one).

Proposed fix: #87 — opt-in options with back-compatible defaults, offline unit tests plus live validation against Resolve 21.

lukeashford:
Closes #88

A set of fixes and small features found while driving the MCP against DaVinci Resolve Studio. Every new option is opt-in, so default behavior is unchanged.

## Server robustness

- **Transport wedge (fix).** Synchronous tool bodies ran inline on the single asyncio event-loop thread, so a blocking Resolve call, subprocess, or the up-to-60s launch wait froze the whole server — including the stdio read loop — until it returned, requiring a client restart. Tool bodies now run in a worker thread (`anyio.to_thread.run_sync`), serialized on a bridge lock so the single-threaded scripting bridge is never entered concurrently. Best-effort and version-coupled to the SDK's tool shape; degrades to inline behavior if that shape changes.
- **Background jobs for long ops.** Transcription, subtitle generation, scene-cut/Dolby analysis, and timeline export/import can exceed the client's tool-window timeout. Passing `background=true` now returns a `job_id` immediately; poll with `resolve_control(action="job_status", params={"job_id": ...})`. Generic registry in `src/utils/background_jobs.py`; every long op routes through it.

## `timeline.create_variant_from_ranges`

- Reports **actual placement** (`items[].placed`, in both frame spaces) alongside the requested range, plus a `placement_mismatches` count — the response no longer echoes requested numbers as if they were results.
- `dry_run` now resolves clip ids and validates frame ranges, so a passing dry run matches what the real call accepts (and a bad id no longer leaves an orphan timeline).
- `pack=true` butts clips together at the end of each track — gap-free regardless of a source/timeline frame-rate mismatch.
- Reports audio presence and warns when a range list yields a video-only (silent) timeline.
- Corrected help: `clip_id` is a media-pool item id (not a timeline-item id); `start_frame`/`end_frame` are source frames, `end_frame` exclusive.

## Discoverability & error quality

- `timeline set_current` accepts a stable `id`/`name`, not only the shifting 1-based index.
- `timeline get_items` validates its track selector (structured error instead of a bare `KeyError`) and gained an `action_help` entry.
- `timeline apply_cuts` reports `skipped` cuts with reasons instead of silently dropping them.
- `action_help()` returns the full valid-action list and distinguishes an unknown action (`UNKNOWN_ACTION`) from a valid-but-undocumented one (`HELP_NOT_REGISTERED`).
- Docs: clarified `begin_run`/`end_run` batching, the timeline-subtitle vs clip-level transcript distinction, and the source/timeline frame convention.

## Testing

70 new offline unit tests (`tests/test_*.py`, stub-based, no Resolve required); the full offline suite passes. (Two pre-existing loader failures — `test_custom_timeline`, `test_improvements` — are unrelated `import requests` cases, untouched here.) The transport offload, background jobs, and `create_variant_from_ranges` placement/pack behavior were additionally validated live against Resolve Studio 21.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -18670,6 +18670,28 @@ def edit_engine(action: str, params: Optional[Dict[str, Any]] = None) -> Dict[st
     ])
 
 
+_TIMELINE_ACTIONS = [
+    "list", "get_current", "set_current", "get_name", "set_name", "get_start_frame",
+    "get_end_frame", "get_start_timecode", "set_start_timecode", "get_track_count",
+    "add_track", "delete_track", "get_track_sub_type", "set_track_enable",
+    "get_track_enabled", "set_track_lock", "get_track_locked", "get_track_name",
+    "set_track_name", "get_items", "delete_clips", "set_clips_linked", "duplicate",
+    "duplicate_clips", "copy_clips", "move_clips", "copy_range", "duplicate_range",
+    "overwrite_range", "lift_range", "story_spine_report", "create_variant_from_ranges",
+    "bulk_set_item_properties", "apply_look_to_items", "thumbnail_contact_sheet",
+    "marker_thumbnail_review", "edit_kernel_capabilities", "probe_edit_kernel_item",
+    "title_property_scan", "set_title_text", "bulk_set_title_text", "create_compound_clip",
+    "create_fusion_clip", "import_into_timeline", "export", "get_setting", "set_setting",
+    "insert_generator", "insert_fusion_generator", "insert_fusion_composition",
+    "insert_ofx_generator", "insert_title", "insert_fusion_title", "get_unique_id",
+    "get_node_graph", "get_media_pool_item", "get_transcript", "propose_cuts", "apply_cuts",
+    "get_mark_in_out", "set_mark_in_out", "clear_mark_in_out", "convert_to_stereo",
+    "get_items_in_track", "get_voice_isolation_state", "set_voice_isolation_state",
+    "extract_source_frame_ranges", "audio_mix_capability_report", "clip_where", "action_help",
+    *_TIMELINE_CONFORM_KERNEL_ACTIONS, *_TIMELINE_AUDIO_KERNEL_ACTIONS,
+]
+
+
 @mcp.tool()
 @_destructive_op("timeline")
 def timeline(action: str, params: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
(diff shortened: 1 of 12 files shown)


samuelgursky:
Thank you for this — excellent work, and the issue write-up in #88 made the review straightforward. I reviewed the full diff, verified the threaded-dispatch SDK coupling against the pinned mcp SDK (1.27) over a real stdio session, and ran the full offline suite (1,483 green).

Per this repo's integration workflow the changes were landed manually on `main` rather than merged, preserving your 10 commits' structure and messages, each with `Co-Authored-By: lukeashford <luke@lukeashford.com>` so the work is attributed to you (b93ab18…6243b45). Released as [v2.62.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.62.0).

The only change from your branch: the dispatch docstring's SDK version note now reads "verified on mcp 1.27" (this repo pins 1.27.0, and the Tool shape is present there).

Closing since the work is on main. Thanks again!

PR #90, opened 2026-07-17
fix(graph): SetLUT fails for LUTs installed by the dctl tool (user LUT dir not resolved)

Mldphotohraphie:
## Problem

`graph set_lut` always returns `{"success": false}` for any LUT or DCTL that was installed via the `dctl` tool, and `get_lut` confirms the node stays empty. This makes it impossible to apply a LUT/DCTL to a node through the API even though `SetCDL` on the same node graph works fine.

## Root cause

`dctl install` (and the `dctl_dir` in `utils/platform.py`) writes LUTs/DCTLs to the **per-user** LUT directory:

- macOS: `~/Library/Application Support/Blackmagic Design/DaVinci Resolve/LUT`

But Resolve's `Graph.SetLUT()` resolves LUT names — and even absolute paths — **only against the master (system) LUT directory**:

- macOS: `/Library/Application Support/Blackmagic Design/DaVinci Resolve/LUT`

So a file the `dctl` tool just installed can never be found by `SetLUT`.

### Verified live (DaVinci Resolve Studio 21.0.2)

```
SetLUT(1, "Foo.cube")               -> True    # Foo.cube in master LUT dir
SetLUT(1, "Foo.cube")               -> False   # same file only in user LUT dir
SetLUT(1, "/abs/user/dir/Foo.cube") -> False   # absolute path to user dir
```

Reproduced with both `.cube` and `.dctl`, relative and absolute paths, node indices 0 and 1.

## Fix

When `SetLUT` fails, locate the LUT (absolute path / user LUT dir / master dir), copy it into the master LUT dir, call `RefreshLUTList()`, and retry with the basename. No behavior change when `SetLUT` already succeeds.

Adds two helpers next to the existing DCTL helpers: `_master_lut_dir()` (per-platform master LUT root) and `_ensure_lut_in_master()`.

### After the fix (same machine, end to end)

```
BEFORE (user dir)  SetLUT = False
AFTER  (relocated) SetLUT = True    GetLUT = 'Foo.cube'
```

## Notes / scope

- Master-dir paths used: macOS `/Library/Application Support/Blackmagic Design/DaVinci Resolve/LUT`; Windows `%PROGRAMDATA%\Blackmagic Design\DaVinci Resolve\Support\LUT`; Linux `/opt/resolve/LUT` (falls back gracefully). Only macOS was verified live.
- Copying into the master dir may require write access to a system directory; on failure the helper returns `None` and the original `False` is surfaced unchanged.
- The granular `graph_set_lut` in `src/granular/graph.py` has the same limitation and could get the same treatment in a follow-up.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -22070,7 +22070,26 @@ def graph(action: str, params: Optional[Dict[str, Any]] = None) -> Dict[str, Any
     elif action == "get_lut":
         return {"lut": g.GetLUT(p["node_index"])}
     elif action == "set_lut":
-        return {"success": bool(g.SetLUT(p["node_index"], p["lut_path"]))}
+        node_index = p["node_index"]
+        lut_path = p["lut_path"]
+        ok = bool(g.SetLUT(node_index, lut_path))
+        if not ok:
+            # Resolve resolves LUTs for SetLUT only against the master LUT dir,
+            # not the per-user dir that dctl install writes to. Relocate and retry.
+            relocated = _ensure_lut_in_master(lut_path)
+            if relocated:
+                try:
+                    _, proj, _lut_err = _check()
+                    if proj:
+                        proj.RefreshLUTList()
+                except Exception:
+                    pass
+                ok = bool(g.SetLUT(node_index, relocated))
+                if ok:
+                    return {"success": True, "resolved_lut": relocated,
+                            "note": "LUT copied into the master LUT dir; "
+                                    "SetLUT does not resolve the user LUT dir."}
+        return {"success": ok}
     elif action == "get_node_cache":
         return {"cache": g.GetNodeCacheMode(p["node_index"])}
     elif action == "set_node_cache":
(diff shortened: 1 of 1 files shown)


samuelgursky:
Tested this live on DaVinci Resolve Studio 21 (a real 1-node graph, identity `.cube` placed only in the user LUT dir). **The core diagnosis holds up** — and importantly, `RefreshLUTList()` alone does *not* fix it:

| # | Path passed to `SetLUT` | `RefreshLUTList` first? | File in… | Result |
|---|---|---|---|---|
| 1 | basename | no | user dir | `False` |
| 2 | absolute user-dir path | no | user dir | `False` |
| 3 | basename | **yes** | user dir | **`False`** |
| 4 | absolute user-dir path | **yes** | user dir | **`False`** |
| — | stock LUT basename (control) | — | master dir | `True` |
| 5 | basename | yes | **master dir** | **`True`** (`GetLUT` confirms) |

Trials 3–4 rule out "just refresh the LUT list" as an alternative — a user-dir LUT stays unresolvable even by absolute path after a refresh. So relocating into the master dir really is required. Nice catch.

**One change I'd ask for before merge — namespace the copy to avoid clobbering stock LUTs.** As written, `_ensure_lut_in_master` flattens to `os.path.basename` and `shutil.copy2`s over the master root, so a user LUT named e.g. `InstantC.cube` would silently overwrite the stock one machine-wide. I tested that a **subfolder-relative** path resolves fine:

| # | Path passed to `SetLUT` | File in… | Result |
|---|---|---|---|
| 6 | `MCP/MCPTest.cube` | master/**MCP**/ | **`True`** (`GetLUT` confirms) |

So copy into `…/LUT/MCP/<basename>` and hand `SetLUT` the relative `MCP/<basename>` instead of a bare basename in the master root. Same mechanism, no collision risk.

Other items:

- **Windows path bug:** `os.environ.get("PROGRAMDATA", r"C:\ProgramData")` — inside a raw string that fallback is a literal `C:\\ProgramData` (double backslash). Drop the `r` or the doubling. (Only bites when `PROGRAMDATA` is unset, but it's wrong.)
- **Tests:** this repo has a large offline suite and this change adds none. Please cover `_master_lut_dir()` per-platform and `_ensure_lut_in_master()` (found-in-user-dir → copied; already-in-master → no-op; missing source → `None`; unwritable master → `None`).
- **`src/granular/graph.py`** has the same limitation (you noted this) — worth the same treatment so the two paths don't diverge.

Non-blocking: writing into `/Library/Application Support/...` can require admin; the helper degrades correctly (returns `None`, original `False` surfaces). Also worth considering the same relocate at `dctl install` time so the LUT is reachable before `set_lut` is ever called — but the on-failure retry is a reasonable safety net on its own.

samuelgursky:
Merged and shipped in **v2.62.3** — thanks @Mldphotohraphie! I extended it slightly on the way in:

- **Namespaced the relocation** into an `MCP/` subfolder of the master LUT dir (SetLUT resolves subfolder-relative paths, verified live) so it can't clobber a stock/vendor LUT sharing a basename.
- **Windows fallback fix**: `r"C:\\ProgramData"` is a literal double-backslash path → `r"C:\ProgramData"`.
- **Granular parity**: same treatment for `graph_set_lut` in `src/granular/graph.py`, sharing one `src/utils/lut_paths.py` helper so the two entrypoints don't diverge.
- **Tests**: offline coverage in `tests/test_lut_paths.py` (relocation, absolute path, no-clobber, no-op when already staged, missing source, unwritable master).

Release: https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.62.3

samuelgursky:
Correction to my earlier comments on this PR: my live verification ran on **DaVinci Resolve Studio 19.1.3.7**, not Studio 21. I carried the 21.0.2 figure over from your report and restated it as the version under test on my side — that was my error.

The findings themselves are unchanged and still hold on 19.1.3.7:
- user-dir LUT → `SetLUT` false, by basename and by absolute path
- still false after `RefreshLUTList()`
- master dir → true; master `MCP/` subfolder → true (`GetLUT` confirms)

So the behavior you found on 21.0.2 reproduces on 19.1.3.7 as well, which makes it look version-independent rather than a 21.x regression. Code comments, CHANGELOG, and the v2.62.3 release notes have been corrected to attribute each observation to the version it was actually made on.

PR #91, opened 2026-07-22
Add waveform silence ripple (Resolve Ripple Delete Silence parity)

EliteSystemsAI:
## Summary
- Adds `plan_silence_ripple` / `execute_silence_ripple` to the edit engine — waveform dead-air removal via ffmpeg `silencedetect`, mirroring Resolve's **Clip → Audio Operations → Ripple Delete Silence** dialog.
- Defaults match common Resolve settings: **−30 dB** threshold, **10-frame** minimum strip, **0** pre-head, **1** post-tail frame. Tunable via params.
- Execution assembles a **variant timeline** from keep ranges (same safety model as `execute_tighten`: original untouched, confirm-gated, audio mirroring).

## Motivation
Resolve exposes Ripple Delete Silence in the UI but not via scripting. Transcript-based `plan_tighten` targets a different signal (speech gaps) and can cut mid-word when transcripts disagree with waveform pauses. This gives agents a safe, waveform-driven alternative.

## Test plan
- [x] `python -m unittest tests.test_silence_ripple tests.test_edit_engine.PlanSilenceRippleTests`
- [ ] Live: `plan_silence_ripple` on a talking-head A-roll track; compare duration/lifts to manual Ripple Delete Silence
- [ ] Live: `execute_silence_ripple` → verify variant is shorter, audible, no mid-word cuts

## Notes for adopters
- Requires readable **file paths** on timeline items (items without paths are skipped, not silently trimmed).
- Tune `threshold_db` / `min_strip_frames` per voice room tone — same as the Resolve dialog.
- Slides/overlays on other tracks are **not** re-timed; run silence ripple on the base A-roll first, then re-place B-roll.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 Release history for the DaVinci Resolve MCP Server. The latest release is summarized in the root README; older entries live here to keep the README focused.
 
+## Unreleased
+
+### Added
+
+- **`plan_silence_ripple` / `execute_silence_ripple`** — waveform-driven dead-air
+  removal that mirrors Resolve's *Ripple Delete Silence* dialog (ffmpeg
+  `silencedetect` → keep-range variant assembly). Tunable `threshold_db`,
+  `min_strip_frames`, `pre_head_frames`, and `post_tail_frames`. Original
+  timeline is never mutated; execution is confirm-gated like `execute_tighten`.
+  New module: `src/utils/silence_ripple.py`; tests in
+  `tests/test_silence_ripple.py` and `tests/test_edit_engine.py`.
+
 ## What's New in v2.62.3
 
 A single grading fix (PR #90, by @Mldphotohraphie), extended and hardened. No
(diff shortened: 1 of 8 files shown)


samuelgursky:
Thanks @EliteSystemsAI — great contribution. This filled a real gap (Ripple Delete Silence has no scripting equivalent), and the implementation followed the edit-engine house patterns closely: half-open keep ranges, confirm gating, variant safety model, audio mirroring. Merged as-is with your authorship preserved, then hardened in two follow-up commits on main:

- Items without a readable media path now ride along whole in the variant instead of being omitted (only items with no media reference at all are dropped, and `skipped` says so explicitly).
- Live validation on production MXF caught a subtle one: those masters carry one mono PCM stream per audio channel, and ffmpeg's default stream selection can land on a dead scratch channel — reading entire takes as silence. Detection now merges all audio streams first (silent only when every channel is, matching Resolve's dialog semantics), and runs with `-vn` so it no longer decodes 4K picture just to scan audio.
- Clearer failure modes: missing ffmpeg reports itself, and a plan whose keep ranges all come up empty warns and is refused at execution.

Live-validated end to end in Resolve Studio (18 lifts at −50 dB, variant assembled with exact video/audio frame-lock across 5 mirrored audio tracks) and shipped in [v2.63.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.63.0), now on npm. Thanks again!

EliteSystemsAI:
Very welcome good sir thanks for putting this together! 🫡

PR #94, opened 2026-07-24
Support Resolve network scripting connections

double2tea:
## Summary

- add a shared Resolve connection helper that supports explicit Network scripting hosts and bounded timeouts
- route compound, granular, installer-generated, and doctor connections through the same behavior
- document Network mode security considerations and add regression coverage for Local and Network configurations

## Root cause

Resolve's one-argument `scriptapp("Resolve")` discovery is appropriate for Local scripting mode, but Network mode requires the explicit host overload. Without a configurable host, the MCP can return no Resolve object or wait on discovery even when Resolve is running on the same machine.

## Behavior

Local mode remains unchanged when `RESOLVE_SCRIPT_HOST` is absent. Network mode is opt-in through `RESOLVE_SCRIPT_HOST`, with `RESOLVE_SCRIPT_TIMEOUT` accepting a positive finite timeout in seconds and defaulting to five seconds.

## Validation

- `venv/bin/python -m unittest tests.test_resolve_connection tests.test_cdl_and_install_config tests.test_static_undefined_names tests.test_action_list_drift tests.test_panel_docs_drift tests.test_api_limitations_doc tests.test_doc_tool_counts tests.test_tool_registration tests.test_mcp_resources` (59 tests passed, 1 conditional skip)
- `venv/bin/python -m compileall -q src scripts tests`
- `git diff --check`
- live Python API connection to DaVinci Resolve Studio 20.3.2.9 in Network mode
- live MCP stdio initialization using protocol 2025-11-25, 34 registered tools, and successful `resolve_control(get_version)`

### DIFF
--- a/docs/SKILL.md
+++ b/docs/SKILL.md
@@ -14,8 +14,14 @@ session — projects, timelines, clips, color grading, Fusion compositions, audi
 render queues, and more — through natural language.
 
 DaVinci Resolve must be running with **Preferences > General > "External scripting
-using"** set to **Local**. The server auto-launches Resolve if it is not running,
-but that first connection can take up to 60 seconds.
+using"** set to **Local**, or set to **Network** with `RESOLVE_SCRIPT_HOST`
+configured to the Resolve host IP (use `127.0.0.1` on the same machine). The
+server auto-launches Resolve if it is not running, but that first connection can
+take up to 60 seconds.
+
+Network scripting permits remote control of Resolve. Use Local mode when remote
+access is unnecessary; otherwise restrict access with host firewall and network
+controls.
 
 **Session-start update note.** The first `resolve_control(action="get_version")`
 of a session returns an `mcp` block with the cached update check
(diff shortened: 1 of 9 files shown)


samuelgursky:
Thanks @double2tea — adopted and shipped in **v2.64.0**: https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.64.0

Your commit landed on `main` with authorship preserved (cherry-picked as 6543ade), so this PR won't auto-close — closing manually.

During adoption I hardened two runtime connection sites that were still on the one-argument `scriptapp("Resolve")` and would not have worked in Network mode (ce4ee6d):
- `src/analysis_dashboard.py` — the panel's own Resolve connector.
- `install.py` `verify_resolve_connection` — the post-install probe (it already had `RESOLVE_SCRIPT_HOST` in its env but the inline script ignored it, so a Network-only install reported 'Resolve not running').

Both now route through your `connect_resolve` helper. Live-validated Local and Network (`--resolve-host 127.0.0.1`) on Studio 19.1.3.7. Appreciate the clean, well-tested contribution.

Issue #73, opened 2026-06-18
Missing API: Fusion Rectangle Mask Creation, Audio Track Channel Mode Conversion, Fusion Title Template Text Modification


swayll:
# Missing API: Fusion Rectangle Mask Creation and Configuration

## Description
The DaVinci Resolve Scripting API lacks methods to programmatically create and configure Rectangle Masks in the Fusion tab. This prevents automation of visual effects that require rounded corners or custom masking on video clips.

| img1 |  img2 |
|-------|--------|
| <img width="908" height="1164" alt="Image" src="https://github.com/user-attachments/assets/43b8ba3e-ddc9-4d08-98ce-d41e0ec7d7a6" /> | <img width="808" height="866" alt="Image" src="https://github.com/user-attachments/assets/2d396f97-13d6-45f5-9048-daea11984cc2" /> | 

### Missing Functionality
- Cannot add Rectangle Mask to a clip via API
- Cannot set mask parameters: Corner Radius, Width, Height, Center X, Center Y
- Cannot connect mask to clip output node

### Example Use Case
Automating rounded corners for video clips in a batch process:
```python
# This should work but doesn't:
clip.AddFusionMask("Rectangle", {
    "corner_radius": 0.445,
    "width": 0.3,
    "height": 0.964,
    "center_x": 0.5,
    "center_y": 0.5
})
```

### Current Workaround
All Fusion mask operations must be performed manually in the Fusion tab for each clip individually.

---

# Missing API: Audio Track Channel Mode Conversion

## Description
The DaVinci Resolve Scripting API does not support changing audio channel configuration (Stereo ↔ Mono) for audio clips or tracks.

<img width="724" height="312" alt="Image" src="https://github.com/user-attachments/assets/8fd1c7c9-4d3d-46bc-9957-76eeb2e02c84" />

### Missing Functionality
- Cannot convert Stereo audio to Mono
- Cannot convert Mono to Stereo
- Cannot query current channel mode
- No batch conversion method

### Example Use Case
Converting all audio clips to Mono for consistent output:
```python
# This should work but doesn't:
audio_clip.SetChannelMode("Mono")
mode = audio_clip.GetChannelMode()  # Should return "Stereo" | "Mono"
```

### Current Workaround
Channel mode must be changed manually in the Inspector panel or Fairlight page for each clip individually. Cannot be automated.

---

# Missing API: Fusion Title Template Text Modification

## Description
[...]

PR #99, opened 2026-07-28
feat(skills): add rough-cut skill and contact-sheet tool

mbproductionmedia:
Adds a `resolve-rough-cut` skill for assembling short-form social rough cuts
from raw BTS/vlog footage, plus the frame-grid tool it depends on.

Both come out of a real vertical (1080x1920) assembly run against Resolve
Studio. The skill records six behaviours that silently produce a wrong result
rather than erroring:

- create_timeline_from_clips `end_frame` is exclusive, so the obvious
  start+duration-1 leaves a 1-frame gap between every clip
- create_timeline_from_clips fails with a bare error unless the Media Pool
  current folder is set to the clips' bin first
- phone footage carries a rotation flag (stored 3840x2160, displays
  2160x3840) and must not be reframed into a vertical timeline
- phone footage is VFR, so the timeline should match Resolve's conformed
  rate rather than the container's r_frame_rate
- timelinePlaybackFrameRate is read-only via SetSetting; it is monitoring
  only and does not affect the render
- Fusion comps built through the API never reach the render, so there is no
  scripted path to burn text over a clip

scripts/contact_sheet.py tiles media_analysis sampled frames into labelled
grids. Frames are extracted at full source resolution, so reading them
individually to satisfy the commit_vision loop costs roughly 1,100 tokens
each and a four-clip shoot exceeds 250 frames. It resolves frames across all
three artifact states and deliberately prefers visual.json while vision is
pending, because those indices are what commit_vision expects findings to be
reported against; analysis.json holds a larger superset whose numbering does
not align.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/.claude/skills/rough-cut.md
+++ b/.claude/skills/rough-cut.md
@@ -1,0 +1,111 @@
+---
+name: resolve-rough-cut
+description: Assembling a short-form social rough cut from raw behind-the-scenes or vlog footage in the DaVinci Resolve MCP. Apply when asked for a rough cut, first cut, assembly, or "make me a timeline" from a folder of footage — day-in-the-life, BTS, process, or product-shoot material destined for Reels, TikTok, or Shorts. Covers ingest, shot finding at scale, vertical timeline setup, and gapless assembly.
+---
+
+# Resolve Rough Cut — Claude Code Skill
+
+Turns a folder of raw footage into an assembled, gapless timeline. **The
+timeline is the deliverable, not a finished video.**
+
+For editorial craft (why a cut works) see `docs/guides/editorial-decision-guide.md`.
+For the analysis layer see the `resolve-media-analysis` skill. This skill is the
+assembly recipe and the traps between them.
+
+## The deliverable contract
+
+Deliver an assembled timeline. Do **not** add titles, captions, text cards,
+effects, transitions, music, or grading unless explicitly asked. Editors who ask
+for a rough cut want shot selection and pacing; styling is theirs. Adding
+graphics is work that gets thrown away — and see the Fusion trap below for why
+it also silently fails.
+
+Render an mp4 only to preview the cut, and say that is what it is.
+
+## Workflow
+
+1. **Probe before importing.** `ffprobe` every clip for `r_frame_rate`,
+   `avg_frame_rate`, and `rotation`. Two things routinely differ from what the
+   filenames and Resolve suggest — see Verified traps.
+2. **Set project settings before any timeline exists.** `timelineFrameRate` is
+   locked once a timeline is created. Use
+   `project_manager safe_set_project_settings`.
+3. **Import into a named bin**, then `media_pool set_current_folder` to it.
+4. **Analyse** — `media_analysis analyze_bin`, `sampling_mode="adaptive_capped"`.
+   Complete the `commit_vision` loop for every clip; leaving one in
+   `pending_host_vision_analysis` is a failure, not a partial success (AGENTS.md).
+5. **Find shots via contact sheets**, not one frame at a time:
+   `python3 scripts/contact_sheet.py <clip_dir> <out_dir>` — see Shot finding.
+6. **Assemble in one call** — `media_pool create_timeline_from_clips` with
+   positioned `clip_infos`.
+7. **Verify** — `timeline detect_gaps_overlaps` must return zero of both, and
+   check the total duration against the target.
+
+## Shot finding at scale
+
+`media_analysis` extracts frames at **full source resolution**. A 35-minute 4K
+clip yields 80 frames at 2160x3840. Reading those individually to satisfy
+`commit_vision` costs ~1,100 tokens each — a four-clip shoot exceeds 250 frames.
+
+Tile them instead. `scripts/contact_sheet.py` burns frame index, timestamp and
+`selection_reason` onto each tile, so per-frame findings stay reportable by
+index. Roughly a 15x saving with no loss of coverage.
+
+Ask for **short clips per subject** rather than 20-35 minute takes when the
+shooter can choose. `adaptive_capped` tops out at 80 frames per clip, so a
+35-minute take gets sampled only to about its first 16 minutes.
+
+## Verified traps
+
+Confirmed live against Resolve Studio, 2026-07-28. Each one silently produces a
+wrong result rather than an error.
+
+| Trap | Symptom | Fix |
+|---|---|---|
+| `clip_infos` `end_frame` is **exclusive** | 1-frame gap between every clip, and matching audio gaps | `end_frame = start_frame + duration` |
+| `create_timeline_from_clips` needs the current folder | Bare `Failed to create timeline from clip_infos`, valid clip_ids | `media_pool set_current_folder` to the clips' bin first |
+| Phone footage carries a `rotation` flag | Stored 3840x2160, displays 2160x3840 | Check `rotation` in ffprobe; Resolve honours it. **Do not reframe** — it is already vertical |
+| Phone footage is **VFR** | `avg_frame_rate` differs per clip and from `r_frame_rate` | Match the timeline to what Resolve conforms to (its reported clip FPS), not to `r_frame_rate` |
+| `timelinePlaybackFrameRate` is read-only | `set_setting` returns False for every value form | Monitoring only — the render follows `timelineFrameRate`. Tell the user to fix it in Master Settings |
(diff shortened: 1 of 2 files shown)


samuelgursky (changes requested):
Thanks for this — the skill is well-written, it matches the `.claude/skills/*.md` conventions, and several of the traps are real findings I want in the repo. The exclusive `end_frame` one in particular is documented nowhere today, and a silent 1-frame gap between every clip is exactly the class of bug that costs someone an afternoon. The contact-sheet cost argument is sound too, and the note about preferring `visual.json` while vision is pending because those are the indices `commit_vision` reports against is the kind of detail that only comes from actually doing the run.

I'd like to take it, but not quite as-is. Four things:

**1. The Fusion claim needs a repro before I can merge it.**

> Fusion comps built through the API never reach the render ... there is no scripted path to burn text over a clip

This contradicts shipped, live-verified tooling. v2.57.0 added `fusion_comp(action="set_text_plus" / "get_text_plus")` and `fusion_comp(action="add_fusion_mask")` for issue #73, and `timeline(action="set_title_text")` covers generator titles. The supported path for burning text over a clip is to insert a Fusion title (`InsertFusionTitleIntoTimeline`) and set its text, not to hand-build a comp on the clip and expect it to render.

I suspect what you hit is real but narrower — a comp built tool-by-tool that never gets connected to `MediaOut`, or one built on a clip whose comp isn't the active one. That's worth documenting precisely. But "never reach the render" as a blanket statement in a skill will steer people away from a surface that works, and skills are expensive to un-teach. Can you post the exact call sequence you used? If it turns out the shipped path also fails for you, that's a more important bug than this PR and I'd want it as its own issue.

Same question, smaller stakes, on "`delete_comp` also fails" — there's a `ti_delete_fusion_comp` in `src/granular/timeline_item.py`; what did it return for you?

**2. The verified traps belong in the API-limitations ledger, not only in a skill.**

`docs/reference/api-limitations.md` is generated from `submit`-tagged `api_truth` entries, and it's what the whole toolset queries at runtime (`resolve_control api_truth "..."`). Findings that live only in a skill file are invisible to everything else and drift silently. The read-only `timelinePlaybackFrameRate` and the current-folder requirement on `create_timeline_from_clips` in particular read like ledger entries. Happy for the skill to keep its table as the narrative version — it just shouldn't be the only copy.

**3. `contact_sheet.py` needs a test.**

`load_frames()` is the part that will break: three artifact states, a deliberate ordering between them, and the 1-based `meta` indexing in `build_sheets` that has to line up with the frame numbering `commit_vision` expects. That reasoning is well-argued in the docstring and completely unguarded. A small test with fixture JSON covering the three states — and specifically that `visual.json` wins while vision is pending — would pin the invariant you're describing. This repo leans hard on that (2247 offline tests as of `main`).

**4. Pillow, minor.**

The repo already treats Pillow as optional and degrades gracefully (`src/utils/analysis_caps.py` no-ops without it). A standalone CLI script hard-failing is a defensible difference, and your error message is good — just worth a line in the skill saying the tool needs `pip install Pillow`, since the skill is where people will meet the script.

None of this is a rewrite. If you'd rather not chase down (1), I'm happy to merge the skill with the Fusion row cut and the rest intact — that's the only part I actively don't want to publish, and the other three are small.

mbproductionmedia:
Here’s the cleaned up version:
------------------------------

Hey @samuelgursky, thank you so much for this - genuinely amazing work.
This skill has been huge for me, it’s let me multitask my BTS content work
alongside client work without losing track of things.

Honestly, all the credit goes to you for building this repo in the first
place and putting it out there for people like me to use. Feel free to do
whatever you think is best to make it even more useful for others - I trust
your judgment on the direction.

On your points - happy to go with dropping the Fusion row for now rather
than digging through the exact call sequence, so we can get this merged.
I’ll add the Pillow line to the skill and take a look at the test for
contact_sheet.py. Will push the updates to the same branch shortly.

Thanks again for the thorough review!
Best
Muhammed Bucheeri

samuelgursky:
I tested this myself rather than leaving it with you, and **you were right — I was wrong to push back on the Fusion row.** Verified live on Studio 19.1.3.7, disposable project, synthetic media.

**Your path fails exactly as you described.** I added a Fusion comp to a media clip via the API, added a Text+ and a Merge, wired `MediaIn1 → Merge.Background`, `Text+ → Merge.Foreground`, `Merge → MediaOut1`, and set `StyledText`. Everything reported success and everything read back correctly — `GetCompCount` 1, `MediaOut1.Input` connected to the Merge, `StyledText` returning the value I'd just set. The render came out as the untouched source clip.

I then pushed it harder than your report did, to rule out my own "you probably didn't connect MediaOut" theory: I rewired `MediaOut1.Input` to the Text+ **only**, with no path from `MediaIn` at all. If the comp were being evaluated, the output could not possibly still be the source clip. It was, pixel for pixel. `LoadFusionCompByName` on the sole comp didn't change it either. So the comp isn't mis-wired — it's never consulted.

**Where my pushback was right, and it's a narrow point:** the failure is specific to comps attached to media-backed clips, not to Fusion through the API generally. `InsertFusionTitleIntoTimeline` + `fusion_comp set_text_plus` **does** render — I set text through the API and it appeared in the output in the same session. So the accurate statement is narrower and more useful than either of us had it:

> A Fusion comp attached to a media clip via the API never reaches the render, even though the whole graph reads back correctly. Fusion titles inserted as their own timeline clip do render, and their Text+ can be set through the API.

That's now recorded in the `api_truth` ledger and `docs/reference/api-limitations.md` (5c66ddf), tagged as a bug to submit to Blackmagic, with your repro as the evidence. Thank you for finding it — it's a genuine silent-failure that had no coverage.

One caveat worth knowing for the skill's advice: the title route can't fully replace what you wanted, because the API can't choose the destination track (issue #74 — `Insert*IntoTimeline` takes no `trackIndex` and always lands on V1). So "burn text over an existing clip" really is blocked end-to-end, just for two stacked reasons rather than the one. Your practical conclusion stands.

**On the PR:** I'd like to take it. Could you reword that row along the lines above — naming the clip-attached case and noting the title path works — so the skill doesn't steer people off a surface that does function? The other three points from my review still stand, and the smallest one is now the most useful: a test for `load_frames()` in `contact_sheet.py`, since the three-artifact-state ordering is the part most likely to rot.

If you'd rather not do another pass, say so and I'll take the PR as-is and adjust that row myself — you've already done the expensive part.

mbproductionmedia:
Hey, really appreciate you going through the trouble of verifying it
yourself - that level of care is rare and I don’t take it for granted.


Given the nuance you found, I think you’re better positioned to word that
row precisely than I am, so please go ahead and adjust it yourself as you
see fit. Same for the rest of the PR - if you want to take it as-is and
make the changes, I’m happy to let you run with it. I’m mostly using the
skill for my own workflow, so I trust you to shape it in the way that best
serves the repo and everyone else using it.


Thanks again for all the work here
Best
Muhammed Bucheeri

samuelgursky:
Merged, and I've finished the four points here as you suggested — thank you for the trust, and for the report that turned into a real ledger entry.

What landed on top of your branch ([fe7bbb0](https://github.com/samuelgursky/davinci-resolve-mcp/commit/fe7bbb0)):

- **Fusion row reworded** to name the clip-attached case specifically, and to point at the title path that does work, so the skill doesn't send people away from a working surface. Your conclusion is kept — it's just now carrying the second reason it's true (issue #74, no destination track from the API).
- **Two more of your findings promoted to the `api_truth` ledger**: `timelinePlaybackFrameRate` being unsettable, and `CreateTimelineFromClips` failing with a bare unhelpful error when the current folder isn't the clips' bin. They're queryable at runtime now (`resolve_control api_truth "..."`) instead of living only in a skill file where nothing else can see them.
- **Pillow line** added where people meet the script.
- **`tests/test_contact_sheet.py`**, 9 tests. The one that matters pins the precedence you documented: while vision is pending, `visual.json`'s indices win over `analysis.json`'s superset. I checked it actually bites by inverting the precedence — two tests fail. That was the right thing to guard: get it backwards and every tile is mislabelled while the contact sheet still looks perfectly correct.

Suite 2252 → 2261, green.

The traps are the valuable part of this PR. The exclusive `end_frame` and the current-folder requirement were both undocumented, and both are the kind that cost someone an afternoon before they think to suspect the API rather than their own arithmetic. Glad the skill is useful to your BTS work — genuinely the best reason for it to exist.

PR #100, opened 2026-07-29
fix: honor RESOLVE_SCRIPT_API/LIB env vars in get_resolve_paths

abbc400:
## What

`get_resolve_paths()` in `src/utils/platform.py` returned only the hardcoded platform defaults. Since `src/server.py` writes those values straight back into `os.environ`:

```python
os.environ["RESOLVE_SCRIPT_API"] = RESOLVE_API_PATH
os.environ["RESOLVE_SCRIPT_LIB"] = RESOLVE_LIB_PATH
```

…any `RESOLVE_SCRIPT_API` / `RESOLVE_SCRIPT_LIB` a client had already set were silently overwritten before the bridge was loaded.

This change lets those env vars win over the platform default — but only when they point at a path that actually exists, so a stale variable can't shadow a working default install. With the vars unset, behavior is unchanged.

## Why

On macOS this breaks every install where `DaVinci Resolve.app` is not under `/Applications` — in my case on an external volume. The result is hard to diagnose, because everything else looks healthy:

- Resolve Studio running, project open
- External scripting set to `Local`
- `fuscript -s` up and listening on 1144

…yet `scriptapp("Resolve")` returns `None`, and passing the correct `RESOLVE_SCRIPT_LIB` into the MCP registration has no effect, since it is overwritten at import time. `scripts/doctor.py` already surfaces the mismatch (`[FAIL] Resolve app: /Applications/...`) but the server itself has no way to act on it.

Verified against DaVinci Resolve Studio 21.0.3 on macOS 15.6 (Apple Silicon): before the change `scriptapp()` returns `None`, after it connects and `doctor.py` reports `[OK] Resolve scripting connection: DaVinci Resolve Studio 21.0.3.7`.

## Notes for the reviewer

- `scripts/doctor.py` already supports an equivalent `RESOLVE_APP` override for its own app-bundle check, so honoring env vars here is consistent with existing behavior rather than a new concept.
- Adds `tests/test_platform_paths.py` — 4 offline tests covering both directions (override applied when the path exists, default kept when the env var is unset or dangling). They fail without the fix and pass with it.
- Full offline suite: `python -m unittest discover -s tests -p "test_*.py"` → 1615 tests. The 2 errors (`test_custom_timeline`, `test_improvements` — loader errors) and 2 failures in `test_script_plugin` are pre-existing on `main` and unrelated; I verified this by running the suite with the change stashed.

### DIFF
--- a/src/utils/platform.py
+++ b/src/utils/platform.py
@@ -55,6 +55,16 @@ def get_resolve_paths():
         lib_path = "/Applications/DaVinci Resolve/DaVinci Resolve.app/Contents/Libraries/Fusion/fusionscript.so"
         modules_path = os.path.join(api_path, "Modules")
     
+    # An explicit install location wins over the platform default: Resolve is
+    # not always under /Applications (external volume, custom install dir).
+    env_api = os.environ.get("RESOLVE_SCRIPT_API")
+    if env_api and os.path.isdir(env_api):
+        api_path = env_api
+        modules_path = os.path.join(api_path, "Modules")
+    env_lib = os.environ.get("RESOLVE_SCRIPT_LIB")
+    if env_lib and os.path.isfile(env_lib):
+        lib_path = env_lib
+
     return {
         "api_path": api_path,
         "lib_path": lib_path,
--- a/tests/test_platform_paths.py
+++ b/tests/test_platform_paths.py
@@ -1,0 +1,82 @@
+"""Offline tests for src.utils.platform.get_resolve_paths (env-var overrides).
+
+Resolve is not always installed under the platform default location (external
+volume, custom install dir). The installer-style env vars RESOLVE_SCRIPT_API and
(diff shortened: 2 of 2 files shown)


samuelgursky:
Merged — thanks. The failure mode is a good catch: everything else looks healthy, which is exactly what makes it expensive to diagnose. Existence-gating the override in both directions is the right call, and the four tests cover the directions that matter.

One note for your own setup, not a problem with the PR: I couldn't reproduce the pre-existing failures you mentioned. On a clean venv, `main` before this PR ran **2237 tests, 0 failures, 0 errors**, and with your branch merged it runs **2241** — your 4 new tests and nothing else. Your run showed 1615 tests collected plus 2 loader errors and 2 `test_script_plugin` failures, which points at a venv missing some optional deps rather than anything real on main. Worth rebuilding the venv via `install.py` if you plan to send more — you're currently flying with about 600 tests not running.

PR #101, opened 2026-07-30
Pin mcp[cli]<2 so install.py doesn't pull the incompatible SDK 2.0

Mastaish:
## Problem

A fresh install currently produces a server that crashes on startup:

```
Traceback (most recent call last):
  File ".../src/server.py", line 160, in <module>
    from mcp.server.fastmcp import Context, FastMCP, Image
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

`install.py` installs the SDK with an **unpinned** requirement:

```python
# install.py, install_dependencies()
subprocess.run([str(pip), "install", "-q", "mcp[cli]"], check=True, ...)
```

`mcp` has since published **2.0.0**, which `pip install mcp[cli]` now resolves to by default. The 2.0 line restructured the SDK — there is no longer a `mcp.server.fastmcp` module (it depends on `httpx2`, `mcp-types`, etc.) — so `server.py`, which imports `from mcp.server.fastmcp import Context, FastMCP, Image`, fails at import and the MCP client reports only "Server disconnected".

## Fix

Add a constraint to `requirements.txt`, which `install.py` installs **after** `mcp[cli]`:

```
mcp[cli]>=1.29,<2
```

Because it runs second, it pins/downgrades the SDK back to the latest 1.x (which provides `mcp.server.fastmcp`) without changing the installer flow. `1.29.0` is the current newest 1.x release.

## Verification

- `from mcp.server.fastmcp import Context, FastMCP, Image` imports cleanly on `mcp 1.29.0`.
- The server completes an MCP `initialize` handshake and identifies as `DaVinciResolveMCP`.
- End-to-end smoke test against DaVinci Resolve Studio 21.0.3.7: create project → import media → build timeline → render, all via the MCP tools.

## Note (out of scope for this PR)

The other half of a broken fresh install is the Python floor: `mcp[cli]` needs 3.10+, so a venv built on macOS system `python3` (3.9) silently installs none of the MCP stack. `install.py` already checks for this, so this PR only addresses the version-resolution side. Happy to add an explicit floor note in the installer output if useful.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,9 @@
 # it, AAF preview honest-refuses (it is never faked); live AAF import into Resolve
 # does not need it (Resolve reads AAF natively).
 pyaaf2>=1.7.0
+
+# mcp — pin below 2.0. server.py imports `mcp.server.fastmcp`, which the 1.x SDK
+# provides. The 2.0.0 major restructured the SDK (no fastmcp module, switched to
+# httpx2) and is incompatible. install.py runs `pip install mcp[cli]` UNPINNED
+# first, so it grabs 2.0.0; this constraint (installed second) pins it back.
+mcp[cli]>=1.29,<2


Mastaish:
Tested this end-to-end on macOS with DaVinci Resolve Studio 21.0.3.7:

- Rebuilt the venv on Python 3.11 and installed with the pin → `mcp 1.29.0`; `from mcp.server.fastmcp import Context, FastMCP, Image` imports cleanly and the server completes an MCP `initialize` handshake as `DaVinciResolveMCP` (before the pin it crashed at import on `mcp 2.0.0`).
- Verified from **both** MCP clients — Claude Desktop and a self-hosted agent — hitting the same live Resolve instance.
- Full workflow through the tools: create project → import footage → build a 10-sec timeline → render an H.264 mp4, plus a read-back of project/timeline state from the second client.

No regressions. Confirmed the fix resolves the "Server disconnected" failure on a fresh install.

samuelgursky:
Merged — thanks, this was breaking every fresh clone and the diagnosis was exactly right.

I added one follow-up on top: the constraint is now also on the `pip install` call in `install.py` itself, not just `requirements.txt`. Your version works because requirements is installed second, but the unpinned first call still downloads 2.x and its httpx2 tree before downgrading it, and the fix quietly evaporates if that ordering ever changes.

There's a drift guard with it (`McpSdkPinTest`) that reads the actual string literals rather than matching text — worth noting because my first cut of that test matched the prose comment on install.py line 40 instead of the pip argument and passed for the wrong reason. It's conditional on `server.py` still importing `mcp.server.fastmcp`, so it retires itself when the server is ported to 2.x rather than standing in the way.

Confirmed 1.29.0 and 2.0.0 both declare `requires_python >=3.10`, so the floor you picked doesn't raise the project's Python floor.

PR #102, opened 2026-07-31
fix(test): pass Advanced test files to Node runner

double2tea:
## Summary
- replace directory arguments to node --test with explicit test-file globs
- keep library and server test scripts independently runnable
- avoid MODULE_NOT_FOUND failures on supported Node versions

## Verification
- Node 22.22.3
- npm test --prefix resolve-advanced
- 701 passed, 30 skipped, 0 failed

### DIFF
--- a/resolve-advanced/package.json
+++ b/resolve-advanced/package.json
@@ -37,9 +37,9 @@
     "_comment": "audio (split/trim/convert) + conform frame ops need system ffmpeg/ffprobe on PATH (NOT bundled — FFmpeg binaries are GPL; this package stays MIT). Install ffmpeg yourself: brew install ffmpeg."
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


samuelgursky:
Merged — thanks, and thanks for pinning it to a Node version. I reproduced it before merging: on 22.22.3 the directory form gives `MODULE_NOT_FOUND` at 4 fail / 0 pass, so the suite wasn't running *at all* on Node 20+, it just wasn't obvious from the exit path. With the globs it's the full 731 / 701 pass / 30 skip on 18, 20 and 22 alike.

One thing that briefly confused my verification, in case it bites you: a checkout whose `better-sqlite3` was built against Node 18 throws 31 failures under Node 22 (`NODE_MODULE_VERSION 108` vs `127`) that look like real test failures. `npm rebuild` clears them — unrelated to this change, but easy to misread as a regression from it.

Issue #70, opened 2026-06-18
AutoSyncAudio silently fails on string-keyed settings (e.g. mode: "waveform") — needs live enum resolution


ohwowtri:
## Summary
`media_pool.auto_sync_audio` and `timeline.safe_auto_sync_audio` both return `success: false` 
with no error message when `settings` uses human-readable strings (e.g. `"method": "waveform"`). 
The underlying Resolve scripting API requires `AUDIO_SYNC_*` enum constants resolved from the 
live `resolve` handle, not plain strings — but neither tool currently does that translation 
before calling `MediaPool.AutoSyncAudio()`.

## Environment
- Server: compound mode, build pulled 2026-06-16
- DaVinci Resolve Studio: 21.0.0

## Reproduction
1. Two MediaPoolItems with embedded audio (confirmed: 2ch / 48kHz / Linear PCM on both test clips)
2. Call `timeline.safe_auto_sync_audio`:
```json
   {"clip_ids": ["<id1>", "<id2>"], "dry_run": false,
    "settings": {"group_id": "test", "method": "waveform", "primary_clip_id": "<id1>"}}
```
3. Response:
```json
   {"success": false, "verified": false, "linked": [], "newly_linked": [],
    "already_linked": [], "count": 2, "missing": []}
```
4. Same result calling the raw `media_pool.auto_sync_audio` action directly.

## Root cause — confirmed by the server's own `api_truth` action
Calling `resolve_control.api_truth(query="AutoSyncAudio")` returns:
```json
{
  "symbol": "MediaPool.AutoSyncAudio",
  "reality": "The boolean return does not reflect whether clips actually linked, and string
              enum keys in `settings` are silently rejected (the call returns False).",
  "recommended": "Resolve the AUDIO_SYNC_* enum constants via the live resolve handle, and
                   verify by reading each clip's 'Synced Audio' property."
}
```
So the bug is self-diagnosed in the knowledge base, but the two tool implementations don't 
apply that fix — they appear to pass the `settings` dict straight through instead of resolving 
`mode`/`method` against `resolve.AUDIO_SYNC_MODE`, `resolve.AUDIO_SYNC_WAVEFORM`, 
`resolve.AUDIO_SYNC_TIMECODE` etc. as live attributes first.

## Suggested fix
Before calling `MediaPool.AutoSyncAudio(clips, settings)`, translate human-readable settings 
into real enum values via the live handle, conceptually:
```python
resolved = {}
[...]

PR #107, opened 2026-08-03
fix(resolve21): AI methods return an error string, not a bool — plus live validation on Studio 21.0.2

AghisSs:
Live-validated the Resolve 21.0 scripting additions against **Studio 21.0.2.4**. The surface was implemented and version-gated but covered only by stubs, and stubs cannot see how Resolve actually reports a missing AI Extras pack.

## The bug

It reports it inconsistently, and one of the shapes is a trap:

| Method | Return when the Extra is absent |
|---|---|
| `AnalyzeForSlate` | `False` |
| `AnalyzeForIntellisearch` | `"Required package 'AI Intellisearch - Faster' is not installed."` |
| `GenerateSpeech` | `"Required Package, 'AI Speech Generator' is not Installed."` |

A non-empty string is truthy, so the existing wrappers were wrong in two different ways:

- `ok = bool(clip.AnalyzeForIntellisearch(...))` → **`{"success": true}` for an analysis that never ran** (`src/server.py`, 8 folder/clip call sites).
- `generate_speech` did `if not new_item: ...` then `new_item.GetName()` → the string sails past the guard and raises **`AttributeError: 'str' object has no attribute 'GetName'`**.

Fixed with `_ai_result` / `_ai_result_payload`: any string return is a failure, and its text becomes the surfaced error. That message is the only machine-readable signal that a pack is missing, since nothing in the scripting API enumerates installed Extras (recorded as a separate `submit: "missing"` entry).

## Also included

These are bundled because they are load-bearing for the above, not as drive-by changes:

- **`project_settings("reset_intellisearch_analysis")`.** `Project.ResetIntellisearchAnalysis` is documented in the scripting README shipped with 21.0.2 (26 May 2026) and present in `dir(project)`, but was absent from the 5 May 2026 copy bundled in `docs/reference/`, so it was never wrapped. Refreshing the bundled doc without the wrapper fails `audit_api_parity`, so both land together. The method-name diff between the two doc versions is exactly this one addition.
- **`tests/live_resolve21_validation.py`** — preflight gate, disposable project pinned to a temp media location, synthetic `ffmpeg` + `say` media, full-dict readback diffing, teardown that restores the originally-open project. `RemoveMotionBlur` renders media so it is behind `--allow-render`; `DisableBackgroundTasksForCurrentResolveSession` is session-wide with no undo so it is behind `--allow-session-mutation` and was **not** executed.
- **`api_truth`** — the string-return bug, the undiscoverable-Extras gap, the absent `resolve.MARKER_*` constants, `ResetIntellisearchAnalysis`, and the audio-classification readback. `VERIFIED_ON` bumped to 21.0.2, `api-limitations.md` regenerated.
- **Coverage docs** — Resolve 21.0.2 recorded as tested, the "Resolve 21 beta APIs are intentionally deferred" note dropped, a Resolve 21 delta row and per-method detail table added. Also fixes the pre-existing 33-vs-34 compound-tool contradiction in `api-coverage.md` (line 9 said 33, line 20 said 34).

## Two corrections to existing records

**`hasattr`/`getattr` attribute fabrication does not reproduce on 21.0.2.4.** The `api_truth` entry (verified on 21.0.0) says the bridge returns a callable for any attribute name, which would make `_has_method` — and every `_requires_method` gate — dead code against a live Resolve. A control probe of a nonexistent method across `Resolve`, `ProjectManager`, `Project`, `MediaPool`, `Folder`, `MediaPoolItem`, `Timeline` and `TimelineItem` returned `getattr`-callable `False` on **all eight**, matching `dir()` in every case. The entry is now scoped as build-dependent rather than absolute. `dir()` remains the recommendation, since it is correct on every build measured. The `has()` docstring in `live_api_gap_verification.py` carries the same absolute claim and may deserve the same treatment.

**`AnalyzeForSlate`'s documented `markerColor` constants do not exist.** The shipped 21.0.2 README says the argument must be one of `resolve.MARKER_BLUE` etc.; on 21.0.2.4, `[c for c in dir(resolve) if c.startswith("MARKER_")]` is **empty**. There is no documented-correct way to call this method. Recorded as a `submit: "bug"` — deliberately *not* tagged `enum`, since that tag denotes the issue-#70 class where a resolver must translate strings into live constants, and here there is nothing to resolve.

## Validation

**Environment:** DaVinci Resolve Studio 21.0.2.4 (build 21.0.20004) · macOS Darwin 25.5.0 · Apple Silicon · Python 3.12.4

**Extras installed at test time: AI Motion Deblur only.** `AnalyzeForIntellisearch`, `AnalyzeForSlate` and `GenerateSpeech` are therefore marked 🔬 untested-for-lack-of-a-pack, never "broken".

| Method | Result | Evidence |
|---|---|---|
| `PerformAudioClassification` | ✅ | Returned `True`; `Category` clip property `""` → `Dialogue` |
| `ClearAudioClassification` | ✅ | Returned `True`; `Category` reset to `Uncategorized` — **not** `""` |
| `RemoveMotionBlur` | ✅ | Returned a `MediaPoolItem`; source media path byte-identical after |
| `ResetIntellisearchAnalysis` | ✅ | Returned `True` |
| `TranscribeAudio(useSpeakerDetection)` | ⚠️ | Parameter accepted; `True` and `False` gave identical transcripts on a deliberately two-voice clip |
| `AnalyzeForIntellisearch` | 🔬 | Requires AI IntelliSearch; returned an error **string** |
| `AnalyzeForSlate` | 🔬 | Requires AI Slate ID; returned `False` |
| `GenerateSpeech` | 🔬 | Requires AI Speech Generator; returned an error **string** |
| `DisableBackgroundTasksForCurrentResolveSession` | 🔬 | Present in `dir()`; **not executed** — session-wide, returns `None`, no `Enable...` counterpart, no undo short of restarting Resolve |

Confirm-gated operations: `RemoveMotionBlur` executed with an explicit token. `GenerateSpeech` executed with `AddToTimeline: false` only. `DisableBackgroundTasksForCurrentResolveSession` not executed.

**Checks run:**

```
tests/test_import.py                        ok
scripts/audit_api_parity.py                 PASS — all checks clean
scripts/gen_api_limitations.py --check      OK
node scripts/agent-rules/generate.mjs --check   in sync
python -m unittest discover -s tests        Ran 2297 tests — OK (skipped=25)
node bin/davinci-resolve-mcp.mjs --version  2.70.0
npm pack --dry-run                          509 files
git diff --check                            clean
```

No version bump or tag — left to the maintainer per `docs/process/release-process.md`. Happy to split this into separate commits if you'd prefer the fix landed independently of the doc refresh; they're bundled because the parity guard couples them.


samuelgursky (commented):
Reviewed on the branch against Resolve's bundled reference and the offline suite. This is careful work — the discipline of marking the three Extras-blocked methods 🔬 rather than "broken" is exactly right, and the live harness is a genuine asset. Notes below, roughly in order of how much they'd matter at merge time.

## The core fix is correct

`_ai_result` / `_ai_result_payload` are sound: a string is a failure with its text surfaced, everything else goes through `bool()`. All eight converted call sites are mechanical and correct, `generate_speech`'s `ok` / `message` correctly survive the `with` block, and the ledger now records the true outcome instead of a fabricated success.

The guards all pass on the branch:

```
scripts/audit_api_parity.py            PASS — all checks clean
scripts/gen_api_limitations.py --check OK
node scripts/agent-rules/generate.mjs --check   in sync
python -m unittest discover -s tests   2297 tests
```

## `remove_motion_blur` reproduces both of the bugs this PR fixes

The two `RemoveMotionBlur` call sites weren't converted. `RemoveMotionBlur` takes an Extras pack like the others — the new `api_truth` entry in this PR says so explicitly ("AnalyzeForIntellisearch, AnalyzeForSlate, GenerateSpeech **and RemoveMotionBlur** each require a separately-downloaded Extras pack"). It was tested with the Extra installed, so the absent-pack return was never observed, but if it returns a string the way its two siblings do:

- **`media_pool_item.remove_motion_blur`** — `if new_clip:` passes, `_clip_file_size` swallows its own `AttributeError`, then `return {"success": True, "new": new_clip.GetName(), ...}` raises **`AttributeError: 'str' object has no attribute 'GetName'`**. That is `generate_speech`'s crash verbatim, in the other method whose documented return type is `MediaPoolItem`.
- **`folder.remove_motion_blur`** — `for pair in (result or [])` iterates the string's characters, `orig, new = pair` raises, `except Exception: continue` eats it, and the action returns **`{"success": True, "created": []}`** with `_rec.success = True` written to the ledger. The silent lie, in the confirm-gated render path.

Both are a one-line change to the helper this PR already adds. Worth doing here — it's the difference between fixing the class and fixing the instances that happened to surface.

## The `hasattr` control probe doesn't test what the original entry claimed

This is the change I'd most want to revisit, because it rescopes a load-bearing record.

The original entry's evidence was *real API method names borrowed from other object types*: "hasattr reported SetStart, Razor, AddNode, GenerateProxy, AddSmartBin etc. as present though none exist." The control probe in `tests/live_resolve21_validation.py` uses a single synthetic string:

```python
fake = "TotallyMadeUpMethod_xyz123"
```

Those are different hypotheses. If the bridge's actual behaviour is "resolve any name known to the RemoteObject method table, regardless of which object you asked", then a nonsense name correctly returns `False` on 21.0.0 as well, and the probe says nothing about the original observation. And that distinction is exactly the case `_requires_method` guards: a Resolve 20 build where the 21-only method doesn't exist on that object, but where the *name* may still be known to the bridge.

The rewritten entry now tells future readers that `_has_method` is "sound on 21.0.2.4". That may well be true, but this probe doesn't establish it. Re-running with the five original names would settle it cheaply, and the harness is already built. Until then I'd keep the entry scoped as "unresolved on 21.0.2.4" rather than "no longer reproduces".

Everything else in that entry checks out — `_has_method` really is `callable(getattr(obj, name, None))`, and `disable_background_tasks_for_current_session` really does `return _ok()` unconditionally.

## The doc refresh surfaces a page the server rejects

The refreshed reference adds `photo` to `OpenPage` / `GetCurrentPage`:

```
OpenPage(pageName)  # ... ("media", "photo", "cut", "edit", ...)
```

but `resolve_control.open_page` validates against a hardcoded enum that doesn't include it, so `open_page("photo")` returns a validation error while the repo now ships a doc saying it's valid. `audit_api_parity` passes because it tracks method names, not enum values — the same blind spot as the codec description-vs-id trap. Either add `photo` to the enum or record it as a known gap so the refresh doesn't quietly introduce a doc/behaviour mismatch.

## The coverage numbers contradict each other

This PR fixes the 33-vs-34 contradiction and introduces a new one:

- Header: `Live Test Pass Rate | 332/332 (100%)`
- Phase table five rows down: `**Total** | **339/340** | **99.7%**`

On `main` those two agreed (331/331). Related, in the same file: `### Untested Methods (5 of 336)` is untouched — stale denominator, and it still lists only the four cloud methods plus `AnalyzeDolbyVision`, while this PR's own narrative adds three Extras-blocked methods and one deliberately-not-executed. The header's `332/337` implies exactly five untested; by the PR's own accounting it should be about nine. The `8/9 (89%)` delta row also counts four 🔬-untested rows as passes.

The candour in the prose is the best thing about this PR. The arithmetic should match it.

## Two smaller things

**The speaker-detection finding won't reach anyone.** `TranscribeAudio(useSpeakerDetection)` producing identical transcripts for `True` and `False` on a deliberately two-voice clip is, after the string bug, the most user-visible result here — and it's recorded only in a coverage table. There's no `api_truth` entry for it, so it never reaches `api-limitations.md` or the agent-facing docs. It wants an entry with `submit: "bug"`.

**`test_generate_speech_string_return_does_not_raise` doesn't test that.** The docstring says "The string must not reach `.GetName()` — that raised AttributeError", but the body only calls `_ai_result(self.SPEECH_MSG)`; it never invokes `project_settings("generate_speech")`. It duplicates `test_ai_result_treats_string_as_failure` under a name that claims regression coverage for the crash, so the crash itself is untested. The four folder-level call sites have no coverage either — only `media_pool_item.analyze_for_intellisearch` is exercised end to end.

## Summary

The central fix, the harness, and the `api_truth` entries are worth landing. Before merge I'd want `remove_motion_blur` converted (same class, one line, and it's the confirm-gated render path), the `hasattr` entry rescoped to "unresolved" until the original names are probed, and the coverage arithmetic reconciled. The `photo` page and the speaker-detection entry are fine as follow-ups. The branch is one commit behind main — worth a rebase before merge.

On the offer to split: no need on my account. The parity guard genuinely couples the wrapper and the doc refresh, and the reasoning for bundling is stated clearly enough to follow.

billcarroll:
@AghisSs — distilling the review above into an actionable list, since the pre-merge asks are spread across a long write-up. Three things before merge:

1. **Convert the two `remove_motion_blur` call sites** (`media_pool_item.remove_motion_blur` and `folder.remove_motion_blur`) to the `_ai_result` / `_ai_result_payload` helpers your PR adds. Per the review they reproduce both bugs you're fixing — the `.GetName()` AttributeError on a string return, and the silent `{"success": True, "created": []}` in the confirm-gated render path. One-line change each with your helper already in place.
2. **Rescope the `hasattr` `api_truth` entry** to "unresolved on 21.0.2.4" — the synthetic-name control probe (`TotallyMadeUpMethod_xyz123`) doesn't test the original observation, which used real method names borrowed from other object types (`SetStart`, `Razor`, `AddNode`, `GenerateProxy`, `AddSmartBin`). Re-running your harness with those five names would settle it properly if you'd rather keep the stronger claim.
3. **Reconcile the coverage arithmetic** — header says 332/332 (100%), the phase table says 339/340 (99.7%), and "Untested Methods (5 of 336)" doesn't count the PR's own three Extras-blocked methods plus the deliberately-not-executed one.

Flagged as fine-for-follow-up: the `photo` page enum/doc mismatch, an `api_truth` entry for the speaker-detection finding (`submit: "bug"`), real end-to-end regression coverage for the `generate_speech` crash, and a rebase (one commit behind).

The core fix and the live harness both got a clear thumbs-up — this is close.

samuelgursky:
@AghisSs — I've pushed the `remove_motion_blur` conversion to this branch myself (`ad07156`) rather than leave it sitting on you. It was the one pre-merge ask that was mechanical, and it's your helper doing the work, so there was nothing to decide.

**Please review it as you would any other change to your PR** — and revert it if you disagree with the shape. Both remaining asks are still yours.

## What it does

Routes both `RemoveMotionBlur` call sites through the `_ai_result` helper this PR adds. `RemoveMotionBlur` needs the AI Motion Deblur Extra exactly as its siblings need theirs — the `api_truth` entry you wrote says so explicitly — so absent the pack its return can be a string too. That left the one action in the family that *renders new media* reproducing both bugs the rest of the PR fixes.

I confirmed both against your unconverted code rather than assuming. Running the new tests with the conversion reverted:

```
ERROR: test_clip_string_return_does_not_raise_attributeerror
AttributeError: 'str' object has no attribute 'GetName'

FAIL: test_folder_string_return_is_not_reported_as_success
AssertionError: True is not false : created:[] alongside success:true was the silent lie
```

The first is `generate_speech`'s crash verbatim. The second returned `{"success": True, "created": []}` with a true success written to the ledger — behind a confirm gate whose entire purpose is that the caller knows new media is about to be rendered.

A third test pins the installed-pack path, so the working case is proven unchanged rather than just assumed.

Worth being clear that this was missed, not skipped: you live-tested `RemoveMotionBlur` with the Extra installed, so the absent-pack return was never observable on your machine. That's the same reason the string bug existed at all.

## Still yours, both judgement calls

**1. The `hasattr` entry.** Your control probe uses one synthetic name (`TotallyMadeUpMethod_xyz123`), but the original 21.0.0 evidence was *real API method names borrowed from other object types* — `SetStart`, `Razor`, `AddNode`, `GenerateProxy`, `AddSmartBin`. If the bridge's actual behaviour is "resolve any name known to the RemoteObject method table", a nonsense name returns `False` on 21.0.0 too and the probe doesn't test the original claim. Re-running with those five names would settle it, and your harness already does the hard part. Until then I'd scope the entry as unresolved on 21.0.2.4 rather than "no longer reproduces".

**2. The coverage arithmetic.** Header says `332/332 (100%)`, the phase table five rows down totals `339/340 (99.7%)`, and `### Untested Methods (5 of 336)` is untouched — while your own narrative adds three Extras-blocked methods plus one deliberately not executed. The prose is the most honest part of this PR; the numbers should match it.

## State

- Full suite on the branch: **2300 tests**, only the two known `test_color_grade_probe` / `test_render_deliver_probe` failures, which are the cwd-dependent guard tests fixed on main in `33b73f8` — unrelated to your work and cleared by a rebase.
- `audit_api_parity` PASS, `gen_api_limitations --check` OK, drift guards OK.
- The branch is now several commits behind main (#108 and #110 merged since). A rebase before merge would also clear those two failures.

Once the `hasattr` scoping and the numbers are settled this is good to go from my side.

samuelgursky:
@AghisSs — rebased onto main and merging this, since it has been quiet and the work is good. Two commits pushed on top of yours; revert either if you disagree.

**`6b78015` — the `hasattr` entry is now scoped as UNRESOLVED rather than "no longer reproduces".** Your control probe used the invented name `TotallyMadeUpMethod_xyz123`, but the 21.0.0 evidence it overturns was *real API method names borrowed from other object types* — `SetStart`, `Razor`, `AddNode`, `GenerateProxy`, `AddSmartBin` reported present on objects that don't have them. If the bridge resolves any name known to the RemoteObject method table rather than literally any string, an invented name is correctly rejected on both builds and the probe never exercised the failing case.

That mattered enough to change rather than leave: the entry told readers `_has_method` is "sound on 21.0.2.4", and `_has_method` is what every `_requires_method` version gate is built on. A gate that silently passes is the exact failure this ledger exists to prevent. The entry now names the five original names as the probe that would settle it. **That re-run needs your 21.0.2.4 — it can't be done here**, and it's the one genuinely outstanding item.

**`ced6f7c` — the `remove_motion_blur` conversion** (from earlier), unchanged.

## What I did not touch, and why

The coverage arithmetic. I went to fix it and found the problem is deeper than stale numbers: **none of the Resolve 21 methods appear in the Complete API Reference per-class tables** — the very tables the `336`/`337` denominator counts.

```
PerformAudioClassification: 0 occurrences
AnalyzeForIntellisearch:    0
GenerateSpeech:             0
RemoveMotionBlur:           0
ResetIntellisearchAnalysis: 0
```

So `337/337` adds one to a denominator whose table doesn't list the method, and the header (`332/332`) versus phase table (`339/340`) disagreement can't be reconciled without deciding how that whole surface should be counted. That predates your PR — the Resolve 21 methods were never in those tables. Picking numbers would have made the contradiction invisible rather than fixed, so I've left it and filed it as its own piece of work.

Your prose is the honest part of this and always was: marking three methods 🔬 untested-for-lack-of-a-pack rather than "broken" is the right call, and it's what made the counting problem visible in the first place.

## State

Rebased onto main (16 commits behind; the only conflict was the generated `api-limitations.md`, resolved by regenerating). Suite **2383 tests OK**, `audit_api_parity` PASS, `gen_api_limitations --check` OK, agent-rules in sync.

Thanks for this — the live validation harness and the string-return finding are the substantial contributions here, and the `_ai_result` helper is now doing work across five actions.

PR #108, opened 2026-08-03
fix: probe_media_pool silent depth truncation + script runner exit-race segfault

billcarroll:
Two small, independently testable fixes, each with a regression test.

## 1. probe_media_pool silently truncates at the depth cutoff

`_folder_probe` reported `subfolder_count` from the list it had *expanded* — always empty below the depth cutoff — so with the default `depth: 1`, every unexpanded folder looked like an empty leaf (`clip_count: 0, subfolder_count: 0`) with no indication anything was hidden.

Observed in the field: a drive-rename relink sweep walked the probe result of a project whose bins nest three levels deep, trusted `subfolder_count: 0`, and skipped ~40 populated bins / 573 clips. `folder.get_subfolders` on the same path correctly showed 3 subfolders; `refresh` changed nothing because nothing was stale — the probe payload was simply lying at the cutoff.

Fix: `subfolder_count` now reports the real `GetSubFolderList()` length at every level regardless of depth, and each folder node carries `truncated: true` when depth stopped expansion, so walkers can detect the cutoff and descend. `media_pool_boundary_report` shares the helper and inherits the fix. Docstring updated.

## 2. fusionscript's segfault-at-exit race flips script results to failure

`fusionscript.so`'s RemoteApp thread keeps dispatching packets from Resolve while the interpreter tears down at process exit, and can SIGSEGV *after* a spawned script has finished its work (crash signature: `Fusion::RemoteApp::FindLocalObject` on the RemoteApp thread while the main thread is inside `exit()` / `__cxa_finalize`). The child then exits -11 and `_execute_python_script` reports `success: false` for a run that succeeded — and macOS shows a crash report.

Fix: run spawned Python scripts via `runpy` and hard-exit with `os._exit(0)` after flushing output, skipping the teardown the race lives in. Real failures are unchanged: exceptions still traceback and exit nonzero, `sys.exit(n)` still propagates (verified by test), and argv semantics are preserved.

## Testing

- New regression tests: `tests/test_media_pool_ingest_probe.py::test_folder_probe_reports_true_subfolder_count_beyond_depth`, `tests/test_script_plugin.py::TestPythonScriptExitGuard` (simulates the segfault with an atexit SIGSEGV handler; POSIX-only, skips elsewhere)
- Full offline suite: 2217 tests green (except 4 pre-existing numpy-missing import errors unrelated to this change)

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -25568,14 +25568,32 @@ def _python_env_for_resolve() -> Dict[str, str]:
 # the interpreter tears down at exit, and can SIGSEGV *after* the script has
 # finished — turning a successful run into exit code -11 / success:false.
 # Run the script via runpy and hard-exit before teardown so the exit code is
-# truthful; on any exception the normal traceback + nonzero exit still happen.
+# truthful. SystemExit must be caught here: uncaught, a plain sys.exit(0) at
+# the end of a script would take the normal teardown path and reopen the
+# segfault window. sys.path[0] is pointed at the script's directory to mimic
+# `python script.py` (under -c it points at the server's cwd, which both
+# breaks sibling imports and lets stray files there shadow real modules).
+# Cost of os._exit: atexit handlers never run and non-daemon threads are not
+# joined — documented in script_plugin's execute action.
 _PY_SCRIPT_EXIT_GUARD = (
-    "import os, runpy, sys\n"
+    "import os, runpy, sys, traceback\n"
     "sys.argv = sys.argv[1:]\n"
-    "runpy.run_path(sys.argv[0], run_name='__main__')\n"
+    "sys.path[0] = os.path.dirname(os.path.abspath(sys.argv[0]))\n"
+    "code = 0\n"
+    "try:\n"
+    "    runpy.run_path(sys.argv[0], run_name='__main__')\n"
+    "except SystemExit as e:\n"
+    "    if isinstance(e.code, int):\n"
+    "        code = e.code\n"
+    "    elif e.code is not None:\n"
+    "        print(e.code, file=sys.stderr)\n"
+    "        code = 1\n"
+    "except BaseException:\n"
+    "    traceback.print_exc()\n"
+    "    code = 1\n"
     "sys.stdout.flush()\n"
     "sys.stderr.flush()\n"
-    "os._exit(0)\n"
+    "os._exit(code)\n"
 )
 
 
@@ -26261,6 +26279,10 @@ def script_plugin(action: str, params: Optional[Dict[str, Any]] = None) -> Dict[
         — args: list of CLI args for the Python subprocess (Python only).
         — timeout: seconds (default 120 for execute, 60 for run_inline).
         — Auto-launches Resolve if not running.
+        — Python scripts hard-exit after the script body (guards against
+          fusionscript's segfault-at-exit race), so atexit handlers do not
+          run and non-daemon threads are not joined. Do cleanup inline or
+          in try/finally, not in atexit.
       run_inline(source, language, timeout?) -> {success, stdout?, stderr?, result?}
         — Python: writes to temp file with `resolve`/`project`/`mp`/`timeline`
           pre-bound, runs as subprocess, captures stdout/stderr.
--- a/src/server.py
+++ b/src/server.py
@@ -10877,9 +10877,10 @@ def _folder_probe(folder, depth: int = 1):
     clips = []
     for clip in (folder.GetClipList() or []):
         clips.append(_media_pool_item_summary(clip))
+    subs = folder.GetSubFolderList() or []
     subfolders = []
     if depth > 0:
-        for sub in (folder.GetSubFolderList() or []):
+        for sub in subs:
             subfolders.append(_folder_probe(sub, depth - 1))
     stale = None
     try:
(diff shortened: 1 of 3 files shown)


samuelgursky (commented):
Reviewed both fixes on the branch, running against the project venv. Thanks for the clear write-up — the field evidence on the probe bug (40 bins / 573 clips skipped) is exactly the kind of detail that makes a report actionable.

## Fix 1 — `_folder_probe` subfolder_count: correct, ready to land

Diagnosis and fix both check out:

- `subfolder_count` has exactly one producer and no internal consumer, so widening it can't break a caller.
- `truncated = len(subs) > len(subfolders)` is right at both ends: with `depth > 0` one entry is appended per sub, so the lengths match and it reads `false`; at `depth == 0` the list is empty and it reads `true` only when children actually exist.
- The `media_pool_boundary_report` claim holds — it delegates to `_media_pool_probe`, which calls `_folder_probe`.
- One improvement the description doesn't mention: `current_folder` is probed at `_folder_probe(current, 0)`, so it *always* reported `subfolder_count: 0`. That's now truthful too.
- Docstring matches the code (`max(0, min(int(depth), 4))`, default 1).

Two non-blocking notes: `GetSubFolderList()` is now called on every folder at the cutoff where it previously wasn't, so there's one extra IPC round-trip per leaf folder (on top of the four already there); and it's unguarded at that level, so a folder that raises would now propagate. `GetClipList()` two lines above has the same exposure, so this is consistent with the surrounding style.

## Fix 2 — the exit guard: right idea, three problems in the implementation

I ran each of these through the real `_execute_python_script` on this branch, patching only `get_resolve`.

### (a) `sys.exit()` bypasses the guard, so the bug still reproduces

`runpy.run_path` lets `SystemExit` propagate; nothing catches it, so the interpreter takes the normal teardown path — precisely the window the segfault lives in. A script ending in `sys.exit(0)` with a segfaulting `atexit` handler:

```
sys.exit(0) + segfault:  success=False  exit_code=-11  stdout: (empty)
```

Identical to pre-fix behaviour, and `sys.exit(0)` at the end of a script is a very common shape. `test_argv_and_failure_exit_codes_are_preserved` exercises this exact code path with `sys.exit(3)` — it just doesn't combine it with the crash, so the gap stays invisible.

### (b) `-c` drops the script's directory from `sys.path`

`python script.py` prepends the script's directory; `python -c ...` sets `sys.path[0] = ''` (the server's cwd — `safe_run` passes no `cwd`). Any script importing a helper module next to it now fails:

```
sibling import:  success=False  exit_code=1
ModuleNotFoundError: No module named 'helper'
```

This works on `main`. The other half is equally awkward: the server's cwd is now on `sys.path` for every spawned script, so a stray `.py` there can shadow a real import.

### (c) `atexit` handlers and non-daemon threads are silently dropped

`os._exit` skips both. Confirmed: an `atexit` cleanup printed nothing (`stdout: body ran` only), and a worker thread never finished. Cleanup-on-exit is a normal pattern in Resolve utility scripts. Unclosed file writes *do* survive, because `run_path`'s globals dict is dropped and CPython refcounting closes them — but that's incidental rather than guaranteed.

### All three fall out of one rewrite

```python
_PY_SCRIPT_EXIT_GUARD = (
    "import os, runpy, sys, traceback\n"
    "sys.argv = sys.argv[1:]\n"
    "sys.path[0] = os.path.dirname(os.path.abspath(sys.argv[0]))\n"
    "code = 0\n"
    "try:\n"
    "    runpy.run_path(sys.argv[0], run_name='__main__')\n"
    "except SystemExit as e:\n"
    "    code = e.code if isinstance(e.code, int) else (0 if e.code is None else 1)\n"
    "except BaseException:\n"
    "    traceback.print_exc()\n"
    "    code = 1\n"
    "sys.stdout.flush()\n"
    "sys.stderr.flush()\n"
    "os._exit(code)\n"
)
```

Assigning `sys.path[0]` rather than inserting fixes both the missing script directory and the cwd-shadowing in one move. Four cases against it: sibling import works; `sys.exit(0)` + segfault gives `rc=0` with `work done` captured; `raise ValueError` gives a traceback and `rc=1`; `sys.exit(3)` gives `rc=3`.

The `atexit` / thread behaviour is inherent to `os._exit` and is better documented in the tool docstring than fixed.

## Tests and branch state

`tests.test_media_pool_ingest_probe` + `tests.test_script_plugin`: 90 tests, OK.

The branch is **15 commits behind main** — it's missing the headless study, the mode-matrix work, the pixel-parity work, and the `get_items_in_track` serialization fix. That explains the test-count difference (2264 here vs 2317 on main from the same runner), and it means the "no new failures" measurement was taken against a tree without ~15 commits of work. GitHub still reports the branch as mergeable, so there's no textual conflict, but a rebase and a re-run before merge would be worth it — particularly for fix 2, since the headless work that landed in those commits covers spawned-process and Resolve-teardown behaviour that this change sits right on top of.

Fix 1 is good to go as-is. Fix 2 needs the guard rewrite plus a test that combines `sys.exit(0)` with the simulated crash — as it stands it half-fixes the stated bug while breaking sibling imports.

billcarroll:
Addressed in 738ca66 (branch rebased onto current main first):

- **Guard rewrite**: applied as suggested — `runpy.run_path` wrapped in try/except, `SystemExit` mapped to its code, other exceptions print their traceback and exit 1, `sys.path[0]` pointed at the script's directory, `os._exit(code)` at the end. One deviation: `sys.exit("message")` now prints the message to stderr before exiting 1, matching CPython's own SystemExit handling, instead of silently mapping to 1.
- **Tests**: added the combined case (`sys.exit(0)` + simulated segfault-at-exit → success, rc 0, stdout captured), plus sibling-import, exception→traceback+rc 1, and `sys.exit("msg")`→stderr+rc 1. All run through the real `_execute_python_script`.
- **`os._exit` cost**: documented in the `execute` action docs — atexit handlers don't run, non-daemon threads aren't joined, do cleanup inline or in try/finally.

Rebase was clean (no conflicts with the headless/mode-matrix work). Full offline suite on the rebased branch: 2328 tests, 0 failures.

Fix 1 untouched per your review.

PR #110, opened 2026-08-03
fix(timeline): thumbnail_contact_sheet silently failed off the Color page

billcarroll:
## The bug

`timeline thumbnail_contact_sheet` (and `marker_thumbnail_review`, which routes through it) returned `"No thumbnail available at frame"` for every sample whenever Resolve was on any page other than Color. The underlying scripting call, `Timeline.GetCurrentClipThumbnailImage()`, only returns data while the Color page is active — on Edit/Cut/etc. it silently returns `None`, so the tool looked like the timeline had no thumbnails rather than naming the real constraint.

## The fix

- `_timeline_thumbnail_contact_sheet` now records the current page, switches to Color for the sampling loop (inside the existing `page_lock()` primitive so concurrent page switches stay serialized), and restores the user's page — alongside the existing timecode restore — in the `finally`.
- If the switch itself fails (headless, or `OpenPage` returns false), the per-frame error now names the requirement: `GetCurrentClipThumbnailImage requires the Color page and automatic switching failed` instead of the misleading generic message.
- `docs/SKILL.md` notes the Color-page requirement and that a brief page flash in the GUI is expected.

## Tests

`tests/test_thumbnail_contact_sheet_page.py` covers switch-and-restore, the already-on-Color no-op, and the failed-switch error message, with a fake Resolve that only serves thumbnails while on the Color page. Offline suite: 2273 tests, no new failures (the 4 pre-existing errors are `numpy`/`requests` missing from the venv).


samuelgursky (commented):
Reviewed on the branch. Good catch, and the diagnosis is confirmed by Resolve's own reference — `docs/reference/resolve_scripting_api.txt` documents `GetCurrentClipThumbnailImage` as returning data "for current media in the **Color Page**". So this is a documented constraint the code was ignoring rather than an undocumented trap, which makes the fix straightforwardly correct rather than speculative. Worth citing that line in the code comment.

## What checks out

- **The `page_lock()` usage is correct.** I checked `src/utils/page_lock.py`: it's a reentrant `RLock` with the `fcntl` file lock taken only at the outermost nesting level. So holding `_page_lock()` and calling raw `r.OpenPage(...)` inside is safe, and it stays safe even if something nested calls `open_page_serialized`. No deadlock.
- Restore is inside the lock and inside `finally`, alongside the existing timecode restore. Right structure.
- The new error message is a real improvement — naming the requirement instead of implying the timeline has no thumbnails.
- Tests: `tests.test_thumbnail_contact_sheet_page` 3/3 OK. Full offline suite 2320, which is exactly main's 2317 plus the three new tests, no new failures. Branch is current with main.

## Four other call sites have the same bug

`GetCurrentClipThumbnailImage` is called in five places and only one is fixed here:

- **`src/granular/timeline.py:940`** (`timeline_get_current_clip_thumbnail`) — returns a bare `{"success": False}` with no error message at all. This is the worst of the four: it's the exact silent failure this PR is fixing, still fully present in the 341-tool granular server.
- **`src/server.py:21199` and `:21208`** (`timeline_markers.get_thumbnail` / `get_thumbnail_image`) — these already *name* the constraint in their error text ("Open the Color page..."), so they're honest, but they ask the user to do by hand what this PR now does automatically. After this change one thumbnail path auto-switches and two adjacent ones tell you to switch yourself.
- **`src/utils/timeline_versioning.py:733`** — best-effort by design and its docstring already names the Color page as a cause, so leaving it is defensible. Worth noting though that version snapshots taken off the Color page will silently carry no thumbnail, and that path is the auto-archive-before-destructive-op hook.

The first three are a small refactor away from sharing the same switch-and-restore. Fixing the one that was reported while leaving four siblings is how this bug comes back.

## Two smaller things

**`GetCurrentPage()` returning `None` leaves the user on the Color page.** The bundled reference lists `None` as a possible return. If that happens (or `GetCurrentPage` raises), `original_page` stays `None`, `on_color` is `False`, the switch to Color goes ahead — and the restore is skipped, because it's guarded on `if r and original_page and original_page != "color"`. Same outcome if the handle is briefly unavailable at capture time. Falling back to a known page, or recording that the page couldn't be captured and surfacing it, would close that.

**No `api_truth` entry.** This is a textbook `silent-failure` entry — a documented constraint that produces an empty result with no error — and `api_truth` has the `mitigation` field for exactly this shape. Without an entry it won't reach the generated `api-limitations.md` or the agent-facing docs, so an agent hitting the granular tool still has nothing to go on. PR #107 adds entries for findings of this exact kind, so it'd be good to keep the batch consistent.

## Summary

The fix is correct and the tests are well-targeted. I'd like the granular call site covered before merge, since it's the same silent failure with a worse error, plus an `api_truth` entry. The `None`-page restore and the two `timeline_markers` sites are fine as follow-ups.

One note on the docs wording: "a brief page flash in the GUI is expected" undersells it slightly — switching to Color on a large timeline can also kick off cache and render work. Worth a word.

billcarroll:
Addressed in f2d3ef8:

- **Granular call site**: the switch-and-restore now lives in a shared `color_page_for_thumbnails` context manager in `src/utils/page_lock.py`, and both `_timeline_thumbnail_contact_sheet` and the granular `timeline_get_current_clip_thumbnail` use it. The granular tool's bare `{"success": False}` is replaced with the same Color-page-naming error.
- **`GetCurrentPage()` → `None`**: closed in the helper rather than deferred — it refuses to switch when the current page can't be captured, so the skipped restore can no longer strand anyone on Color. New test covers it (`test_uncapturable_page_never_switches_so_restore_cannot_strand_on_color`).
- **`api_truth` entry**: added with `submit: "missing"` and `mitigation: [color_page_for_thumbnails, _timeline_thumbnail_contact_sheet]`; `api-limitations.md` regenerated, drift guard passes.
- **Reference citation**: the helper docstring and the contact-sheet comment now cite the "for current media in the Color Page" line from `docs/reference/resolve_scripting_api.txt`.
- **Docs wording**: SKILL.md now says landing on Color can kick off cache/render work for the current clip, not just a page flash.

Left as follow-ups, as you suggested: the two `timeline_markers` sites (`get_thumbnail` / `get_thumbnail_image`) that still ask the user to switch by hand — with the helper extracted they're each a small change, but they alter two more tools' behavior so I'd rather land them separately — and the `timeline_versioning` best-effort path stays as-is by design.

Full offline suite: 2321 tests, 0 failures (your 2320 plus the new None-page test).

samuelgursky (commented):
This addresses everything I raised, and the shared `color_page_for_thumbnails` is a better shape than what I suggested — pulling it into `page_lock.py` next to the primitive it depends on means the granular server gets the fix by construction rather than by remembering to copy it.

Particularly good calls:

- **The `None`-page resolution.** Not attempting the switch when `GetCurrentPage()` can't be captured is the right trade. It means no thumbnails in that case rather than a user stranded on the Color page, and the docstring says so explicitly. Failing visibly beats succeeding somewhere the user didn't ask to be.
- **The granular site now has a real error message** instead of a bare `{"success": False}` — that was the worst of the four.
- The `api_truth` entry cites the reference line, and `mitigation` names both the helper and the caller, so the guard test has something to hold onto. `gen_api_limitations.py --check` comes back clean.

Verified: `tests.test_thumbnail_contact_sheet_page` 4/4, full suite 2321 = main's 2317 plus the four new tests, no new failures.

The two `timeline_markers` sites and `timeline_versioning` are still on the old path, which is what we agreed — they already name the constraint, so they're fine as follow-ups.

## One thing to pull out before merge

`look.cube` (a 4-byte file containing `LUT`) is committed at the repo root. It's a test artifact, not source.

Its provenance is worth knowing, because it explains a test failure you may have seen: `tests/test_color_grade_probe.py:200` calls `_safe_export_lut` with `os.path.join(os.getcwd(), "look.cube")` and asserts the export is *refused* because the path isn't temporary. If the suite runs from a directory the guard considers temporary, the guard permits it instead — the assertion fails **and** the export really writes `look.cube` into the working directory. Same root cause as `test_validate_render_settings_requires_temp_target_when_requested` failing in the same run.

So: `git rm look.cube` and it's clean. Nothing wrong with your change.

The test writing a real file into cwd on its failure path is a repo-side wart rather than anything to do with this PR — I'll deal with that separately.

Once `look.cube` is out this looks good to me.

billcarroll:
Removed in 47f7d2d — it was a test artifact from running the suite in my worktree, swept in by `git add -A`. Checked #108 and #111 for the same mistake; both are clean.

PR #111, opened 2026-08-03
docs(api-truth): four verified edit-session findings (DeleteClips, mixed-fps append, ImportMedia)

billcarroll:
Records four behaviors verified during a live cut-video edit session on DaVinci Resolve Studio 21.0, so agents can look them up instead of rediscovering them:

- **`Timeline.DeleteClips` flaky first attempt** — can return `False` with valid, present items; an identical immediate retry succeeds (same shape as `ProjectManager.DeleteProject`). Tagged `submit: bug`.
- **`Timeline.DeleteClips` leaves linked audio** — the API deletes exactly the items passed; the UI's linked-selection behavior does not apply, so orphaned audio collides with later appends.
- **`MediaPool.AppendToTimeline` mixed-fps duration floor** — source→timeline frame conversion rounds down, landing a planned range one frame short (e.g. 24.0/29.97 source in a 23.976 timeline).
- **`MediaPool.ImportMedia` current-folder only** — no destination parameter; imports land in the current bin.

SKILL.md gains matching guidance in the `delete_clips` and `append_to_timeline` sections, and `api-limitations.md` is regenerated (drift guard passes).

Full offline suite: 2321 tests, 0 failures.

### DIFF
--- a/docs/SKILL.md
+++ b/docs/SKILL.md
@@ -1307,7 +1307,12 @@ Key actions:
   instead of walking tracks by hand. Filters may be passed inline or as a
   `filters` dict; a mistyped filter name is rejected rather than silently
   matching everything. Returns `{clips, match_count, total_clips}`.
-- `delete_clips(clip_ids, ripple?)` — IDs are unique IDs from `get_items`
+- `delete_clips(clip_ids, ripple?)` — IDs are unique IDs from `get_items`.
+  Two verified quirks (see `api_truth`): the call can return `success: false`
+  on the first attempt with valid IDs — re-list and retry once before failing;
+  and deleting a video item does NOT delete its linked audio — pass the linked
+  audio item IDs explicitly, then `detect_gaps_overlaps` across both track
+  types.
 - `duplicate_clips(clip_ids?, selected?, target_track_index?, track_offset?, placement?, record_frame?, record_frame_offset?, copy_properties?, include_linked?)` —
   duplicate existing video timeline items by re-appending the same Media Pool
   item with the same source trim; `selected=True` uses Resolve's selected/current
@@ -1722,6 +1727,14 @@ media_pool(action="append_to_timeline", params={"clip_infos": [
   {"clip_id": "<uuid>", "start_frame": 0, "end_frame": 100, "record_frame": 1200, "track_index": 4}
 ]})
 ```
+
+Mixed-fps caution: `start_frame`/`end_frame` are SOURCE frames, and a source
+whose fps differs from the timeline's rounds DOWN on conversion — a 24.0 or
+29.97 clip appended into a 23.976 timeline can land one frame short of its
+slot. Plan durations in timeline frames, extend `end_frame` by a source frame
+when the floor misses, and finish with `detect_gaps_overlaps` (see
+`api_truth`). `import_media` always lands in the CURRENT bin — call
+`set_current_folder` first; there is no destination parameter.
 
 ### 4. Inspect and annotate timeline items
 
--- a/docs/reference/api-limitations.md
+++ b/docs/reference/api-limitations.md
@@ -12,7 +12,7 @@ that none exists).
 
 **Verified on:** DaVinci Resolve Studio 21.0.0
 
-**Totals:** 22 missing capabilities, 20 bugs / unreliable behaviors.
+**Totals:** 22 missing capabilities, 21 bugs / unreliable behaviors.
 
 The authoritative source is the runtime-queryable `api_truth` ledger
 (`resolve_control api_truth "<query>"`); this document is generated from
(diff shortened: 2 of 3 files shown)


samuelgursky (commented):
Four findings from real edit-session work, written up so the next agent doesn't rediscover them — this is what `api_truth` is for, and the linked-audio and mixed-fps entries in particular are the kind of thing that costs an afternoon the first time. The drift guard is clean (`gen_api_limitations.py --check` OK).

Three things worth resolving before it lands.

## The DeleteProject analogy argues against the recommendation

The flaky-DeleteClips entry says "Same failure shape as `ProjectManager.DeleteProject`" and then recommends "retry the identical call once".

But the existing DeleteProject entry, a few hundred lines up in the same file, records the opposite conclusion. Its flakiness is not random — it has a mechanism:

> Returns False (no deletion) when the target project is, or recently was, the current project [...] CloseProject reliably releases it: a delete that had failed six times in a row, a second apart, succeeded on the first attempt after one.

Six failures a second apart is precisely the case where "retry once" does not help. So invoking that analogy points a reader at a workaround the analogue disproves. Either the shapes really are the same — in which case there is probably a *state* precondition for DeleteClips still to be found, and the entry should say the cause is unknown rather than implying randomness — or they are not, and the comparison is better dropped. As written it reads as "we know what this is", and I don't think we do yet.

Worth capturing what the session actually saw: was the retry immediate, and had anything else touched the timeline between the two calls?

## Four unguarded call sites, no mitigation

If a first-attempt `False` is real, these are all exposed today, each a bare `bool()` with no readback or retry:

- `src/server.py:4167` — duplicate-and-delete source cleanup
- `src/server.py:4282` — overlap clearing before a duplicate
- `src/server.py:4375`
- `src/server.py:20709` — the `delete_clips` action itself

(`src/server.py:16734` is `MediaPool.DeleteClips`, a different method — not affected.)

The entry tells *callers* to verify and retry, but the tool is the caller for every agent using this server. Recording a flaky API and leaving four call sites bare puts the burden on everyone downstream. Either add the readback-and-retry-once behind `delete_clips` and name it in a `mitigation:` field — the pattern #107 uses — or state explicitly in `recommended` that the call sites are currently unguarded, so the gap is visible rather than implied.

## Two entries look like they want `submit` tags

Only the flaky entry carries `submit: "bug"`, so it is the only one of the four that reaches `api-limitations.md`. Two of the others look like they qualify:

- **`ImportMedia` has no destination parameter** — that is the textbook `submit: "missing"` case, and structurally identical to entries already tagged that way.
- **The mixed-fps duration floor** — a source→timeline conversion that rounds down and silently leaves a one-frame gap is a reportable off-by-one, not just a usage note.

The linked-audio one I'd leave untagged: the API deleting exactly what it was passed is defensible behavior, and the finding is that it differs from the UI, which is guidance rather than a defect.

## Minor

The branch is 4 commits behind main, which is why the suite shows two failures in `test_color_grade_probe` / `test_render_deliver_probe`. Those are the cwd-dependent guard tests fixed on main in 33b73f8 — a rebase clears them, and they are unrelated to this change.

Happy to see this land once the DeleteClips framing is settled; the other three entries I'd take as-is.

billcarroll:
Review addressed in f248334 (rebased onto main, conflict on the generated doc resolved by regenerating):

- **DeleteClips framing**: took the wording option — the `recommended` field now states explicitly that the MCP `delete_clips` tool does not yet implement the readback-and-retry and call sites are currently unguarded. Implementing the retry behind `delete_clips` (with a `mitigation:` entry, the #107 pattern) is queued as a follow-up code PR so this one stays docs-only.
- **Submit tags**: mixed-fps duration floor → `submit: "bug"`; ImportMedia destination gap → `submit: "missing"`; linked-audio left untagged as intended UI/API divergence.
- Regenerated `docs/reference/api-limitations.md` via `scripts/gen_api_limitations.py` (totals now 23 missing / 22 bugs); drift test and the full offline suite pass (2343 tests, the two cwd-dependent failures are gone post-rebase).

samuelgursky:
The submit tags and the unguarded-call-sites note are both exactly right, thank you. But I think point 1 got read as the same thing as point 2, and it isn't — so let me restate it more precisely, because it's the only thing left on this PR and it's one sentence of wording.

My concern was not *where* the retry is implemented. It was the **causal claim** in the `reality` field, which is unchanged:

> Can return False on the first call even when every item in the list is a valid, present TimelineItem; an identical immediate retry succeeds. **Same failure shape as ProjectManager.DeleteProject.**

Read the `DeleteProject` entry that cites — it's a few hundred lines up in the same file. Its flakiness is *not* random; it has a mechanism, and retrying is specifically what does **not** clear it:

> Returns False (no deletion) when the target project is, or recently was, the current project [...] CloseProject reliably releases it: a delete that had failed six times in a row, a second apart, succeeded on the first attempt after one.

So the entry currently says "this is the same shape as that one" and then recommends the thing that demonstrably didn't work for that one. Both cannot be right. Either:

- they really are the same shape — in which case there is probably a **state precondition** for `DeleteClips` still to be found, a single retry will clear it only sometimes, and the entry should say the cause is unknown rather than implying randomness; or
- they aren't — and the comparison is doing unearned work and is better dropped.

What I'd take is simply the smaller claim, e.g.:

> Can return False on the first call even when every item in the list is a valid, present TimelineItem; an identical immediate retry succeeded in the one observed instance. **Cause not established** — possibly a state precondition, as with `ProjectManager.DeleteProject`, whose superficially similar flakiness turned out to have a specific trigger that retrying does not clear.

That keeps your observation intact — it happened, the retry worked — and stops the entry asserting a mechanism nobody has evidence for.

This matters more now than when I first raised it, because #114 turns "retry once" from advice into **shipped behaviour**. If the flakiness does have a trigger, that helper will paper over it intermittently and silently fail the rest of the time, and the ledger will be the thing that told us it was fine. #114 itself is in good shape — I've just confirmed `932d7c7` fixes the blind-readback bug — so this wording is the only thread across both PRs.

No need to solve Resolve's behaviour. Just don't let the record claim more than the session saw.

samuelgursky:
Pushed the wording as `1e29a76` so this isn't blocked on a round trip — the substance is entirely yours, I've only removed a claim.

The entry now records what the session saw (a False, items confirmed present, one identical retry succeeding) and says the cause is not established. `DeleteProject` is still mentioned, but as the reason to be *suspicious* of retry-as-answer rather than as grounds for it. Revert or reword if you read the evidence differently — it's your observation.

`api-limitations.md` regenerated, drift check clean, suite 2343 OK. Merging this and #114 together now.

Thanks for the turnaround on the submit tags and the unguarded note — both were exactly right.

PR #113, opened 2026-08-03
fix(media-pool): honor list metadata keys, and name delete_timelines' expected param

billcarroll:
Two tool-surface bugs found during live cataloguing work:

**media_pool_item get_metadata with a list-valued `key` silently returned the full dict.** Resolve's keyed getters take one string; handed a list they ignore it and return everything — silent misbehavior for callers expecting a subset. A list key now returns exactly that subset (missing keys map to `null`), and an empty or non-string list is a clear error. The same guard covers the identical pattern in `get_third_party_metadata` and `get_clip_property`.

**media_pool delete_timelines called with `timeline_names` leaked a bare `KeyError('timeline_ids')`.** It now returns a proper error naming `timeline_ids` (a non-empty list of timeline unique IDs) and, when `timeline_names` was passed, explicitly says timelines are matched by unique ID, not name.

Offline tests in `tests/test_keyed_param_guards.py` (10 cases); full offline suite passes (2339 tests, 26 skipped). Branched from current main, independent of #107/#108/#110.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -16606,11 +16606,18 @@ def media_pool(action: str, params: Optional[Dict[str, Any]] = None) -> Dict[str
 
         return _run_maybe_background("media_pool.import_timeline", p, _work)
     elif action == "delete_timelines":
+        ids = p.get("timeline_ids")
+        if not isinstance(ids, list) or not ids:
+            hint = (" ('timeline_names' is not supported — timelines are matched"
+                    " by unique ID, e.g. from timeline.get_unique_id)"
+                    if "timeline_names" in p else "")
+            return _err("delete_timelines requires 'timeline_ids', a non-empty"
+                        " list of timeline unique IDs" + hint)
         count = proj.GetTimelineCount()
         timelines = []
         for i in range(1, count + 1):
             tl = proj.GetTimelineByIndex(i)
-            if tl and tl.GetUniqueId() in p["timeline_ids"]:
+            if tl and tl.GetUniqueId() in ids:
                 timelines.append(tl)
         if not timelines:
             return _err("No timelines found")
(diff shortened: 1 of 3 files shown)


billcarroll:
Live-validated against a running Resolve (read-only — no project mutation): list-key subset, scalar string-key path, empty-list rejection, and both `delete_timelines` param errors all behave as described. Probe committed as `tests/live_keyed_param_validation.py`.

samuelgursky (commented):
Both fixes are correct and the tests are well-chosen. Branched from current main, so the suite comes back genuinely clean here — 2339, no failures, no artifacts. Notes are all minor.

## `_keyed_get`

The diagnosis is right: Resolve's keyed getters take one string, and handed a list they ignore it and return everything, so the caller gets a superset it did not ask for and cannot tell. Subsetting locally is the only fix available — there is no batch getter — and routing all three getters through one helper is the right call, since the bug was identical in each.

Two small things:

**Missing keys report differently through the two paths.** A string key for an absent field returns Resolve's `""`; a list key returns `None`:

```
key="NoSuch"    -> ''
key=["NoSuch"]  -> {'NoSuch': None}
```

I think `None` is the better answer — it separates "absent" from "present but empty", which `""` cannot — but the divergence should be deliberate and written down, otherwise a caller migrating from string to list gets a different sentinel with no notice. A line in the docstring covers it.

**Only `get_metadata` documents the new list form.** `get_third_party_metadata` and `get_clip_property` gained exactly the same capability but their docstring lines are unchanged, so two-thirds of the surface is undiscoverable. Same one-line note on each.

## `delete_timelines`

Straightforwardly right, and the `timeline_names` hint is the kind of error that saves someone a round trip — it answers the question the caller was actually asking rather than just naming the missing key. `test_valid_ids_pass_validation` landing on "No timelines found" to prove validation was passed rather than skipped is a nice touch.

One consideration rather than a request: the repo has `_validate_params` (`src/utils/contracts.py`) for this shape of check. Hand-rolled is fine for a two-line guard, but if you touch this again, the contract helper gets you the standard error envelope for free.

## Verified

- Full offline suite on the branch: **2339 tests, OK**.
- The string-vs-list sentinel divergence above, reproduced directly.
- `_keyed_get` placement is fine — module-level, defined before first call at runtime.

Nothing here blocks merge. The two docstring lines would be good to have before it lands, since they are the difference between a feature being discoverable and not.

billcarroll:
Both docstring notes added in the latest push: all three keyed getters now document the list form, and each notes the deliberate sentinel divergence — list form maps missing keys to `null` (absent vs present-but-empty), string form returns Resolve's `""`. The `test_list_key_missing_field_is_none` case already pins the `null` behavior. Noted on `_validate_params` for the next touch.

PR #114, opened 2026-08-03
fix(timeline): DeleteClips readback-and-retry on flaky first attempt

billcarroll:
Stacked on #111 (branch includes its commits; merge #111 first and this PR reduces to the fix below).

## What

`Timeline.DeleteClips` can return False on the first call even when every item in the list is a valid, present TimelineItem — an identical immediate retry succeeds. It can also return False while actually deleting. PR #111 documented this in the api_truth ledger but left the MCP call sites unguarded.

This adds `_timeline_delete_clips_verified`: on a False return it re-lists the tracks (`_timeline_items_still_present`); if the target items are gone the delete is reported as success; if any survive, the identical call is retried once, with a final readback. Items that expose no usable unique ID are treated as unverifiable — still-present — so an unconfirmed delete is never reported as verified-gone.

All four timeline DeleteClips call sites route through the helper: the `delete_clips` action, `lift_range`, `duplicate_clips` (delete_sources), and `copy_range` (overwrite).

## Ledger

The 'Timeline.DeleteClips (flaky first attempt)' entry gains `mitigation: ["_timeline_delete_clips_verified", "_timeline_items_still_present"]` (the PR #107 pattern) and drops the now-false 'call sites are currently unguarded' sentence; `docs/reference/api-limitations.md` regenerated via `scripts/gen_api_limitations.py`.

## Tests

`tests/test_delete_clips_readback_retry.py` — 11 offline tests with a scripted-lie fake timeline: helper unit coverage (true first call, false-but-deleted, false-then-retry, both-noop failure, unverifiable IDs) plus dispatcher and `lift_range` integration. Full offline suite: 2354 tests, OK (26 skipped).

Live validation is destructive (deletes timeline items); not run in this session — the offline fake reproduces the observed lie shapes from the Studio 21.0 session behind #111.


samuelgursky (commented):
This is the right response to the review on #111 — a mitigation named in the ledger rather than advice to every downstream caller, and the `mitigation:` field wired up so the guard test has something to hold. Routing all four call sites through one helper is the right shape, and the scripted-lie fake is a good way to test a bug you can't reproduce on demand.

Two things need fixing before it lands, one of them serious, plus a framing question that this PR inherits from #111 and now makes load-bearing.

## Blocking: a blind readback turns a failed delete into a reported success

`_timeline_items_still_present` returns `False` — "they're gone" — whenever the readback finds nothing, and it cannot distinguish "nothing found because the delete worked" from "nothing found because the readback couldn't see". Both paths swallow errors:

```python
try:
    track_items = tl.GetItemListInTrack(track_type, index) or []
except Exception:
    track_items = []
```

Run against a timeline where the delete genuinely fails and the readback is blind:

```
readback empty   -> delete_clips reports success=True  (item was NEVER deleted)
readback raises  -> delete_clips reports success=True  (item was NEVER deleted)
```

Both cases return `True` from `_timeline_delete_clips_verified` after a `DeleteClips` that deleted nothing and returned `False` throughout. The old code returned `bool(tl.DeleteClips(...))` — an honest `False`. So on this path the change **introduces** the silent-lie class the whole series is fixing, in a destructive action, and `lift_range` and `copy_range` inherit it.

The asymmetry is already right in the ID branch — no usable ID means "still present", i.e. don't claim success. The same conservatism needs to apply to the readback: a walk that raised, or that could not enumerate any track, is *unknown*, not *empty*. Something like a tri-state (`present` / `absent` / `unknown`) with `unknown` never counting as verified-gone would close it, since `_process_is_alive` in the bridge just landed on exactly that shape for the same reason.

## Unverifiable items are deleted twice

The docstring says an unverifiable delete "must not be reported as verified-gone", which is correct for *reporting*. But in the control flow `still_present == True` also means *retry the destructive call*, so the conservative reporting choice becomes an aggressive acting choice:

```
unverifiable items  -> result=False, DeleteClips called 2 time(s)
```

For items whose IDs can't be read, the helper can never confirm anything, so the second call is guaranteed to add no information — it's a destructive operation performed purely because verification was impossible. Worth separating "should I retry?" from "may I claim success?" rather than deriving both from one boolean.

Related and worth a sentence in the docstring: with `ripple=True` the retry is not idempotent. If the first call deleted some items and left others, the second passes the original list — including handles to items that are already gone — back into a rippling delete. I could not make that misbehave in a fake, and it may well be fine, but the reasoning should be written down in a confirm-gated path that renders timeline changes.

## The framing question this now makes load-bearing

My #111 note about the `DeleteProject` analogy is still open, and it matters more here, because "retry once" has gone from advice to shipped behavior.

The existing `DeleteProject` entry records that its flakiness has a *mechanism* — the project being, or recently having been, current — and that retrying does **not** clear it: "a delete that had failed six times in a row, a second apart, succeeded on the first attempt after one [CloseProject]". If `DeleteClips` really is the same shape, then a single retry is the wrong mitigation and will paper over the problem intermittently while silently failing the rest of the time. If it isn't the same shape, the entry shouldn't invoke the comparison.

Not asking you to solve Resolve's behaviour — just to avoid building product behaviour on a causal model that the cited precedent argues against. Recording "cause unknown; a single retry cleared it in the one observed instance" would be honest and would still justify this helper.

## Verified

- Full offline suite on the branch: **2354 tests, OK**.
- `_timeline_track_count` and `_safe_timeline_item_id` both exist; the helper is correctly placed.
- The two failure modes above, reproduced directly against the branch.
- The stacking is handled well — dropping the now-false "call sites are currently unguarded" sentence from the entry in the same PR that makes it false is exactly right, and easy to forget.

The core idea is sound and I'd like it in. The blind-readback path is the one I'd hold the merge for.

samuelgursky (commented):
`932d7c7` fixes it. Re-running the two reproductions from my review against the branch:

| case | before | now |
|---|---|---|
| readback raises | `success=True`, item never deleted | **`success=False`** |
| unverifiable ID | `success=False`, `DeleteClips` called **twice** | `success=False`, called **once** |
| genuine flaky (false → retry succeeds) | works | still works, 2 calls |

Full suite on the branch: **2366 tests, OK**.

The tri-state is the right shape, and separating "may I claim success?" from "should I retry?" is exactly the distinction that was missing — `unknown` now reports failure *and* declines the second destructive call, so the conservative choice no longer buys a delete with no information. "A sighting is definitive even if another track failed" is a nice touch: it keeps a positive finding usable when the walk is only partly readable.

**One correction to my own review.** My third case — a completed walk that enumerates tracks fine and finds nothing — still returns `absent` → success, and I want to withdraw that as a defect. My fake described an incoherent scenario: an item that was never on any track to begin with. A finished walk across every track that doesn't see the item *is* positive evidence of absence, and reporting the delete's goal as satisfied is defensible. The real bugs were the two above, and both are fixed.

Recording the `ripple=True` non-idempotence in the docstring rather than claiming to have solved it is the right call — "that could not be made to misbehave against a fake; it is recorded, not resolved" is precisely the kind of thing that should be written down rather than quietly assumed.

The only thing still open on this PR is the framing question it inherits from #111, which I've followed up on over there — it isn't code, and it doesn't need to hold this up if the wording lands first.

samuelgursky:
Rebased onto main and merging. #111 has landed, so this now reduces to the two code commits as you intended.

**A correction I owe you.** I pushed a wording fix for the `DeleteClips` framing to #111 (`1e29a76`) before spotting that you had already fixed it yourself in `932d7c7` — I reviewed that commit's *code* and never opened its `api_truth` diff, then told you on #111 that the point was still outstanding. It wasn't, and I was wrong to say so.

Your version is also the better one, so I dropped mine in the rebase conflict and kept yours:

> Cause unknown — do NOT read this as the ProjectManager.DeleteProject shape: that one has an identified mechanism (the project being, or recently having been, current) that retrying does not clear, whereas a single retry cleared this in the one instance seen. One observation is not a mechanism; if a retry is ever seen to fail repeatedly here, this entry needs revisiting.

That names DeleteProject's actual mechanism rather than gesturing at it, and adds a falsification condition mine didn't have — "if a retry is ever seen to fail repeatedly, this entry needs revisiting" is the sentence that stops the record calcifying. Mine just said the cause was unestablished.

Rebase notes, since I did it rather than handing it back:

- Both conflicts were in `docs/reference/api-limitations.md`, which is generated — resolved by running `scripts/gen_api_limitations.py` rather than hand-merging.
- The one real conflict was in `api_truth.py`, between my #111 wording and yours here. Yours kept, entirely.

Verified after the rebase: `gen_api_limitations --check` OK, `audit_api_parity` PASS, suite **2376 tests OK**, and both reproductions from my review still behave — `readback raises -> success=False, calls=1` and `unverifiable id -> success=False, calls=1`.

PR #115, opened 2026-08-03
fix(rough-cut): reconcile skill with the api_truth ledger (import order, Fusion render claim)

bolnet:
Two rows of `.claude/skills/rough-cut.md` contradict the `api_truth` ledger the skill itself points at. Both were found while following the skill against Resolve Studio 21.0.3.7 on MCP v2.71.1.

### 1. Import order is backwards

Step 4 read:

> **Import into a bin named for the episode**, then `media_pool set_current_folder` to it.

`ImportMedia` has no destination parameter and always lands in the **current** folder, so importing first puts the clips wherever the current folder happens to be. The ledger records this as `MediaPool.ImportMedia (current-folder destination only)`, added in v2.71.1 — the new entry invalidates the older skill text.

Reordered to: create bin → `set_current_folder` → import, with a note to restore the previous current folder if it matters.

### 2. The Fusion row states a claim the ledger has retracted

The traps table still asserts a comp attached to a media clip *"never renders"*. The ledger entry was corrected on 2026-08-02 and explicitly calls the blanket form **too broad**: a comp wired `MediaIn → Blur → MediaOut` **does** render, and an unrooted `MediaOut` does not get silently bypassed — it fails the render job outright, which is a different and more dangerous failure than "no effect".

Replaced with the corrected finding, keeping the actionable guidance (use Fusion titles as their own timeline clip for text over picture).

### 3. Version stamps

The table header carried a date (`2026-07-28`) but no build number. The scripting API changes per patch release, so a date alone isn't reproducible. Each row now names the build it was confirmed on, and the header says to re-confirm rows older than your build.

Also adds two rows already in the v2.71.1 ledger but missing from the narrative table:
- **mixed-fps duration floor** — `start/endFrame` are source frames; source fps ≠ timeline fps floors the conversion and the range lands a frame short
- **`ImportMedia` current-folder-only**

### 4. Stale-ledger footgun

Added a note that a **running** MCP process keeps executing the version it started with, so `git pull` does not refresh the ledger until restart. Concretely: on v2.70.4, `api_truth "DeleteClips"` returns zero facts that v2.71.1 carries — which reads as "no known issues" rather than "stale server".

Docs only; no code paths touched.

### DIFF
--- a/.claude/skills/rough-cut.md
+++ b/.claude/skills/rough-cut.md
@@ -43,10 +43,14 @@ Render an mp4 only to preview the cut, and say that is what it is.
    Do not defer this to the end or write it off as cosmetic; a mismatch has been
    reported to affect playback and output, not just the display. On an existing
    project, read both back and only raise it if they are wrong.
-4. **Import into a bin named for the episode**, then
-   `media_pool set_current_folder` to it. One bin per episode, not one shared
-   dump — it keeps `analyze_bin` scoped and the Media Pool readable across a
-   long series.
+4. **Create a bin named for the episode, `media_pool set_current_folder` to it,
+   THEN import.** That order is mandatory, not stylistic: `ImportMedia` has no
+   destination parameter and always lands in the *current* folder, so importing
+   first puts the clips wherever the current folder happens to be — see the
+   `MediaPool.ImportMedia (current-folder destination only)` entry in the
+   `api_truth` ledger. Restore the previous current folder afterwards if it
+   matters. One bin per episode, not one shared dump — it keeps `analyze_bin`
+   scoped and the Media Pool readable across a long series.
 5. **Analyse** — `media_analysis analyze_bin`, `sampling_mode="adaptive_capped"`.
    Complete the `commit_vision` loop for every clip; leaving one in
    `pending_host_vision_analysis` is a failure, not a partial success (AGENTS.md).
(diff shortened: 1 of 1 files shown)


samuelgursky (commented):
Welcome, and thanks — reconciling a skill against the ledger it points at is exactly the kind of drift nobody notices until it produces a wrong result. I checked both substantive claims against `api_truth` and both hold.

**Import order.** Correct, and the ledger's own `recommended` field says the same thing in the other direction: *"SetCurrentFolder to the target bin first (media_pool set_current_folder), import, then restore the previous current folder if it matters."* The skill was telling people to do it backwards.

**Fusion row.** Also correct. The entry was corrected on 2026-08-02 and calls the blanket form too broad in those words, and your row captures the part that actually matters — an unrooted `MediaOut` *fails the render job* (887-byte file, job status `Failed`) rather than being silently bypassed. That's a materially more dangerous failure than "no effect", and the old row would have sent someone hunting for a missing effect instead of a failed render.

**Nice catch on the step reference.** Changing "step 2" to "step 3" fixes a pre-existing off-by-one — step 3 is the frame-rate setup step, step 2 is the add-a-timeline-to-an-existing-project rule. That wasn't in your description and it's a real correction.

**The stale-ledger warning is the most valuable thing here.** A running MCP process executing an old ledger while `git pull` says otherwise, and an `api_truth` miss reading as "no known issues" rather than "stale server", is a trap that would cost someone a long time. Good instinct to write it down.

## One thing to fix: four of the build stamps assert provenance that doesn't exist

This is worth getting right precisely because your reasoning for adding stamps is sound. A date alone isn't reproducible — agreed. But a build number that was never recorded is worse than a vague date, because it *looks* authoritative and a future reader will trust it rather than re-confirm.

Checking each row against the ledger:

| Row | Ledger provenance | Your stamp |
|---|---|---|
| `end_frame` exclusive | "Verified by readback probe on **Resolve Studio 21.0**" | Studio 21.0 ✅ |
| Mixed-fps duration floor | Studio 21.0 (v2.71.1) | Studio 21.0 ✅ |
| `ImportMedia` destination | Studio 21.0 (v2.71.1) | Studio 21.0 ✅ |
| Fusion comp | "**Studio 19.1.3.7**, corrected 2026-08-02" | 19.1.3.7 ✅ |
| `create_timeline_from_clips` | "Resolve Studio in PR #99" — **no build** | Studio 21.0 ❓ |
| `timelinePlaybackFrameRate` | "against Resolve Studio" (PR #99) — **no build** | Studio 21.0.2 ❓ |
| `rotation` flag | **not in the ledger at all** | Studio 21.0 ❓ |
| VFR | **not in the ledger at all** | Studio 21.0 ❓ |

The four you verified against ledger entries are exactly right. The other four come from PR #99, which recorded only "against Resolve Studio" and never named a build — and two of them aren't in the ledger at all. `timelinePlaybackFrameRate` picking up `21.0.2` looks like it came from the module-level `VERIFIED_ON`, which is the ledger's overall verification build rather than that entry's provenance.

If you re-confirmed any of these yourself on 21.0.3.7 — which your description implies you were running — then stamp them **21.0.3.7** and say so; that's new information and strictly better than what's there. For any you didn't, something like `Studio (build unrecorded, PR #99)` is honest, still actionable, and has the useful side effect of marking exactly which rows are worth re-confirming first.

Either way the table ends up more trustworthy than the date it replaced, which is what you were after.

## State

Docs-only, no code paths touched, and `.claude/skills/` is hand-authored rather than generated, so nothing regenerates from it. Suite 2376 OK, agent-rules in sync.

The branch is 2 commits behind main — worth a rebase, since the mixed-fps and `ImportMedia` ledger entries you cite landed in v2.71.1 and the current release is v2.72.0.

Fix the four stamps and I'll merge this.

samuelgursky:
Merged as 0ab2243, shipped in [v2.81.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.81.0). Landed by cherry-pick rather than the merge button — the file moved to `.claude/skills/resolve-rough-cut/SKILL.md` in 0a318a5 after you opened this, so the branch went stale on a rename. Git followed it cleanly on rebase and the commit keeps your authorship; GitHub will show this as closed.

I checked all three claims against the ledger before taking them and they hold:

- `MediaPool.ImportMedia (current-folder destination only)` — api-limitations.md:210. Step 4 on main still had the order backwards.
- The Fusion correction — api-limitations.md:329 explicitly calls the blanket "never renders" form too broad, and records that an unrooted `MediaOut` *fails* the render job rather than being bypassed. That's a materially more dangerous failure than the one the table described, so replacing the row mattered more than the wording change it looks like.
- The mixed-fps duration floor — api-limitations.md:413.

The stale-ledger note is the part I'd have missed. A running MCP executing the version it started with means `api_truth` can return a confident empty answer that reads as "no known issues" when it actually means "stale server" — and the `DeleteClips` example makes it concrete rather than theoretical. That failure mode applies well beyond this skill.

Per-row build stamps over a single header date is also the right call; the API does change per patch release, and the header date was quietly inviting people to trust rows across builds.

PR #117, opened 2026-08-05
fix(timeline): DeleteClips silently no-ops off the Edit page — switch pages for the call

billcarroll:
## The defect

`Timeline.DeleteClips` is page-gated: with the UI on the Fairlight page it deterministically returns False and deletes nothing, no matter how many times it is retried. The readback-and-retry helper from #114 handles the *flaky* False correctly, but against this failure it could only fail honestly — retrying cannot clear a page gate — so every delete-capable tool action failed whenever the user had left the UI on another page. A Fairlight or Color session is a completely ordinary place for them to be.

**Live verification (2026-08-04, Studio 21.0):** three identical retries against 132 valid, unlocked, present TimelineItems on the Fairlight page all returned False with all 132 still on the track; a single `Resolve.OpenPage("edit")` followed by the same call returned True and left 0 items. Track lock and enable state were confirmed clear before and after — a page gate, not a lock.

## The fix

`_timeline_delete_clips_verified` gains a keyword-only `resolve=None`. When a handle is supplied, the guard switches to the Edit page for the call and restores the caller's page in a `finally`; guard failures are swallowed, so the delete attempt itself stays the source of truth. All delete-capable paths thread the handle through — `timeline` `delete_clips` / `move_clips` / `overwrite_range` / `lift_range` / `apply_cuts`, and `edit_engine` `execute_swap`'s video and linked-audio lifts.

The handle is an explicit parameter rather than an internal `get_resolve()` deliberately: offline tests call this helper directly, and on a machine where Resolve is live, an internal connect would flip the user's UI page mid-test-run. With `None` the helper behaves exactly as before this change.

## The ledger

The `api_truth` DeleteClips entry is restructured into the two distinct failures now known to share this call:

1. **Wrong page** — deterministic, mechanism identified, verified as above. This is the case the previous entry's falsification condition anticipated ("if a retry is ever seen to fail repeatedly here, this entry needs revisiting") — revisiting it found the mechanism.
2. **Flaky first attempt** — unchanged in substance: one observation, cause still not established, and the entry now notes that the page state at the time was not recorded, so it cannot be ruled in or out as failure 1 in disguise. Its falsification condition is retained, page-qualified.

`api-limitations.md` regenerated from the ledger; drift check OK.

## Verification

- 7 new tests in `tests/test_delete_clips_readback_retry.py` (PageGuardTest + a dispatcher threading test): switch-to-edit happens before the delete and the page is restored after, including on a failed delete; already-on-edit touches no pages; a raising guard never blocks the delete; `OpenPage` refusing means no restore of a page never left; `resolve=None` is byte-for-byte the pre-guard behavior.
- Full offline suite on this branch (2.77.0 base): **2406 tests, OK (26 skipped)**.
- `gen_api_limitations.py --check` OK.
- No fresh live delete was run for this PR — the mechanism and the `OpenPage("edit")` remedy were verified live on 2026-08-04 as described above, and a repeat would destructively delete real timeline items.

### DIFF
--- a/docs/reference/api-limitations.md
+++ b/docs/reference/api-limitations.md
@@ -374,13 +374,17 @@ values, or automation-hostile modal prompts.
 - **Reference:** [issue #77](https://github.com/samuelgursky/davinci-resolve-mcp/issues/77)
 - **Tags:** unreliable-return, silent-failure, metadata, reel-name
 
-### Timeline.DeleteClips (flaky first attempt)
+### Timeline.DeleteClips (requires the Edit page; flaky first attempt)
 
 - **Object:** `Timeline`
 - **Signature:** `([TimelineItem], ripple) -> bool`
-- **Behavior:** Can return False on the first call even when every item in the list is a valid, present TimelineItem; an identical immediate retry succeeds. Observed once, on Studio 21.0 during a cut-video edit session (items confirmed still present after the False, deleted cleanly on retry). Cause unknown — do NOT read this as the ProjectManager.DeleteProject shape: that one has an identified mechanism (the project being, or recently having been, current) that retrying does not clear, whereas a single retry cleared this in the one instance seen. One observation is not a mechanism; if a retry is ever seen to fail repeatedly here, this entry needs revisiting.
-- **Workaround / current handling:** Treat a False return as advisory: re-list the track and check whether the items are actually gone; if still present, retry the identical call once before failing. A readback that raised, enumerated nothing, or covered items whose unique ID cannot be read is UNKNOWN, not gone — never report an unverifiable delete as success, and do not spend a second destructive call on an outcome you equally cannot read.
-- **Tags:** unreliable-return, flaky, timeline, edit
+- **Behavior:** Two distinct failures share this call.
+
+  **1. Wrong page (deterministic, has a mechanism).** With the UI on the Fairlight page, DeleteClips returns False and deletes nothing, no matter how many times it is retried. Verified 2026-08-04 on Studio 21.0: three identical retries against 132 valid, unlocked, present TimelineItems all returned False with all 132 still on the track; a single `Resolve.OpenPage("edit")` followed by the same call returned True and left 0 items. Track lock and enable state were confirmed clear before and after, so this is a page gate, not a lock. This is the case the previous entry's falsification condition anticipated — a retry seen to fail repeatedly — and revisiting it found the mechanism.
+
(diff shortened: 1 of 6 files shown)


samuelgursky:
Thanks — the diagnosis here is excellent, and the live evidence (three identical retries on Fairlight leaving all 132 items, then one `OpenPage("edit")` clearing them) nails the mechanism. Merging with two changes pushed to this branch in ddcef92.

**1. The switch needed to be serialized.** The guard was a bare `OpenPage` pair at the callsite, outside `src/utils/page_lock.py`. That module exists for exactly this case and is explicit about it: Resolve has one globally-active page, so an unserialized switch-work-restore races every other page-switching operation — `color_page_for_thumbnails` takes the Color page the same way. Under the threaded dispatch added in v2.62.0, a concurrent thumbnail capture can flip to Color mid-delete, and the delete then lands on the Color page and deterministically fails. That reintroduces this exact bug through the back door whenever two operations overlap. (`server.py` was already importing `page_lock` and `open_page_serialized` — nothing had used them.)

The guard is now `edit_page_for_timeline_edits`, a sibling of `color_page_for_thumbnails` in `page_lock`, same shape and same restore-only-a-page-we-actually-left rule. Your `resolve=None` design is untouched and still the right call.

**2. `apply_cuts` was flipping the page twice per cut.** The guard sat inside the per-cut helper while the loop calls it N times. Measured with a fake:

```
cuts: 5   OpenPage calls: 10   ['edit','fairlight','edit','fairlight', ...]
```

From Fairlight, 50 cuts is 100 page switches — each a visible flash, and `page_lock`'s own docstring notes a switch can kick off cache/render work. The guard is now held once around the loop; the per-delete guard nests inside for free, since `page_lock` is reentrant and an inner guard finds the page already on Edit.

**Everything else was verified as-is and needed no change.** All delete paths are threaded correctly — `copy_clips` and `copy_range` genuinely can't reach the delete (gated on `delete_sources` / `overwrite`), and `execute_swap`'s `_r` comes from `_project_context(need_resolve=True)`, so it's a real handle rather than `None`. The restore logic was already right: `previous_page` only got set when `OpenPage("edit")` actually succeeded, so it could never restore a page it never left. The `api_truth` restructure into two distinct failures is a genuine improvement, and noting that the flaky observation's page state wasn't recorded — so it can't be ruled in or out as failure 1 in disguise — is the honest call.

All 7 of your guard tests pass unchanged against the refactor. 4 added: the delete runs holding the lock (depth 1, not 0), nested guards switch once rather than once per delete, the unheld loop pays 2N as the counter-example, and the lock is released with the user's page restored when the delete raises.

Offline suite 2429 passed. Static checks, drift guards, and `gen_api_limitations --check` clean. Still no fresh live delete, for the reason you gave — a repeat would destructively delete real timeline items.

Shipping as v2.78.1.

Issue #26, opened 2026-04-22
Windows + Resolve 20.3 + multiple Python installs: server crashes (0xC0000005) on import; fix is PYTHONHOME in .mcp.json env


Dylanyz:
## Summary

On Windows with multiple Python versions installed, the `davinci-resolve` MCP server fails to connect because `import DaVinciResolveScript` segfaults (`0xC0000005` / exit `-1073741819`) inside `PyInit_fusionscript`. Root cause is **not** the MCP server — Resolve scans the Windows registry at scripting-init time and binds against the **highest registered Python version**, regardless of which Python actually launched the process. If that highest version is incompatible (3.13 / 3.14 with Resolve 20.x), the C extension load access-violates.

The fix is non-invasive: set `PYTHONHOME` in the `.mcp.json` env block so Resolve binds against the venv's Python only, scoped to the MCP server's child process. No registry edits, no Python uninstalls, no global env vars.

## Environment

- Windows 11
- DaVinci Resolve **Studio 20.3.1 build 6**
- `External scripting using = Local`
- Python versions on system: 3.11.9, 3.12.10, 3.13.12, 3.14.3 (all from python.org / winget, registered under `HKCU\SOFTWARE\Python\PythonCore\`)
- davinci-resolve-mcp v2.1.0
- Claude Code MCP client

## Repro

1. Have Python 3.14 (or 3.13) installed and registered in `HKCU\SOFTWARE\Python\PythonCore\3.14`
2. `python install.py --clients claude-code` → installer succeeds, venv created with Python 3.12
3. `claude mcp list` → `davinci-resolve: ✗ Failed to connect`
4. Manual import test in the venv: segfaults with `Exit code 139` / `0xC0000005`

The crash reproduces with Blackmagic's own bundled example `python_get_resolve.py` from `C:\ProgramData\Blackmagic Design\DaVinci Resolve\Support\Developer\Scripting\Examples\`, confirming it's an upstream Resolve+Python interaction, not MCP-server logic. But MCP users hit it as a "server won't connect" failure with no obvious diagnostic.

## Diagnostic clues that point at this root cause

- `ctypes.CDLL("fusionscript.dll")` loads cleanly — so the DLL is intact and dependencies resolve
- The crash happens specifically inside Python's extension loader (`PyInit_fusionscript`)
[...]

PR #118, opened 2026-08-05
fix: scrub PYTHONHOME/PYTHONPATH before spawning the whisper CLI subprocess

staahlarkitektur:
## Problem

`.mcp.json` sets `PYTHONHOME`/`PYTHONPATH` for this server's own process so it can `import DaVinciResolveScript` against Resolve's bundled Python (3.10 in my setup). `_run_command` spawns the `whisper` CLI via `subprocess.run(...)` with no explicit `env=`, so the child inherits that override.

`whisper.exe` is itself a Python program — often a different interpreter/version than the server's own (in my case: whisper installed under Python 3.14's user site-packages). A Python 3.14 interpreter inheriting a Python 3.10 `PYTHONHOME` tries to load 3.10's stdlib against its own compiled C extensions and crashes almost immediately:

```
AssertionError: SRE module mismatch
```

This crash is fast (well under a minute) — on its own it isn't a hang, though it can look like one if something else delays the response (see the two other PRs I'm opening alongside this one).

## Fix

- `_run_command` gains an optional `env: Optional[Dict[str, str]] = None` parameter, passed through to `subprocess.run(..., env=env)`. Default `None` preserves the existing inherit-everything behavior for every other call site.
- `_transcribe_with_whisper_cli` builds a scrubbed env for the `whisper` subprocess specifically: copies `os.environ`, drops `PYTHONHOME`/`PYTHONPATH`, and sets `PYTHONIOENCODING=utf-8` (whisper's own `--help` text includes CJK characters that crash on a non-UTF-8 Windows console).

## Testing

Added `RunCommandEnvOverrideTests` (2 tests) verifying `_run_command` correctly passes through a custom `env=` and doesn't leak parent-only variables, while preserving inherit-all behavior when `env=` is omitted. Confirmed independently: running `python -m whisper <file>` manually outside the server's environment transcribes correctly.

### DIFF
--- a/src/utils/media_analysis.py
+++ b/src/utils/media_analysis.py
@@ -2376,13 +2376,18 @@ def build_plan(
     }
 
 
-def _run_command(args: List[str], timeout: int = COMMAND_TIMEOUT_SECONDS) -> Tuple[int, str, str]:
+def _run_command(
+    args: List[str],
+    timeout: int = COMMAND_TIMEOUT_SECONDS,
+    env: Optional[Dict[str, str]] = None,
+) -> Tuple[int, str, str]:
     try:
         proc = subprocess.run(
             args,
             capture_output=True,
             timeout=timeout,
             check=False,
+            env=env,
         )
     except subprocess.TimeoutExpired as exc:
         stdout = exc.stdout.decode("utf-8", errors="replace") if exc.stdout else ""
@@ -4023,7 +4028,21 @@ def _transcribe_with_whisper_cli(path: str, artifacts: Dict[str, Any], transcrip
     ]
     if transcription.get("language"):
         cmd.extend(["--language", str(transcription["language"])])
-    code, _, stderr = _run_command(cmd, timeout=int(transcription.get("timeout", 1800)))
+    # PYTHONHOME/PYTHONPATH are set in this server's own environment to point
+    # at Resolve's bundled Python (for DaVinciResolveScript) — inherited by a
+    # child process, they corrupt a *different* Python interpreter's own
+    # stdlib resolution. whisper's CLI is itself a Python program (frequently
+    # a different interpreter/venv than this server's), so it must not
+    # inherit these. Confirmed by hang: whisper_cli silently stalled
+    # (ffmpeg child pegged at 0% CPU, never exiting) under the inherited
+    # env, and completed normally in under 5 minutes once these were
+    # stripped. PYTHONIOENCODING=utf-8 avoids a separate UnicodeEncodeError
+    # crash in whisper's own argparse help text on non-UTF-8 consoles.
+    whisper_env = dict(os.environ)
+    whisper_env.pop("PYTHONHOME", None)
+    whisper_env.pop("PYTHONPATH", None)
+    whisper_env["PYTHONIOENCODING"] = "utf-8"
+    code, _, stderr = _run_command(cmd, timeout=int(transcription.get("timeout", 1800)), env=whisper_env)
     if code != 0:
         return {"success": False, "backend": "whisper_cli", "error": stderr.strip() or "whisper CLI failed"}
     json_files = sorted(Path(work_dir).glob("*.json"), key=lambda p: p.stat().st_mtime, reverse=True)
--- a/tests/test_media_analysis.py
+++ b/tests/test_media_analysis.py
@@ -4,6 +4,7 @@ import os
 import shutil
 import sqlite3
 import subprocess
+import sys
 import tempfile
 import unittest
 import unittest.mock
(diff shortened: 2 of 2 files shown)


samuelgursky (commented):
Reviewed against `main` @ v2.79.2. Merges cleanly; full offline suite passes (2438, exactly +2 for the new tests).

**The diagnosis is right, and the bug is broader than the PR claims.** This isn't a quirk of your `.mcp.json` — it's the shipped Windows configuration:

- `install.py:476` writes `PYTHONHOME` into installer-generated client configs on Windows (documented at `docs/install.md:199`, from #26).
- `src/server.py:792-793` sets `os.environ["PYTHONHOME"] = sys.base_prefix` on Windows whenever it isn't already set — so the server poisons its own environment even when the config doesn't.

Every Windows install therefore hands a foreign `PYTHONHOME` to every subprocess it spawns. Thanks for chasing this down; it's exactly the kind of thing that's invisible from a Mac.

### Blocking: the code comment contradicts the PR body

The comment you're adding says:

> Confirmed by hang: whisper_cli silently stalled (ffmpeg child pegged at 0% CPU, never exiting) under the inherited env, and completed normally in under 5 minutes once these were stripped.

The PR body says the opposite, and I think the body is the one that's correct:

> This crash is fast (well under a minute) — on its own it isn't a hang.

The committed comment records a diagnosis you retracted, and pins the hang on the wrong cause — that hang is #120's shim problem, not this one. This repo keeps an api_truth ledger precisely because a confident-but-wrong explanation left in the tree outlives the person who knows better. Please rewrite it to the finding you actually landed on: a foreign `PYTHONHOME` makes a *different* Python interpreter load a mismatched stdlib against its own C extensions and die fast (`AssertionError: SRE module mismatch`).

### Design: the scrub belongs one level up

The poison is process-wide, but the scrub is at a single call site. `_run_command` is shared with `src/utils/silence_ripple.py` and `src/utils/deep_vision.py`; those spawn ffmpeg/ffprobe so they're fine today, but the next Python-based tool anyone wires in walks into the same trap with no sign of it. A named helper — `_child_env_for_non_resolve_child()` — would close the class rather than the instance, and give the next reader somewhere to hang the explanation.

Not blocking on that if you'd rather keep the diff tight.

### Merge order matters — please read alongside #120

#120 rewrites `_run_command` and **drops the `env=` parameter this PR adds**. If this lands first and #120's conflict is resolved by taking its side wholesale, `_transcribe_with_whisper_cli`'s `env=whisper_env` call raises `TypeError` at runtime. Your `RunCommandEnvOverrideTests` does catch it — provided both test sets survive the tests-file conflict (all three of your PRs append at the identical spot, so they collide there too).

I've integrated the two locally and verified the combination; details in the review on #120.

### Also

`RunCommandEnvOverrideTests`'s docstring cites `davinci-resolve-mcp CONTRIBUTION_NOTES.md, Bug 1`. That file isn't in this repo — same in #119 and #120, six references total. The comment needs to carry the finding itself.

Tests are good: real subprocess, both directions (explicit env and the inherit-all default), no mocking theatre.

samuelgursky:
Merged, and shipped in [v2.80.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.80.0). Thank you.

You were more right than the PR claimed. I went looking to confirm this was your `.mcp.json` and found it's the shipped configuration:

- `install.py:476` writes `PYTHONHOME` into installer-generated client configs on Windows — documented at `docs/install.md:199`, added for #26.
- `src/server.py:792` sets `os.environ["PYTHONHOME"] = sys.base_prefix` on Windows whenever it isn't already set, so the server poisons its own environment even when the config doesn't.

So every Windows install hands a foreign `PYTHONHOME` to every subprocess it spawns. That's now recorded in the comment at the call site, along with the note that any Python-based tool added there needs the same scrub — the fix is at one call site because whisper is the only Python child today, and I'd rather the next person adding one reads it than rediscovers it.

Two things changed on the way in.

The merge conflict with #120 was the one I flagged, and it was real: #120's rewrite dropped the `env=` parameter, so resolving in its favour would have left `_transcribe_with_whisper_cli` calling `env=whisper_env` against a signature that no longer accepted it — a `TypeError` at transcription time, Windows only, on a path that already looked like it hung. Resolved by hand; the tree-killing `Popen` body keeps `env=`. Your `RunCommandEnvOverrideTests` would have caught it either way, which is a good argument for the tests you wrote.

I also rewrote the inline comment. It said the inherited env caused a silent stall with an ffmpeg child pegged at 0% CPU — but your PR description retracted that ("this crash is fast... on its own it isn't a hang"), and that stall was #120's shim problem. I trust the description over the comment, so the comment now records the fast, specific failure: `AssertionError: SRE module mismatch`, a foreign stdlib loaded against its own compiled extensions. Worth saying why I bothered — this repo keeps an api_truth ledger because a confident-but-wrong explanation left in the tree outlives whoever knew better, and a comment blaming the wrong cause would have sent the next person hunting a hang that isn't there.

Both tests came through unchanged.

PR #119, opened 2026-08-05
fix: alias background/async_job to prefer_handle for analyze_* actions

staahlarkitektur:
## Problem

`background=true`/`async_job=true` was silently a no-op on `analyze_clip`, `analyze_bin`, `analyze_file`, `analyze_project`, and `analyze_sequence`. `_run_maybe_background` — the only code that actually starts an off-thread `background_jobs` job — is wired to exactly three timeline actions (`export_timeline_checked`, `import_timeline_checked`, `create_subtitles_from_audio`) and never to these. The `analyze_*` actions collapse straight to `action="plan"` and, when not a dry run, run the full analysis synchronously inside the same coroutine the caller is awaiting — no `job_id` is ever returned, and the docstring never documented `background` as a param for them either.

The real async path for analysis work is `prefer_handle=true`, which reroutes to `start_batch_job` (its own durable, slice-runnable job machinery, polled via `batch_job_status`) — a completely different, already-correct code path.

In practice this meant a caller passing `background=true` on `analyze_clip` got no indication anything was wrong — the call just ran to completion (or didn't) with no job_id, indistinguishable from a hang.

## Fix

Rather than build a second job-tracking system, `background`/`async_job` are now aliases for `prefer_handle`: added `_media_analysis_wants_batch_job(p)`, a small pure helper next to `_media_analysis_bool`, and swapped it in at the one dispatch site that used to check `prefer_handle` alone. Same `dry_run` guard as before — a dry_run call still returns the synchronous plan regardless of which alias was used. Docstring updated to document all three params and the `{job_id, status}` return shape.

Extracted as a standalone function (not an inline condition) specifically so it's unit-testable with plain dicts, no Resolve mocking required.

## Testing

Added `MediaAnalysisWantsBatchJobTests` (6 tests) covering all three aliases, string-coerced truthy values, the no-opt-in default, and the `dry_run` interaction.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -8049,6 +8049,25 @@ def _media_analysis_bool(value: Any, default: bool = False) -> bool:
     return bool(value)
 
 
+def _media_analysis_wants_batch_job(p: Dict[str, Any]) -> bool:
+    """True when an analyze_* call should divert to the durable batch-job path.
+
+    `prefer_handle` is the documented opt-in; `background`/`async_job` are accepted
+    as aliases. Both previously looked like the async opt-in used by the timeline
+    actions (export/import/create_subtitles_from_audio) but were silently ignored
+    here, leaving callers with no job_id and no way to return early — see
+    davinci-resolve-mcp CONTRIBUTION_NOTES.md, Bug 2a. `dry_run` calls never divert,
+    same as `prefer_handle` alone did before.
+    """
+    if _media_analysis_bool(p.get("dry_run"), False):
+        return False
+    return (
+        _media_analysis_bool(p.get("prefer_handle"), False)
+        or _media_analysis_bool(p.get("background"), False)
+        or _media_analysis_bool(p.get("async_job"), False)
+    )
+
+
 def _media_analysis_target_dict(raw_target: Any, p: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
     p = p or {}
     if raw_target is None:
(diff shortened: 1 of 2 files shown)


samuelgursky (commented):
Reviewed against `main` @ v2.79.2. Merges cleanly; full offline suite passes (2442, exactly +6).

The bug you found is real: `background`/`async_job` are silently ignored on the `analyze_*` actions, and a caller gets no job_id and no indication anything was wrong. Worth fixing. But I don't think aliasing to `prefer_handle` is the right fix, because of what's on the other end of it.

### Blocking: the alias promises background execution and doesn't deliver it

Everywhere else in this server, `background=true` routes through `_run_maybe_background` → `background_jobs.start_job(...)`, which **actually starts the work off-thread**. It finishes on its own; you poll `resolve_control(action='job_status')` until it does.

`start_batch_job` does something different. It **creates a queued job and runs nothing** — the row goes in with status `"queued"` (`src/utils/media_analysis_jobs.py:405`), and nothing in `server.py` pumps it. The caller has to drive `run_batch_job_slice` themselves.

So after this change, an agent that learned the server-wide `background` contract and passes it to `analyze_clip` gets a job that never progresses, and polls forever. That's a new silent hang — and arguably worse than the bug being fixed, because the pre-fix behavior at least completed the analysis.

Two ways out, either fine by me:

1. **Don't alias.** Have `background`/`async_job` on `analyze_*` return an explicit error naming `prefer_handle` / `start_batch_job`. Turns a silent no-op into a loud one, which is the actual complaint.
2. **Alias, but say what you handed back.** Make the diverted response state that the job is queued and must be driven with `run_batch_job_slice`, so a caller can't mistake it for the `background_jobs` contract.

`prefer_handle` doesn't have this problem because its name doesn't promise anything about execution. `background` does.

### The new docstring is wrong about the return shape

The tool docstring now documents:

```
-> {clips, manifest} | {job_id, status} if handled
```

`create_batch_job` actually returns `{"success", "job", "plan"}` (`src/utils/media_analysis_jobs.py:444-456`). `job_id` is nested inside `job`, and there's no top-level `status`. `_media_analysis_wants_batch_job`'s own docstring repeats the claim.

This is the same failure mode as the bug you're fixing — a caller trusting the documented shape and finding nothing there — so it should get the same care.

### Wrong premise in the PR body

> `_run_maybe_background` — the only code that actually starts an off-thread `background_jobs` job — is wired to exactly three timeline actions

It has 12 call sites in `src/server.py`: 5917, 6265, 7612, 16940, 17240, 17243, 17632, 17635, 21202, 21668, 21672, 21685. Doesn't change the fix, but it's part of the reasoning for it, and the "only three actions" framing is what makes aliasing look safe.

### The fix is incomplete under one preference

The divert is evaluated *after* `p["dry_run"]` is normalized from the `dry_run_first_default` preference (`src/server.py:18804`). A user who turns that on gets `background=true` silently ignored again — same symptom, same silence. It ships default-off so this isn't a regression, but the hole you're closing still has this corner in it. An explicit note in the response ("async opt-in ignored: dry_run is on by preference") would finish the job.

### What's good

Extracting `_media_analysis_wants_batch_job` as a pure function instead of leaving it an inline condition was the right call, and the six tests are the payoff — real coverage with no Resolve mocking. Please keep that shape in whatever the fix turns into.

One note: `_media_analysis_bool(p.get("dry_run"))` vs the old raw `not p.get("dry_run")` is a no-op at this call site, since `dry_run` is already normalized to a bool upstream at 18804. Harmless, just not doing anything.

### Housekeeping

`MediaAnalysisWantsBatchJobTests`'s docstring cites `davinci-resolve-mcp CONTRIBUTION_NOTES.md, Bug 2a`. That file isn't in this repo (same in #118 and #120). And all three of your PRs append test classes at the identical spot in `tests/test_media_analysis.py`, so they conflict with each other there — trivial to resolve, but worth knowing before you rebase.

samuelgursky:
Merged, and shipped in [v2.80.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.80.0) — with the fix taken further than the PR, so I want to be clear about what changed and why.

The bug you found is real and was worth finding: `background`/`async_job` were accepted on the `analyze_*` actions and silently ignored, so the call ran the whole analysis inline and returned no `job_id`. From the caller's side that is indistinguishable from a hang, which is the worst way for a parameter to fail.

I didn't take the aliasing, because of what sits on the other end of `prefer_handle`. `start_batch_job` only inserts a row at status `"queued"` — nothing in `server.py` advances it, and the caller is expected to drive `run_batch_job_slice`. Everywhere else in this server, `background=true` goes through `_run_maybe_background`, which starts the work and lets it finish on its own. So aliasing would have handed an agent that learned the second contract a job that never moved: a quieter failure than the one you were fixing, and arguably worse, since the pre-fix behaviour at least completed the analysis.

So the two opt-ins stay distinct and the missing half got built:

- `prefer_handle=true` → creates the job, hands it back **queued**. Your contract, unchanged.
- `background=true` / `async_job=true` → creates the job **and drives it to completion off-thread**, via a new `start_batch_job_runner`. Poll `batch_job_status` until terminal.

One design note in case it's useful: the runner deliberately does *not* go through `background_jobs.start_job`, even though that's the obvious reuse. That wrapper holds `resolve_busy.long_resolve_op`, which serializes callers against Resolve's single-threaded scripting bridge — and analysis touches that bridge nowhere. `media_analysis` and `media_analysis_jobs` drive ffmpeg, whisper and vision over file paths only. Holding the gate for an hour of transcription would have locked the editor out of Resolve for the duration and bought nothing. A process-wide slice lock bounds the real cost instead, so several queued analyses interleave a clip at a time rather than starting N ffmpeg passes at once.

Two smaller corrections from the review, both fixed:

- The docstring advertised `{job_id, status}`; the real envelope is `{success, job, plan}` with the id nested at `job.job_id`. Now documented as it is, plus `running` and a `note` naming the next call so the queued and running routes can't be mistaken for each other.
- The divert sat *after* `dry_run` was resolved from the `dry_run_first_default` preference, so a user with that preference on still got `background=true` swallowed in silence. An explicit `dry_run` still wins; an inherited one no longer does.

Your `_media_analysis_wants_batch_job` became `_media_analysis_async_mode` and returns a mode rather than a bool, so its tests were rewritten — but extracting it as a pure function instead of leaving it an inline condition was the right call and I kept that shape. Testable with plain dicts and no Resolve mocking is exactly what that dispatch site needed.

Thanks for all three of these. Windows is where this server gets the least direct testing from me, and specific, measured reports from that platform are worth a lot.

PR #120, opened 2026-08-05
fix: kill the whole process tree on _run_command timeout, not just the direct child

staahlarkitektur:
## Problem

`subprocess.run(timeout=...)`'s default `TimeoutExpired` handling calls `proc.kill()` on the direct child only. On Windows, a bare-name PATH lookup (e.g. `ffmpeg`) can resolve to a package-manager shim (Chocolatey's ShimGen shims, npm shims, etc.) rather than the real binary — the shim spawns the actual work as its own child, i.e. a grandchild of this process. Killing the shim's PID leaves that grandchild running untouched, and because it inherited the shim's stdout/stderr pipe handles, the fallback `communicate()` (which `subprocess.run` calls with no timeout after a kill) blocks until the grandchild exits on its own — a normally-completing pass just returns late with a misleadingly-labeled "timed out" result; a genuinely-stalled one hangs forever.

Confirmed empirically on a Chocolatey-managed machine: `ffmpeg` on PATH resolved to a 392KB ShimGen shim, not the real ~101MB binary. A 5s timeout against an ~82s real `ffmpeg` pass had zero effect — the call returned after the full 82s with a "timed out after 5s" message attached to what was actually the complete, correct output.

## Fix

`_run_command` now spawns via `Popen` (new process group on Windows via `CREATE_NEW_PROCESS_GROUP`, new session on POSIX via `start_new_session`) and uses `communicate(timeout=...)` directly, so a timeout can be handled with a real tree-kill instead of trusting a single-PID `kill()`. The new `_kill_process_tree(pid)` helper uses `taskkill /F /T /PID` on Windows and `os.killpg(pid, SIGKILL)` on POSIX — the same platform-branch idiom already used elsewhere in this codebase for process termination (`src/utils/app_control.py`).

This fixes every `_run_command` caller at once (the whisper CLI call and all `ffmpeg` passes in `_readthrough_analysis`) without any call-site changes; the function's signature and `Tuple[int, str, str]` return contract are unchanged. No new dependency — `taskkill`/`killpg` are both stdlib-reachable.

## Testing

Re-ran the original repro against this fix: the same `timeout=5` call that previously took 82.2s (kill had no effect) now returns in ~5s with the process tree actually gone.

Added `RunCommandTimeoutKillsProcessTreeTests` (proves the kill actually cuts the wait short using a `sys.executable` child with no shim involved, and that ordinary non-timeout completion is unaffected) and `KillProcessTreeWindowsTests`/`KillProcessTreePosixTests` (platform-branch dispatch, mocked, whichever applies to the CI runner's OS).

### DIFF
--- a/src/utils/media_analysis.py
+++ b/src/utils/media_analysis.py
@@ -17,6 +17,7 @@ import os
 import platform as _platform
 import re
 import shutil
+import signal
 import sqlite3
 import subprocess
 import sys
@@ -2376,23 +2377,57 @@ def build_plan(
     }
 
 
-def _run_command(args: List[str], timeout: int = COMMAND_TIMEOUT_SECONDS) -> Tuple[int, str, str]:
-    try:
-        proc = subprocess.run(
-            args,
-            capture_output=True,
-            timeout=timeout,
-            check=False,
+def _kill_process_tree(pid: int) -> None:
+    """Best-effort: terminate pid and its descendants, not just the direct child.
+
+    Popen.kill() only reaches the immediate child. On Windows a PATH lookup can
+    resolve to a shim (Chocolatey, npm, etc.) that spawns the real work as a
+    grandchild, which survives a plain kill() untouched — confirmed for `ffmpeg`
+    on a Chocolatey-managed machine (davinci-resolve-mcp CONTRIBUTION_NOTES.md,
+    Bug 2b: a 5s timeout had zero effect on an ~82s real ffmpeg pass, because
+    killing the shim's PID left the real ffmpeg.exe grandchild running). Killing
+    the whole tree instead of one PID fixes this regardless of what kind of
+    wrapper sits on PATH.
+    """
+    if os.name == "nt":
+        subprocess.run(
+            ["taskkill", "/F", "/T", "/PID", str(pid)],
+            capture_output=True, check=False,
         )
-    except subprocess.TimeoutExpired as exc:
-        stdout = exc.stdout.decode("utf-8", errors="replace") if exc.stdout else ""
-        stderr_tail = exc.stderr.decode("utf-8", errors="replace") if exc.stderr else ""
-        return 124, stdout, f"Command timed out after {timeout}s. {stderr_tail}".strip()
+        return
+    try:
+        os.killpg(pid, signal.SIGKILL)
+    except ProcessLookupError:
+        pass
+
+
+def _run_command(args: List[str], timeout: int = COMMAND_TIMEOUT_SECONDS) -> Tuple[int, str, str]:
+    popen_kwargs: Dict[str, Any] = {}
+    if os.name == "nt":
+        popen_kwargs["creationflags"] = subprocess.CREATE_NEW_PROCESS_GROUP
+    else:
+        popen_kwargs["start_new_session"] = True
+    try:
+        proc = subprocess.Popen(
+            args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **popen_kwargs
+        )
     except OSError as exc:
         return 127, "", str(exc)
-    stdout = proc.stdout.decode("utf-8", errors="replace") if proc.stdout else ""
-    stderr = proc.stderr.decode("utf-8", errors="replace") if proc.stderr else ""
-    return proc.returncode, stdout, stderr
+    try:
+        stdout, stderr = proc.communicate(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        # subprocess.run's default TimeoutExpired handling only kills the direct
+        # child, which is insufficient when that child is itself a wrapper — see
+        # _kill_process_tree. The follow-up communicate() (no timeout) drains
+        # whatever's left now that the tree is actually gone.
+        _kill_process_tree(proc.pid)
+        stdout, stderr = proc.communicate()
+        stdout_s = stdout.decode("utf-8", errors="replace") if stdout else ""
+        stderr_s = stderr.decode("utf-8", errors="replace") if stderr else ""
+        return 124, stdout_s, f"Command timed out after {timeout}s. {stderr_s}".strip()
+    stdout_s = stdout.decode("utf-8", errors="replace") if stdout else ""
+    stderr_s = stderr.decode("utf-8", errors="replace") if stderr else ""
+    return proc.returncode, stdout_s, stderr_s
 
 
 def _write_json(path: str, payload: Dict[str, Any]) -> None:
--- a/tests/test_media_analysis.py
+++ b/tests/test_media_analysis.py
@@ -2,9 +2,12 @@ import asyncio
 import json
 import os
 import shutil
+import signal
 import sqlite3
 import subprocess
+import sys
 import tempfile
+import time
 import unittest
 import unittest.mock
 from typing import Any, Dict, Optional
@@ -58,6 +61,8 @@ from src.utils.media_analysis import (
     execute_plan,
     execute_plan_async,
     executing_clips,
+    _kill_process_tree,
+    _run_command,
     plan_requires_capabilities,
     load_report,
     mark_registry_stale_for_clip,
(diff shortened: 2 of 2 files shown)


samuelgursky (commented):
Reviewed against `main` @ v2.79.2. Merges cleanly; full offline suite passes (2439 + 1 platform skip, exactly +4).

This is the strongest of the three. The diagnosis is genuinely good — a shim on PATH means the kill hits the wrapper while the real work continues holding the inherited pipe handles, and the no-timeout `communicate()` that `subprocess.run` does after a kill then blocks on them. That's a hang with a completely misleading label on it ("timed out after 5s" attached to 82 seconds of correct output), and it would have been very hard to find from the outside. The mechanism you reached for is right, and it matches the platform-branch idiom already in `src/utils/app_control.py`.

Two things to fix before this lands, both in the recovery path.

### 1. `_kill_process_tree` can raise

It catches only `ProcessLookupError`. But `os.killpg` raises `PermissionError` on EPERM, and on Windows `subprocess.run(["taskkill", ...])` raises `FileNotFoundError` if taskkill isn't on PATH. Either escapes `_run_command` and breaks its `Tuple[int, str, str]` contract — at the exact moment the caller is already handling a failure. A best-effort helper shouldn't be able to throw; catch `OSError`.

### 2. The post-kill `communicate()` is unbounded

```python
_kill_process_tree(proc.pid)
stdout, stderr = proc.communicate()   # no timeout
```

The tree kill is best-effort by construction. If it partially fails — a descendant that detached itself, taskkill denied — that call blocks forever on pipes the survivor still holds. That's the same hang, moved into the recovery path. It needs a bound, and an explicit give-up when the bound is hit.

### Worth an explicit decision: POSIX orphans

`start_new_session=True` detaches children from the server's process group, so ffmpeg/ffprobe no longer receive SIGINT when the server is Ctrl-C'd — they keep running as orphans. Pre-fix they died with the terminal. This applies to every `_run_command` caller, including `silence_ripple.py` and `deep_vision.py`. It's the price of `killpg` and probably worth paying, but it should be a decision rather than a side effect.

### Minor

- `subprocess.run` kills the child on *any* exception on the way out; `Popen` doesn't. Under this server's threaded dispatch, a cancellation or KeyboardInterrupt mid-`communicate` now leaves an orphaned tree. A `except BaseException: _kill_process_tree(...); raise` covers it.
- `CREATE_NEW_PROCESS_GROUP` isn't what makes the Windows tree kill work — `taskkill /T` walks parent-child links, not process groups. It's harmless and reasonable to keep for signal isolation, but the code reads as though it's load-bearing. `start_new_session` *is* load-bearing on POSIX, where `killpg` needs the child to lead its own group.

### Integrating with #118

#118 adds an `env=` parameter to `_run_command`; this PR's rewrite drops it. If #118 lands first and this one's conflict is resolved by taking its side wholesale, the whisper call raises `TypeError` at runtime. #118's tests catch that, so long as both test sets survive the tests-file conflict — all three of your PRs append at the identical spot.

I merged the two locally (this one first, then #118) with the fixes above and ran the full suite: 2443 + 1 skip, green. The combined version:

```python
def _kill_process_tree(pid: int) -> None:
    """Best-effort: terminate pid and its descendants, not just the direct child.

    Popen.kill() reaches only the immediate child. On Windows a bare-name PATH
    lookup can resolve to a wrapper — a Chocolatey/npm shim, a pip console
    script — that runs the real work as a grandchild, which a single-PID kill
    leaves untouched. Measured: `ffmpeg` on PATH was a 392KB shim, and a 5s
    timeout against an ~82s real ffmpeg pass had no effect at all, because the
    surviving grandchild still held the stdout/stderr handles it had inherited
    and the follow-up read blocked until it finished on its own.

    Failure here is never fatal. The caller is already on its error path and
    owes its own caller a (code, stdout, stderr) tuple, so this must not raise:
    `taskkill` can be absent from PATH and `killpg` can return EPERM, which is
    why the whole branch catches OSError rather than only ProcessLookupError.
    """
    try:
        if os.name == "nt":
            subprocess.run(
                ["taskkill", "/F", "/T", "/PID", str(pid)],
                capture_output=True,
                check=False,
            )
        else:
            os.killpg(pid, signal.SIGKILL)
    except OSError:
        pass


# How long to wait for the pipes to drain after a tree kill. The kill is
# best-effort, so this read has to be bounded: anything that escaped it still
# holds the inherited pipe handles, and an unbounded read there would hang for
# exactly the reason the kill exists.
_POST_KILL_DRAIN_SECONDS = 5


def _run_command(
    args: List[str],
    timeout: int = COMMAND_TIMEOUT_SECONDS,
    env: Optional[Dict[str, str]] = None,
) -> Tuple[int, str, str]:
    """Run args to completion and return (returncode, stdout, stderr).

    Spawned via Popen rather than subprocess.run so a timeout can kill the whole
    process tree instead of one PID — see _kill_process_tree.

    `env=None` inherits this process's environment, matching what subprocess.run
    did. Pass an explicit mapping for a child that must not inherit it: on
    Windows this server sets PYTHONHOME so the fusionscript bridge can find
    Resolve's Python, and a child that is itself a *different* Python (the
    whisper CLI) dies loading a foreign stdlib against its own C extensions.

    Returns 124 on timeout, 127 when the binary cannot be spawned.
    """
    popen_kwargs: Dict[str, Any] = {}
    if os.name == "nt":
        # Isolates the child from console signals sent to the server. Note this
        # is not what makes the tree kill work — taskkill /T walks parent-child
        # links, not process groups. start_new_session is load-bearing on POSIX,
        # where killpg needs the child to lead a group of its own.
        popen_kwargs["creationflags"] = subprocess.CREATE_NEW_PROCESS_GROUP
    else:
        popen_kwargs["start_new_session"] = True
    try:
        proc = subprocess.Popen(
            args,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            env=env,
            **popen_kwargs,
        )
    except OSError as exc:
        return 127, "", str(exc)
    try:
        stdout, stderr = proc.communicate(timeout=timeout)
    except subprocess.TimeoutExpired:
        _kill_process_tree(proc.pid)
        abandoned = False
        try:
            stdout, stderr = proc.communicate(timeout=_POST_KILL_DRAIN_SECONDS)
        except subprocess.TimeoutExpired:
            # A descendant outlived the tree kill and still holds the pipes.
            # Give up the output rather than block — a stalled caller is a
            # worse outcome than a timeout report with no stderr tail.
            stdout, stderr = b"", b""
            abandoned = True
        stdout_s = stdout.decode("utf-8", errors="replace") if stdout else ""
        stderr_s = stderr.decode("utf-8", errors="replace") if stderr else ""
        detail = " Output abandoned: a descendant survived the kill." if abandoned else ""
        return 124, stdout_s, f"Command timed out after {timeout}s.{detail} {stderr_s}".strip()
    except BaseException:
        # subprocess.run kills the child on any exception on the way out;
        # Popen does not. Under the server's threaded dispatch a cancellation
        # or KeyboardInterrupt here would otherwise leave an orphaned tree.
        _kill_process_tree(proc.pid)
        raise
    stdout_s = stdout.decode("utf-8", errors="replace") if stdout else ""
    stderr_s = stderr.decode("utf-8", errors="replace") if stderr else ""
    return proc.returncode, stdout_s, stderr_s
```

Plus two tests covering the fixes — that the platform call's `OSError` is swallowed, and that `_run_command` still returns promptly when the kill doesn't take (patching `_kill_process_tree` to a no-op and shrinking the drain, so it proves the bound rather than the kill). Happy to push these onto your branch or open them as a follow-up, whichever you prefer.

### Housekeeping

`_kill_process_tree`'s docstring and the test docstrings cite `davinci-resolve-mcp CONTRIBUTION_NOTES.md, Bug 2b`. That file isn't in this repo (same in #118 and #119). The measurement is the valuable part — 392KB shim, 5s timeout, 82s pass — so state it inline, as the version above does.

Your tests are well-judged, particularly using `sys.executable` to isolate the kill mechanism from the shim scenario, and the deliberately loose timing bound.

samuelgursky:
Merged, and shipped in [v2.80.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.80.0). Thank you for this one in particular — the shim/grandchild finding is genuinely good work.

It's the kind of bug that's close to invisible from where I sit: the symptom is a correct result arriving late with a wrong label on it, on a platform I develop for blind. "392KB shim, not the real ~101MB binary" is exactly the measurement that makes a diagnosis checkable instead of plausible, and going back to re-run the original repro against the fix (82.2s → ~5s) is what took it from a theory to a finding. That's how I'd want every bug report to arrive.

Two follow-up commits on top of the merge, both in the recovery path rather than the mechanism:

- `_kill_process_tree` could raise. `os.killpg` returns EPERM as `PermissionError`, and `taskkill` raises `FileNotFoundError` when it isn't on PATH — either escaped `_run_command` and broke its `(code, stdout, stderr)` contract at the exact moment the caller was already handling a failure. Now catches `OSError`.
- The `communicate()` after the kill had no timeout. Since the kill is best-effort by construction, a descendant that escaped it still holds the pipes and that read blocks forever — the same hang, moved one level in. Now bounded, and it says so in the result when it gives the output up rather than pretending it drained cleanly.
- Smaller: `subprocess.run` kills the child on any exception on the way out and `Popen` doesn't, so a cancellation mid-`communicate` under threaded dispatch would have orphaned the tree. `BaseException` now kills before re-raising.

I also moved your measurement inline. `CONTRIBUTION_NOTES.md` isn't in this repo — I think it was local to your working copy — so the references would have been dead ends for the next reader. The numbers were the valuable part and they're now in the docstring itself.

One thing that came out of the merge and wasn't in anyone's description: `start_new_session=True` detaches children from the server's process group, so `ffmpeg`/`ffprobe` no longer die when the server is Ctrl-C'd — they orphan. That's the price of `killpg`, it applies to `silence_ripple` and `deep_vision` too, and I judged it worth paying. Noting it here so it's on the record rather than discovered later.

Your tests came through as-is. Using `sys.executable` to isolate the kill mechanism from the shim scenario was the right instinct, and so was the deliberately loose timing bound.

PR #121, opened 2026-08-06
fix(claude-code): the .claude/skills/ skills never loaded — plus the guards they assume

Grimthereapper:
## The bug

None of the ten domain skills in `.claude/skills/` were loading.

Claude Code discovers skills at `.claude/skills/<name>/SKILL.md`. These were
loose `.md` files at the top level of that directory — a layout the loader does
not scan. The result was silent: no warning, no error, just ten skills that no
agent could ever see, while the generated domain-routing table in `AGENTS.md`
and the index in `docs/README.md` both advertised them as available.

Verified by observation, not inference: with the files loose, a Claude Code
session in this repo lists no `resolve-*` skills. After the move, all ten appear.

## What changed

**The fix.** Each skill moves into a directory named for its frontmatter `name`
(`color-grade.md` → `resolve-color/SKILL.md`, and so on). Git records all ten as
100% renames, so history follows. Path references in `docs/README.md` and seven
kernel docs are updated, and `docs/README.md` now states the directory
requirement so this cannot regress quietly.

**The guards.** With the skills reachable, this adds the enforcement they assume
— two `PreToolUse` hook scripts:

- `frame_verification_guard.py` denies grade-applying actions on
  `timeline_item_color` (`safe_set_cdl`, `safe_apply_drx`, `propose_grade`, …)
  until the session has actually pulled and looked at a Resolve-rendered frame.
  `safe_copy_grade` and `bulk_match_to_hero` additionally always ask, since those
  apply a whole-grade artifact across clips and can overwrite hand-work with no
  recovery path. `dry_run` passes through untouched.
- `source_media_guard.py` denies shell commands that write, move, or delete
  source media outside a scratch root. Reads pass — `ffprobe`, and `ffmpeg`
  writing into scratch.

`AGENTS.md` has required frame references and source-media safety all along, but
both were prose. A prose rule decays as a session gets long; a hook does not.

**The hooks are opt-in.** The scripts ship, but nothing wires them up — an
earlier revision of this branch enabled them from a committed
`.claude/settings.json`, which would have turned them on for every contributor
who cloned the repo, including people with no Resolve running who never touch a
grade. Enabling hooks in someone else's session is not the repository's call.
`docs/README.md` now carries the exact block to paste into a personal,
gitignored `.claude/settings.local.json`, plus the note that hooks are read at
session start and need a restart.

**Two review subagents** in `.claude/agents/`, which run in their own context so
frame images stay out of the main session:

- `cut-reviewer` screens an assembled timeline from its frames — pacing, shot
  order, continuity, coverage gaps. It is told explicitly that a metadata summary
  is not a review, because assembling through an API succeeds loudly and fails
  quietly.
- `grade-match-verifier` measures shot match numerically from rendered frames,
  and must report the pixel count behind every masked measurement — a near-empty
  skin mask returns a delta near zero and reads as a perfect match.

**Two skills** outside the domain routing: `house-style` (Claude-only)
accumulates editorial corrections so the same note is not given twice, and
`/resolve-session` reports connection, edition, project, and timeline state
before editing starts.

## For the reviewer

- **The rename is the load-bearing part.** The hooks, agents, and new skills are
  additive and can be dropped independently if they are not wanted; the directory
  move is the actual bug fix.
- **Nothing is enabled on your behalf.** No committed settings file, no hooks
  active by default. A fresh clone behaves exactly as it does today unless
  someone opts in.
- **Hook behavior is tested.** Both scripts were exercised against 13 cases
  covering deny, ask, and silent-allow paths — including `ffprobe` reads,
  `ffmpeg` into scratch, chained `ffprobe && rm`, redirection onto a media file,
  and ordinary repo commands like `git status` and the test runner. Allow cases
  exit 0 silently and stay out of the way.
- **Action names were read from `src/server.py`**, not guessed —
  `_COLOR_GRADE_KERNEL_ACTIONS`.
- `tests/test_agent_rules_drift.py` passes, and
  `scripts/agent-rules/generate.mjs --check` reports in sync. No skill names
  changed, so the generated `AGENTS.md` block needs no regeneration.
- **`house-style` ships as a scaffold, deliberately.** Its sections are empty and
  it carries the capture protocol instead. It also states that personal colour
  and look preferences belong in a user-level skill, not in a file that travels
  with the repository.
- Nothing under `src/` is touched.


samuelgursky:
Reviewed this by checking it out and exercising the hooks rather than reading them. The core finding holds up: in a Claude Code session in this repo, none of the ten `resolve-*` skills appear in the available-skills list. They have never loaded. The rename is a real fix for a real, silent bug, and it's the part of this PR that matters.

I verified the supporting claims too — all five `GRADE_APPLY_ACTIONS` and all five `EVIDENCE_ACTIONS` do exist in `_COLOR_GRADE_KERNEL_ACTIONS` (`src/server.py:22236`), the `{action, params}` shape matches the real tool signature, `generate.mjs --check` reports in sync, and `test_agent_rules_drift.py` passes on a merged tree. Making the hooks opt-in in 297ea58 was the right call; agreed that enabling hooks in someone else's session isn't the repo's decision.

I've pushed two commits to this branch.

**0801b6c — `is_scratch` matched substrings.** This one is a genuine escape, not a nit. The function lowercased the path and looked for `/renders/`, `/exports/`, `/tmp/` and friends anywhere inside it, with no normalization:

```
rm /Volumes/CARD/exports/MASTER_A001.mov   -> allowed
rm /Users/x/Client/renders/hero_v3.mov     -> allowed
rm /Volumes/CARD/renders/../A001.mov       -> allowed, and `..` walks back out
```

A delivery drive with an `exports` or `renders` folder sat outside the guard entirely — which is the exact case the guard exists for.

The fix normalizes first (`expanduser`/`abspath`/`normpath`) so traversal can't outrun a marker, then matches whole path components instead of substrings. It also splits the markers by what they license: a real scratch root (`/tmp`, `/private/tmp`, `/var/folders`, `$RESOLVE_MCP_SCRATCH`, a `claude-<uid>` session dir, the analysis project root) exempts anything, while `proxies`/`renders`/`exports` exempt only a **write** — a render landing where it belongs — and no longer exempt `rm`, `mv`, `truncate`, `dd`, `shred`, or `unlink`. A camera card with an `exports` folder is still a camera card.

26 cases: the three escapes now deny, writes into `renders/proxies/exports` still pass, and every allow/deny from your original matrix is unchanged (`ffprobe`, `ffmpeg` into scratch, chained `ffprobe && rm`, redirection onto media, glob `rm`, quoted paths, `git status`, pytest).

I also added a short "what this does not catch" note to the module docstring. Testing turned up three classes that pass and always will, because the hook reads argv lexically: a directory delete with no media extension (`rm -rf /Volumes/CARD/DCIM` — arguably the worst command in this family), indirect deletes (`find … -delete`, `… | xargs rm`), and scripts that write media themselves (`python transcode.py --out …`). Not worth chasing; worth stating, so nobody reads the hook as a boundary. It's a tripwire for the common direct mistake.

**0eb6637 — `scripts/agent-rules/README.md` still documented `.claude/skills/*.md`** in three places, including the "New domain" checklist. That's the layout that caused the bug, sitting in the file someone reads when adding a skill. `docs/README.md` covers the requirement now; this is the other half.

Two things I did not change, flagging instead:

- `house-style` is committed and instructs the model to append editorial corrections to it. This repo keeps content- and client-specific notes out of the tree, and the file's own closing line concedes it "should be gitignored rather than committed" if entries get person-specific. I'd rather gitignore it now than after the first client note lands in git history.
- The `color_group` half of the frame guard's matcher is a no-op — the hook only inspects `timeline_item_color` actions. Harmless, but it implies coverage that isn't there.

Also note `grade_version_restore` writes grade state and isn't gated. I think that's right, since restore is recovery, but it's worth being a decision rather than an omission.

Nice work finding this. The skills being invisible while `AGENTS.md` advertised them is exactly the kind of failure that hides indefinitely.

PR #122, opened 2026-08-06
docs(readme): add Simplified Chinese translation (README.zh-CN.md)

chenyuxiaojin:
## What

A full Simplified Chinese translation of the README, current as of v2.80.1, plus a one-line language switcher at the top of the English README.

## Scope, deliberately narrow

- **README only.** Nothing under `docs/` is translated — in particular the generated references (`docs/reference/api-limitations.md` etc.) stay single-source with their drift guards, and every link in the translation keeps pointing at the English docs.
- The translation opens with a note that the English README is canonical and names the version it corresponds to, so a lagging translation can never masquerade as current.

## Maintenance

I'll keep it in sync with README changes — the closing section of the translation says so publicly, so readers know who to ping. If it ever goes stale enough to mislead, feel free to delete it; a missing translation is better than a wrong one.

## Translation notes

- Post-production terms use the mainland industry vocabulary (conform → 套底, take → 条次, turnover → 交接, dead-space → 死气口); product and tool names (bin, Fusion, Fairlight, DRX, ColorSlice…) stay in English.
- "What This Does Not Do" was translated with particular care to preserve its voice — it is the most load-bearing section of the README for a new user deciding what to trust.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # DaVinci Resolve MCP Server
 
+English | [简体中文](README.zh-CN.md)
+
 [![Version](https://img.shields.io/badge/version-2.80.1-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
 [![npm](https://img.shields.io/npm/v/davinci-resolve-mcp.svg?label=npm&color=CB3837)](https://www.npmjs.com/package/davinci-resolve-mcp)
 [![API Coverage](https://img.shields.io/badge/API%20Coverage-100%25-brightgreen.svg)](docs/reference/api-coverage.md)
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -1,0 +1,258 @@
+# DaVinci Resolve MCP 服务器
+
+[English](README.md) | 简体中文
+
+[![Version](https://img.shields.io/badge/version-2.80.1-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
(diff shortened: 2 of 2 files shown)


samuelgursky:
Merged as f3e2917, on `main` now. Landed by cherry-pick because a branch ruleset blocks the merge button; the commit keeps your authorship, so GitHub shows this as closed rather than merged.

Two things changed on the way in.

**Brought current to v2.83.0 rather than merged as-is.** You wrote against v2.80.1 and the English README had moved twice since. It turned out to be only two hunks — the version badge, and the Linux bridge sentence from #129 — so I carried both across instead of merging a file whose own header pointed three releases back. The Linux paragraph is my translation, not yours; if the phrasing is off (particularly 直接对系统 Python 列出脚本), correct it.

**Your maintenance offer is now a checklist item instead of a promise.** `docs/process/release-process.md` lists `README.zh-CN.md` among the surfaces every release must update, and names deleting the file and the language switcher as an acceptable outcome if it can't be kept current.

That framing came from your own PR description — "a missing translation is better than a wrong one" — and the reason it needed to be written down is the thing you designed in deliberately: the version line. A translation with no version marker just goes quietly old. Yours makes a *specific claim* about which release it matches, so going stale turns it into a false statement rather than a vague one. That's better, because it's checkable — but nothing in CI can check it, since a lagging translation is still valid Markdown. Hence a human step on the checklist.

On scope: keeping it to the README and leaving `docs/` untranslated was the right call, and the drift guards on the generated references are exactly why.

It ships in the next release. v2.83.0 is tagged but GitHub Actions is mid-outage, so npm is a version behind until that clears.

Issue #125, opened 2026-08-06
Whisper swallows immediate retakes — "pause + second take" gets stuffed into one word's duration, and every transcript-reading feature is blind to it


chenyuxiaojin:
## The blind spot (measured 2026-07-31, mlx_whisper large-v3-turbo, Chinese speech, macOS)

When a speaker re-reads a sentence immediately (the normal way people self-correct while recording), whisper — full-pass and isolated-window transcription alike — outputs the text **once**, aligns it to the **first** take, and absorbs "pause + entire second take" into the duration of a single word. Measured: one single-character word spanning **6.06s**, another spanning **4.4s**, on material where real speech runs 4-7 chars/s.

Consequences, all downstream of one lie:

- Any transcript-reading layer (retake detection, filler-word analysis, `rank_takes`-style fluency scoring, word-level cuts) **never sees these retakes** — the transcript says the sentence was spoken once, cleanly.
- Segment-level SRT and word timestamps inflate around the swallow point, so cuts placed on word boundaries near a retake land inside speech.

## Energy detection cannot catch it either

The swallowed region is not silence: breathing and keyboard noise in the dead span peaked at **-12.1 dB**, *louder* than adjacent real speech at -16.7 dB. `silencedetect` at any threshold fails — our silence-based pass recalled 3 of 17 ground-truth instances.

## Ground truth

On one real 25-minute talking-head recording, after a silence-based tighten pass, the editor's manual cleanup added 22 cuts; **17 of them (~29s) were this class**, concentrated at the head of the "final take" segment of a retake pair — re-read once, pause, and the pause plus the false start got absorbed into the keeper's first word.

## A probe that works (offered in case it's useful)

Recursive re-transcription of stretched words: candidates = segments with implausible speech rate (≥2.5s duration and <4.5 chars/s) → extract the audio window, re-transcribe in isolation → look for large inter-word gaps and **stretched words** (single word ≥1.2s) → re-transcribe the sub-window *excluding the first take's span* — the second take's start then reveals itself, because whisper can no longer align it onto the first. Against 8 ground-truth cases: 4 exact hits (≤0.22s error), 3 correctly flagged for manual review, 1 miss (when alignment smoothing left no stretched word — a known residual).

## Suggestions (any subset)

- A documented limitation in the media-analysis guide: word timestamps are not trustworthy around immediate retakes; the transcript reports a retaken sentence as spoken once.
[...]

PR #126, opened 2026-08-06
feat(skills): add resolve-tighten-recording — long single-take dead-air pass

chenyuxiaojin:
## What

A second editing skill alongside `resolve-rough-cut`: the **subtractive** workflow. Where rough-cut selects shots from a folder of many clips, this one removes dead air from **one long single-take recording** (talking-head, screencast, tutorial) via `plan_tighten` → review → `execute_tighten` — a tool pair that has engine support but no skill documenting how to drive it safely.

## Evidence basis

Every claim was re-measured on **Resolve Studio 21.0.1.11 / MCP v2.80.1 (2026-08-06)** against a real 28.5-minute 30fps recording, or carries its own earlier measurement date inline. The frame-exact checks:

- plan `keep_ranges` are **source** frames with an exclusive `end_frame` (matches the existing `api_truth` AppendToTimeline entry), variant record positions are the cumulative sum — verified against `clip_where` readback per range.
- `structural_diff.added` `in_frame`/`out_frame` are **record** coordinates of the variant, not source frames — the skill's central trap table exists because feeding one where the other is expected assembles the wrong content with correct cut lengths and no error.
- Audio mirroring verified by item count (113 video + 113 audio = 226), not by the success envelope.
- `confirm_token` two-step verified within one MCP session; cross-restart deliberately marked unverified.
- `clip_where` reads the current timeline only (no `timeline_name` filter) — noted, since the skill leans on it for verification.

## Scope choices

- The "what tightening cannot hear" section routes to `plan_transcript_tighten` for fillers/tight restarts, marks its filler set as English-only, and keeps real retakes as an explicit human pass — matching the heuristic's own in-code scoping. It cross-references #125 (whisper swallows immediate retakes) as the class neither transcript nor silence tools can see.
- The multi-layer rebuild recipe (stacked camera + screen tracks) uses only public tool calls plus ffmpeg cross-correlation for alignment verification; measured residual on real dual-layer material was 34ms against a 50ms pass bar (2026-07-03).
- Nothing here executes anything: the skill teaches plan → review → confirm and the verification order, in line with the deliverable-contract style of `rough-cut.md`.

### DIFF
--- a/.claude/skills/tighten-recording.md
+++ b/.claude/skills/tighten-recording.md
@@ -1,0 +1,157 @@
+---
+name: resolve-tighten-recording
+description: Tightening one long single-take recording — talking-head, screencast, tutorial, podcast video — by removing dead air in the DaVinci Resolve MCP. Apply when asked to tighten a recording, remove silences or dead air, cut the pauses out of a long take, or turn a raw one-take recording into a first cut. The subtractive counterpart to resolve-rough-cut, which selects shots from many clips; this skill removes time from one clip.
+---
+
+# Resolve Tighten Recording — Claude Code Skill
+
+Turns one long raw take into a tightened variant timeline. **The variant is the
+deliverable; the original timeline is never touched.**
+
+Everything below was measured live against Resolve Studio 21.0.1.11 / MCP
+v2.80.1 on 2026-08-06, on a real 28.5-minute 30fps screen recording, unless a
+different date is given.
+
+## The deliverable contract
+
+Deliver a tightened **variant** timeline, plan → review → confirm. Do not
+grade, caption, or add anything; a tighten is pure subtraction. The variant
+runs generous on purpose — recovering over-cut material is slow and invisible,
+trimming further is fast and visible.
+
+Two things always go back to the editor with the variant:
+
+- **The largest lifts.** A lift is only *silence to the microphone*. Measured
+  on real material (2026-07-31): a 29s lift was 70% silent on-screen
+  demonstration — real content, cut because nobody spoke over it. Anything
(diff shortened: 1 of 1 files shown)


samuelgursky:
Merged as 7eb6ad0, shipped in [v2.81.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.81.0). Landed by cherry-pick rather than the merge button, so GitHub will show this as closed — the commit keeps your authorship.

One correction went in with it, and it's the bug this repo fixed the day before: the file needed to be `.claude/skills/resolve-tighten-recording/SKILL.md`, not `.claude/skills/tighten-recording.md`. Claude Code only discovers skills at `<name>/SKILL.md`; a loose `.md` at the top level of that directory silently never loads. Ten skills sat in exactly that state until 0a318a5 (v2.80.2) moved them, which your branch predates. Merged as-is it would have been a well-measured skill that no agent ever saw.

The content is what made this easy to take. Two things in particular:

- The source-vs-record coordinate table is the right centerpiece. "Correct cut lengths, wrong content, no error" is the failure mode that survives every verification a careful person would think to run, and it deserves to be the first thing an agent reads.
- Verifying audio by item count (113 + 113 = 226) rather than by the success envelope, and marking cross-restart `confirm_token` as unverified instead of assuming it. Both are the habit this repo is trying to hold.

I also added a pointer from `resolve-edit`'s `edit_engine` bullet to this skill, and put both assembly recipes in the `docs/README.md` skill index — neither `resolve-rough-cut` nor this one had ever been listed there.

Your cross-reference to #125 is now the only place the whisper blind spot is documented for someone driving a tighten; that issue is still open and I've replied to it separately.

Issue #129, opened 2026-08-06
Free-edition in-app bridge confirmed working on Linux (Fedora 43, Resolve free 20.3.2.9)


christag:
The README's bridge section notes validation on macOS (free 21.0.3.7, Studio 19.1.3.7) and Windows 11 (#109, #112), with no Linux reports. Here's one: the bridge installs, lists, and serves end-to-end on Linux.

## Environment

| | |
|---|---|
| OS | Fedora Linux 43 (Kinoite, immutable/ostree), kernel 7.0.9 |
| Resolve | **Free** 20.3.2.9, direct download at `/opt/resolve` |
| davinci-resolve-mcp | v2.78.1 |
| System Python | `/usr/bin/python3` → 3.14.4 |
| MCP client | Claude Code (stdio), `DAVINCI_RESOLVE_BRIDGE=1` |

## What happened

1. `scripts/install_resolve_bridge.py` deployed to `~/.local/share/DaVinciResolve/Fusion/Scripts/Utility/` — no warnings, nothing skipped.
2. After a Resolve restart with a project open, the `.py` scripts appeared in **Workspace ▸ Scripts** (no framework-Python analog needed on Linux — Resolve enumerated them against the system 3.14 interpreter directly).
3. `resolve_bridge_probe` ran twice; `resolve_bridge` started and held `127.0.0.1:49632` (listener owned by a `ScriptState` process).
4. External MCP calls through the bridge succeeded: `resolve_control.get_version` returned 20.3.2.9, and `project_manager.get_current` read the live open project by name.

## host-model-probe.json

```json
{
  "blocking_required": true,
  "edition": "free",
  "interpreter_persisted": false,
  "model": "child_process",
  "parent_corroborates": false,
  "parent_name": "GUI Thread",
  "parent_pid": 762639,
  "pid": 763040,
  "previous_run_in_this_interpreter": null,
  "product": "DaVinci Resolve",
  "resolve_object_available": true,
  "self_executable": "/usr/bin/python3",
  "self_is_resolve": false,
  "version": "20.3.2.9"
}
```

Notable for the host-model map: Linux runs Workspace scripts as a **child process** (`model: child_process`, `self_executable` = system python), unlike the macOS in-process behavior the probe was built around, and the bridge works fine under that model.

Happy to re-run the probe or the differential harness on future builds if Linux data is useful. I'll follow up with a small PR updating the README's validation paragraph to include this report.

[...]

Issue #109, opened 2026-08-03
Bridge on Resolve Free


alpaolo:
Resolve 21 free with cursor: 
MCP connected
Bridge connected:
======================================================================
DaVinci Resolve MCP - in-app bridge
======================================================================
  product     : DaVinci Resolve 21.0.1.11
  listening   : 127.0.0.1:49632
  config      : C:\ProgramData\Blackmagic Design\DaVinci Resolve\Fusion\.davinci_mcp_runtime\bridge.json
  host model  : child_process (blocking=True)
  operations  : 15
  generation  : 0

  The MCP server connects with DAVINCI_RESOLVE_BRIDGE=1.
  This script now HOLDS the listener open and will look busy until
  Resolve exits. That is expected — closing it stops the bridge.
======================================================================

But the control panel say: "Resolve unavailable"

samuelgursky:
Thanks — and the detail that made this diagnosable was pasting the bridge banner alongside the panel's message. A working bridge and a panel that says "Resolve unavailable" in the same breath is the whole bug, and it wouldn't have been obvious from either half on its own.

## What's happening

The control panel is a separate process from the MCP server, and it has its own Resolve connector. That connector did this:

```python
try:
    import DaVinciResolveScript as dvr_script
except Exception as exc:
    return None, f"Resolve scripting API unavailable: {exc}"
```

On the free edition that import is exactly what fails — Blackmagic's module ships with the installer, not the App Store build, so there's no `Developer/Scripting/Modules` tree to import from. The panel returned right there, before ever calling `connect_resolve`, whose whole point is that it accepts `None` in bridge mode:

> `dvr_script` may be None in bridge mode: the bridge does not need Blackmagic's module at all, which is precisely why it reaches editions the module cannot.

So the panel could never connect on the free edition no matter how healthy your bridge was. Your bridge is fine — the banner you pasted shows it listening, and the MCP server reaching Resolve confirms it end to end. It was the panel asking the wrong question.

The MCP server already carried a guard against this exact failure; the panel's connector was missed. That makes it the third connector to be overlooked when a transport is added, so the fix is scoped accordingly.

## Fixed on `main` in 06e4278

The panel now consults the bridge first and treats both the environment setup and the module import as optional when it's enabled — the same ordering the MCP server uses. Four tests cover it, including one that reproduces your symptom by blocking the import with the bridge switched on. Suite 2317 → 2321.

[...]

PR #130, opened 2026-08-06
docs: record Linux validation of the free-edition in-app bridge

christag:
The README's bridge section tracks per-platform validation status: macOS validated at launch, Windows confirmed via #109/#112, Linux silent. #129 now provides a Linux report — free 20.3.2.9 on Fedora 43, bridge installing to `~/.local/share/DaVinciResolve/Fusion/Scripts/Utility`, listing against the system Python (3.14), and serving end-to-end through `DAVINCI_RESOLVE_BRIDGE=1` with live reads of the open project.

This PR adds one sentence to the validation paragraph so all three platforms' status is on record, matching the pattern used when the Windows paths were confirmed. Probe JSON and environment details are in #129.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,12 @@ Validated on free 21.0.3.7 and Studio 19.1.3.7, both macOS. The Windows paths
 added in v2.70.1 (issue #106) shipped unverified; reports on free 21.0.1.11
 (issue #109) and free 21.0.3.7 (issue #112) have since shown the bridge
 installing, listing and serving from **both** `%PROGRAMDATA%` and `%APPDATA%` on
-Windows 11, so those paths are now confirmed rather than assumed.
+Windows 11, so those paths are now confirmed rather than assumed. Linux is
+confirmed as well: a report on free 20.3.2.9 (issue #129, Fedora 43) shows the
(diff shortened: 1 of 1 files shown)


samuelgursky:
Merged as 6f881cc, shipped in [v2.83.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.83.0). Landed by cherry-pick because a branch ruleset blocks the merge button; the commit keeps your authorship, so GitHub will show this as closed rather than merged.

I checked the substantive claim rather than taking the paragraph on trust: `install_resolve_bridge.py:162` targets `~/.local/share/DaVinciResolve/Fusion/Scripts/Utility`, exactly the path #129 reports serving from. Matches.

One line changed on merge. The paragraph ended "all three platforms now rest on reports rather than assumptions", which undersells macOS — that one was validated directly here, not by user report. Now: "No platform now rests on an assumption: macOS was validated directly, Windows and Linux on user reports." Same fact, correct provenance per platform.

The detail in #129 that earns its place beyond the README line is the host model: Linux runs Workspace scripts as a **child process** with `self_executable` = system python, where the probe was built around macOS in-process behaviour — and the bridge works fine under it. That is the kind of thing that would otherwise be discovered by someone debugging a failure that isn't one. Thanks for including the probe JSON.

Issue #131, opened 2026-08-06
Resolve 21.0.4 API coverage: MediaPoolItem.GetTimeline, Timeline.GetSelectedClips, new SetRenderSettings keys


legionsound:
Resolve 21.0.4 (released 2026-08-05/06) added three scripting-API surfaces, per the shipped `Developer/Scripting/README.txt`:

1. `Timeline.GetSelectedClips()` → currently selected timeline items (README line ~400)
2. `MediaPoolItem.GetTimeline()` → timeline object for a Media Pool timeline entry (README line ~309)
3. New `SetRenderSettings` keys: `UseFullExtents` (Bool), `AddFrameHandles` (int >= 0, ignored when full extents enabled), `DataBurnIn` (string, e.g. "Same as project" / "None") (README ~863-866)

## Current state in v2.82.1

- `_get_selected_timeline_items` already duck-probes `GetSelectedTimelineItems`, `GetSelectedItems`, `GetSelectedClips` — so (1) works by luck via the third probe name. Suggest putting the now-documented `GetSelectedClips` first.
- `MediaPoolItem.GetTimeline` has no references anywhere in the codebase — (2) is unreachable through the `media_pool_item` tool.
- The three render keys are absent from `_RENDER_SETTING_KEYS`, so `validate_render_settings` reports them as `unknown_keys` and they get no type validation.

## Live evidence (Studio 21.0.4.5, macOS)

Probe via the bundled venv against a real project:

```
version: 21.0.4.5
MediaPoolItem.GetTimeline exists: True
MediaPoolItem.GetTimeline() -> PyRemoteObject name: <timeline name>
SetRenderSettings({'UseFullExtents': False, 'AddFrameHandles': 0, 'DataBurnIn': 'None'}) -> True
Timeline.GetSelectedClips() -> callable, returns list (empty when nothing selected)
```

## Local patch

I've patched my local checkout (branch off v2.82.1) with:
- `media_pool_item(action="get_timeline", {clip_id})` → `{is_timeline, timeline: {name, unique_id, start_frame, end_frame}}`, with a clear error on pre-21.0.4 builds
- selection probe reordered to try `GetSelectedClips` first
- the three keys added to `_RENDER_SETTING_KEYS` + typed validation (`AddFrameHandles` int >= 0, `UseFullExtents` bool, `DataBurnIn` string)

[...]

PR #142, opened 2026-08-08
feat(granular): bring the twelve 21.0.4 surfaces to the granular server

### LINKED ISSUE #140: Granular surface: the 21.0.4 preset-enumeration methods are compound-only
The 21.0.4 preset-enumeration surfaces land on the **compound** tools only. On the granular surface the discoverability hole they were filed to close is still open.

## Current state

`src/granular/resolve_control.py` already wraps most of the preset family:

- layout presets: `SaveLayoutPreset`, `LoadLayoutPreset`, `UpdateLayoutPreset`, `ExportLayoutPreset`, `ImportLayoutPreset`, `DeleteLayoutPreset`
- burn-in presets: `ImportBurnInPreset`, `ExportBurnInPreset`

Every one of those takes a **preset name**, and none of the granular tools can enumerate names. That is exactly the complaint in #136 and #137 — closed on the compound surface by #139, untouched here.

Not wrapped anywhere in `src/granular/`:

| Method | Compound equivalent (shipped) |
|---|---|
| `Resolve.GetLayoutPresetList` | `layout_presets list` |
| `Resolve.GetBurnInPresetList` | `render_presets list_burnin` |
| `Resolve.DeleteBurnInPreset` | `render_presets delete_burnin` |
| the six `*UserPreferencesPreset` methods | `resolve_control *_user_preferences_preset` |
| `ProjectManager.GetProjectAttributesInCurrentFolder` | `project_manager list_attributes` |

`MediaPoolItem.GetTimeline` and `Timeline.GetSelectedClips` from #131/#133 are in the same position — compound-only. (The `GetSelectedClips` hits under `src/granular/` are `MediaPool.GetSelectedClips`, a different class.)

## Not a regression

This is the established pattern, not something #139 broke: #133 also shipped compound-only, and the granular surface has been trailing the 21.0.4 delta since then. Filing it so the gap is tracked rather than rediscovered.

## Scope note

`test_doc_tool_counts` statically counts `@mcp.tool` decorators in `src/granular/*.py` and asserts the number quoted across README, `docs/contributing.md`, `docs/SKILL.md`, `docs/reference/api-coverage.md`, `resolve-advanced/README.md`, and `.github/copilot-instructions.md`. Adding granular wrappers moves the count off 341, so those six docs need updating to whatever the guard prints in the same change.

Same `_requires_method`-style 21.0.4 guarding as the compound actions, and the `LoadUserPreferencesPreset` blast-radius warning has to be carried over — it swaps the user's global preferences session-wide.

legionsound:
Closes #140.

All twelve 21.0.4 methods that were compound-only now have granular equivalents, in the existing granular idiom (module-level tools, `get_resolve()` / `_get_mp()` / `_get_timeline()` helpers, `_requires_method` guards at 21.0.4):

| Granular tool | Wraps | File |
|---|---|---|
| `get_layout_preset_list` | `Resolve.GetLayoutPresetList` | `resolve_control.py` |
| `get_burn_in_preset_list` / `delete_burn_in_preset` | `GetBurnInPresetList` / `DeleteBurnInPreset` | `resolve_control.py` |
| six `*_user_preferences_preset` tools | the six `*UserPreferencesPreset` methods | `resolve_control.py` |
| `get_project_attributes_in_current_folder` | `PM.GetProjectAttributesInCurrentFolder` | `project.py` |
| `get_clip_timeline` | `MediaPoolItem.GetTimeline` | `media_pool_item.py` |
| `get_selected_timeline_items` | `Timeline.GetSelectedClips` | `timeline.py` |

The issue scoped the ten preset/attribute methods and noted the two #131 stragglers were "in the same position" — I took all twelve so the whole 21.0.4 delta closes on granular in one pass. Trim the last two if you'd rather keep them separate.

Both carried-over items from your #140 comment are in:

- **`load_user_preferences_preset` has the SESSION-WIDE warning** in its docstring, same wording class as the compound action. It was not executed during verification. `import_user_preferences_preset` carries the not-auto-loaded note and the takes-its-name-from-the-file behavior measured on 21.0.4.5.
- **The doc-count guard: 341 → 353 everywhere.** README (including the badge), `docs/SKILL.md`, `docs/reference/api-coverage.md`, `docs/contributing.md`, `.github/copilot-instructions.md` — plus three `341-tool` mentions inside `src/server.py` the guard doesn't check, and the six generated agent-rule files via `scripts/agent-rules/generate.mjs` (its own drift guard was the thing that caught them).

`get_selected_timeline_items` is deliberately not named `get_selected_clips` — that name is taken by the MediaPool selection tool, which is the exact confusion the issue flagged.

## Evidence (Studio 21.0.4.5, macOS)

Through the granular tools themselves:

```
get_layout_preset_list                    -> {"presets": []}
save_layout_preset_tool("MCP_GRAN_VERIFY")-> {"success": true}
get_layout_preset_list                    -> {"presets": ["MCP_GRAN_VERIFY"]}
delete_layout_preset_tool(...)            -> {"success": true}; list -> []
get_burn_in_preset_list                   -> {"presets": []}
save/get/delete user_preferences_preset   -> full round-trip, list tracks state exactly
get_project_attributes_in_current_folder  -> 8 projects, all four attribute keys
get_clip_timeline(timeline entry)         -> {"is_timeline": true, "timeline": {"name": "...", 12,459 frames}}
get_clip_timeline(ordinary clip)          -> {"is_timeline": false, "note": "...not a timeline entry."}
get_selected_timeline_items               -> {"count": 0, "items": []}   # empty selection is an answer
```

One incidental observation while verifying: immediately after a cold Resolve launch, `GetProjectAttributesInCurrentFolder` returns `{}` for ~3 minutes — the same early-boot window where `GetProjectListInCurrentFolder` returns `[]`, so it's the known startup behavior, not a quirk of the new method. Both populated together (8/8) once the database finished attaching.

## Tests

`tests/test_granular_2104_parity.py` — 17 stub contract cases: answers, guards, the single-arg import form, the SESSION-WIDE docstring presence, and the empty-selection answer.

Full suite: **2560 tests OK, 27 skipped**, via `venv/bin/python -m unittest discover -s tests`. That includes your `test_doc_tool_counts` and the agent-rules drift guard passing at the new counts.


samuelgursky (dismissed):
Approving. The twelve tools are right, and I verified the count against the runtime registry rather than the decorators alone: `asyncio.run(mcp.list_tools())` returns **353** with zero duplicate names, all twelve present and `_requires_method`-guarded at 21.0.4. Return shapes match their compound counterparts (`{"presets": [...]}`), and `_requires_method` does reach the granular modules through `from src.granular.common import *` — that `__all__` excludes only dunders, so the single-underscore helper comes through. Offline suite reproduces at 2560 OK here.

Taking all twelve rather than ten was the right call. `get_selected_timeline_items` over `get_selected_clips` likewise — the collision with the Media Pool tool is real.

## One thing to fix before merge: four stale `341`s, one of which fails

`unittest discover` does not collect `tests/test_import.py` — it is pytest-style, and the file says so at line 90. So its `assert total == 341` survived your green 2560-test run. It fails when run as a script, which is exactly how the publish gate runs it (`.github/workflows/npm-publish.yml` line 46, and step one of `docs/process/release-process.md`). It would have landed as a release-time failure, not a merge-time one.

Three more the doc-count guard does not look at:

- `src/resolve_mcp_server.py:37` — logs `"341 granular tools"` at every startup
- `docs/install.md:136,151` — and line 135 says **32** compound, stale since the compound server reached 34
- `README.zh-CN.md:8,79,193` — Tools badge, server-modes table, metrics table

None of this is on you: the guard exists precisely to catch this class and its file list was incomplete. Worth widening in the same PR so the next count change fails offline. Suggested additions to the `checks` list in `tests/test_doc_tool_counts.py`:

```python
("README.md", f"MCP%20Tools-{comp}%20({gran}%20full)"),
("README.zh-CN.md", f"MCP%20Tools-{comp}%20({gran}%20full)"),
("README.zh-CN.md", f"`src/resolve_mcp_server.py` | {gran} |"),
("README.zh-CN.md", f"**{comp}** 复合 / **{gran}** 细粒度"),
("docs/install.md", f"`src/server.py` | {comp} |"),
("docs/install.md", f"`src/resolve_mcp_server.py` | {gran} |"),
("docs/install.md", f"full {gran}-tool server"),
("src/resolve_mcp_server.py", f"({gran} granular tools)"),
("tests/test_import.py", f"assert total == {gran}"),
```

The English README carries the same badge the Chinese one does and was not checked either; yours happened to be correct.

I have this as a commit against your branch — with it applied: full suite 2560 OK, `python tests/test_import.py` exits 0, `node scripts/agent-rules/generate.mjs --check` in sync. It will land on the branch shortly if you do not get there first.

## Two observations, neither blocking

**Selection shape.** `get_selected_timeline_items` does `tl.GetSelectedClips() or []` and iterates. The compound path routes the same call through `_coerce_item_list` (`src/server.py:3535`), which explicitly handles a dict return. If Resolve hands back a dict, the granular tool reports the right `count` with every entry an empty `{}`. Your live evidence only exercised the empty selection, so the populated shape is stub-only either way — not something to chase without a 21.0.4 box to measure on.

**Pre-21.0.4 divergence.** Compound `_get_selected_timeline_items` falls through two legacy names and then `GetCurrentVideoItem` with a warning; the granular tool hard-errors at the version guard. Defensible, since granular is one-tool-per-method by design — noting it because the two servers now answer the same question differently on older builds.

On evidence: the 21.0.4.5 results are yours. The Resolve here is 19.1.3.7, so I verified the guard paths through your stubs, not the live calls — the coverage doc already records these as reporter-measured and that stays accurate.

samuelgursky (approved):
Re-approving at e8ab4b1 — the count patch is on the branch now. Re-verified against this head: full offline suite 2560 OK, `python tests/test_import.py` exits 0, `node scripts/agent-rules/generate.mjs --check` in sync, and the widened doc-count guard passes over README.md, README.zh-CN.md, docs/install.md, src/resolve_mcp_server.py and tests/test_import.py. Earlier review stands otherwise.

PR #144, opened 2026-08-09
Report source_fps beside source frames (a WAV counts at 24 fps on any timeline)

rusanivsky:
An audio timeline item's `source_start`/`source_end` are counted in the **media's** frame rate, not the timeline's — and a WAV carries no native rate, so Resolve reports **24** for it. Read back at the timeline rate, a WAV offset lands minutes from the real position in the file, and nothing errors: `probe_timeline_structure` derives `source_end` as `source_start + timeline_duration`, so the start/end pair stays internally consistent whatever rate the caller assumes.

Measured live on **Studio 21.0.3.7** against a 29.97 fps timeline: a `ZOOM0028.WAV` item at `source_start: 60687` is 2528.6 s into the file; dividing by the timeline rate says 2024.9 s — **504 s out**. Video is unaffected and reports its own 29.97.

## What this changes

Documenting the trap only helps a caller who thinks to look it up, so this adds the guard as well as the ledger entry.

**`source_fps` beside the frames.** Every timeline-item summary now carries `source_fps` plus `source_start_seconds` / `source_end_seconds` derived with it, so the frame number always arrives with its unit attached:

```
audio ZOOM0028.WAV  source_start=60687  source_fps=24.0   -> 2528.625 s
video IMG_0001.mov  source_start=20379  source_fps=29.97  ->  679.98 s
```

The rate is read from the media-pool item's `FPS` property, never assumed — an unreadable rate reports `null`, so callers see *unknown* rather than a guess.

**No extra bridge calls on the probe path.** `_timeline_item_conform_summary` already pulled the full property dict for every item (for `file_path` / `media_status`), so it now fetches the media-pool item and its properties once and hands both to the summary instead of re-fetching.

**A second unit trap, found while pinning the rate.** The rate was established by regression rather than assumed: across 12 items of the same WAV, `GetSourceStartFrame` advances at **24.000 fps** against the item's own `GetSourceStartTime` (24.0000 / 24.0007 / 23.9995 over spans up to 22 minutes). The same measurement showed that on an audio item `GetLeftOffset` advances at **29.970** — the *timeline* rate. The two readers describe the same edit point in different frame spaces (60687 vs 75784 for one item); on video they share the source space.

`_timeline_item_source_start` falls back to `GetLeftOffset`, so that fallback would have paired a timeline-frame number with the media rate. `_timeline_item_source_start_with_origin` now reports which reader produced the value, and the summary reports `source_fps: null` on the audio fallback rather than converting it wrong.

## Also here

- `create_variant_from_ranges` accepts an undocumented per-range `track_index` (1-based within `track_type`; missing tracks are added, so multicam angles rebuild onto V2/V3 instead of collapsing onto V1). It was absent from the action help — now documented there, in the action list, in the example, and in `docs/SKILL.md`.
- The trap is recorded in the `api_truth` ledger (`submit: bug`, so it reaches `api-limitations.md`) and in `resolve-rough-cut`'s Verified traps table.

## One caveat, recorded rather than papered over

Resolve's model of the test WAV is 133003 frames (`Duration 01:32:21:19` at 24 fps = 5541.79 s) while its true PCM length is 266264768 samples / 48 kHz = **5547.18 s** — a 0.097% difference I could not explain. So `frames / source_fps` is exact in Resolve's source-time space, which is the space every other Resolve call uses, but may sit ~2 s off the byte position at a 40-minute-deep offset. That caveat is in the ledger entry.

## Validation

- Full offline suite: **2546 tests, OK.** (4 pre-existing loader errors on `main` from a missing `numpy` in my venv — `test_colorimetry`, `test_custom_timeline`, `test_image_qc`, `test_improvements` — untouched by this branch.)
- 13 new cases in `tests/test_source_frame_rate.py`, including the "unreadable rate must read as unknown, not 24" case and both fallback-unit cases.
- `scripts/gen_api_limitations.py --check`: clean.
- Verified live on Studio 21.0.3.7 before and after.


samuelgursky:
Merged and shipped in [v2.91.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.91.0). Thank you — the regression against `GetSourceStartTime` to *pin* 24 rather than assume it is what made this reviewable, and the `source_fps: null` on the `GetLeftOffset` fallback is the right call.

Two follow-ups landed on top, both worth flagging back to you.

**`source_end_seconds` carried the same unit mix-up the field exists to stop.** `source_end` is `source_start + duration`, and that `duration` is `GetDuration` — a *timeline* duration. So it is already unit-mixed the moment the rates differ, exactly as your PR body says; dividing it by `source_fps` then converts the mixed number. The seconds now come from `GetSourceStartTime`/`GetSourceEndTime` (seconds, no rate inference at all), then `GetSourceEndFrame / source_fps`, and read `null` rather than convert the derived value. `source_end` itself is unchanged so no consumer moves, but it is annotated as unit-mixed where it is assigned.

**Re-measured on Studio 19.1.3.7** with synthetic media, so this is not a 21.x regression — and the older build gives a clean demonstration of the conversion happening in the open:

```
300 s 48 kHz WAV                     -> FPS 24
append source frames 4800-5235       -> timeline duration 543  (= 435 x 29.97/24)
source_end (derived)                    5343   -> /24 = 222.625 s
GetSourceEndTime                                        218.133 s   <- true
GetSourceStartTime                                      200.000 s   (= 4800/24, exact)
```

4.49 s out on a clip 18.1 s long. The matching 29.97 video item was unaffected either way (24.524 read vs 24.525 derived), which is the video-is-fine half of your finding reproducing independently. Both second-readers exist on 19.1.3.7, so the `GetSourceEndFrame` fallback is only for builds older still.

Your 21.0.3.7 numbers are in the ledger entry labelled as yours; I have not re-run them here, this machine is on 19.1.3.7.

One thing I did **not** change, noted in case you want it: `extract_source_frame_ranges` builds pull ranges from that same derived `source_end`, so for a mixed-rate audio item it over-pulls by the rate ratio. Safe direction (a longer range), but wrong, and it is the same root cause.

samuelgursky:
Follow-up you should know about, because it corrects part of the entry your PR added. Shipped in [v2.93.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.93.0).

## A WAV is not 24 fps

The ledger said a WAV "carries no frame rate, so Resolve falls back to 24", and told callers to treat one as 24 fps. Measuring it directly on 19.1.3.7 with one 400.000 s file:

```
project at 24    -> import -> FPS 24.0    Duration 00:06:40:00   (9600 frames  = 400 s)
project at 29.97 -> import -> FPS 29.97   Duration 00:06:39:18   (11988 frames = 400 s)
project A: set 29.97 AFTER import         -> clip still reads 24.0
```

A WAV takes the **project's `timelineFrameRate` at import** and freezes it. So the trap is not "audio is always 24" — it is "the clip kept the rate the project had when it was imported, and the project moved afterwards." Your 21.0.3.7 reading of 24 is fully consistent with this; that project was at 24 when the WAV came in. But a WAV imported into a 29.97 project has no mismatch at all, and anyone who followed the old advice on such a file would have converted a correct number into a wrong one. Every site now says read `source_fps`, never assume.

Nothing in your measurement was wrong — the 56871 / 24 = 2369.6 s arithmetic holds. It was the *explanation* attached to it that generalised one project's rate into a property of the format.

## And the one you flagged in the PR body

You wrote that `source_end` is derived as `source_start + timeline_duration` and stays "internally consistent whatever rate the caller assumes." That is exactly right, and it was worse than internally consistent — it was wrong by up to 149 frames against the `endFrame` actually sent, which `extract_source_frame_ranges` then turned into pull ranges 543 and 749 frames wide for clips consuming 435 and 600.

`source_end` is now `round(GetSourceEndTime × source_fps)` — a source frame by construction, still exclusive, exact on every item measured in both regimes, and identical to the old value wherever the old value was already right.

Worth recording: `GetSourceEndFrame` was the obvious fix and it is **not** usable raw. Over 12 items it reads **exclusive when the source and timeline rates match, inclusive when they differ** — off by one in precisely the case you would reach for it. Not a media-type split; the same WAV imported at 29.97 into a 29.97 timeline reads exclusive like video.

One process note, since your PR set the standard for it: my first measurement pass zipped `AppendToTimeline`'s return against the request list, two entries came back unreadable, and the misalignment produced numbers that looked exactly like reader noise — I nearly wrote up a per-media-type convention from them. Everything above comes from appending one range at a time. Thanks again for the original find; it has now paid for itself twice.

PR #145, opened 2026-08-12
feat: add hardened social delivery workflow actions

justin-lawes:
## Summary

- add six compound social-delivery workflow actions across project manager, timeline, color, and render APIs
- add detailed timeline inventory and per-source-FPS exposure planning
- harden resumed render identity, strict output QC, recoverable grade backups, and source-safe cover exports
- document the workflow contract and migrate the external `ig-post` skill to use the compound actions
- merge current upstream v2.93 into the feature branch without rewriting published history

## Safety and behavior

- mutating setup/render actions default to `execute=false`
- bulk DRX/CDL application creates and verifies local grade versions before replacing any graph
- queued render jobs are never started during dry-run; resumed jobs must match timeline and delivery identity
- cover still exports refuse overwrite, source conflicts, unsupported extensions, and paths outside the analysis root
- markerless timelines receive deterministic uniform cover samples

## Verification

- focused social-delivery and adjacent Render/Deliver contracts: 89 passed
- current upstream baseline: 2,556 tests; 22 failures, 69 errors, 49 skips
- integrated feature branch: 2,609 tests; 22 failures, 69 errors, 49 skips
- exact failing-test comparison: `FAILURE_SET_DELTA=none` (91 identities on each tree)
- `git diff --check`: clean
- external `ig-post` skill: 70 tests passed and official skill validator passed; committed separately as `justin-lawes/claude-skills@0395231`

The remaining full-suite failures/errors are pre-existing Windows/environment baseline failures, not feature-only identities.

## Live validation status

Validated successfully on Windows against DaVinci Resolve Studio 21.0.4.5. The repaired fresh-process Render/Deliver harness connected to Resolve, created a disposable project, generated synthetic media, created a timeline, exercised 19 capability checks, and cleaned up the project, media, and render output.

Results: 15 supported, 3 partially supported, 1 version/page dependent, 0 unsupported, and 0 errors. No user source media was touched.

The v2.93 stale MCP import stub encountered during preparation is repaired and covered by a fresh-process regression test.

## Review

An independent Codex review was completed. Its findings were addressed, including resume safety/identity, strict QC semantics, recoverable grade backups, safe cover export boundaries, timeline activation checks, per-source FPS, disabled clips, and markerless sampling. A Claude review was attempted but its local OAuth session had expired.


samuelgursky:
Reviewed this properly — ran both trees locally and read the whole `server.py` diff. The suite claim holds, and it holds on a platform you didn't test: macOS gives 2610 on `main` and 2663 here, same two pre-existing failures on both, so `FAILURE_SET_DELTA=none` reproduces independently of your Windows baseline. The shape of the work is right — dry-run defaults, confirm tokens matching the existing `apply_grade_from_drx` gate, structured codes, and the cover export refusing overwrite, source-path collision, and anything outside the analysis root. That last one is the kind of guard that usually gets added after something goes wrong, not before.

Four things before I can merge it. I have 21.0.4.5 Studio here now, so the ones that needed a live Resolve are answered rather than asked.

**1. `poll_interval_seconds=0` busy-spins the Resolve API.** `max(0.0, float(...))` permits zero and `if poll_interval:` then skips the sleep. I measured it against a fake project: **1,787,013 `GetRenderJobStatus` calls in 1.0s**. With the default `max_wait_seconds=300` that is five minutes of unthrottled IPC into a Resolve that is mid-render. Please floor the interval, and put a ceiling on `max_wait_seconds` too — it is unbounded user input on a blocking call.

**2. The `JobName` fallback fabricates an output path.** `_render_job_output_path` falls back to `job.get("JobName")` as a filename, so a job named `social_master` targeting `/deliveries` produces `/deliveries/social_master` — no extension, no such file. `_delivery_output_qc` then returns `OUTPUT_FILE_MISSING` / `status: "failed"` at a path that was never an output. A render that succeeded gets reported as a failed delivery, pointing at a file that never existed. `JobName` is a display name; it should not be in that chain. Dropping it lets the honest `OUTPUT_PATH_UNAVAILABLE` branch do its job.

**3. Zero checks reports `passed`.** `_delivery_output_qc` ends on `all(row["pass"] for row in checks)`, and `all([])` is `True` — a spec that yields no comparable fields returns `success: True, status: "passed"`. I confirmed it with `{"video": {"bitrate": "10M"}}`. It is not reachable through the shipped targets today, since `to_qc_spec` always writes `codec` into `video`, so this is latent rather than live. I still want it closed, because it is the top-level `success` gate of `complete_delivery_job` and "QC passed" has to mean QC ran.

Patch for those three:

```diff
--- a/src/server.py
+++ b/src/server.py
@@ _delivery_output_qc
-    return {"success": all(row["pass"] for row in checks), "status": "passed" if all(row["pass"] for row in checks) else "failed", "path": path, "checks": checks, "probe": probe, "spec": spec}
+    if not checks:
+        return {"success": False, "status": "unavailable", "code": "QC_NO_COMPARABLE_FIELDS",
+                "path": path, "checks": checks, "probe": probe, "spec": spec}
+    passed = all(row["pass"] for row in checks)
+    return {"success": passed, "status": "passed" if passed else "failed",
+            "path": path, "checks": checks, "probe": probe, "spec": spec}

@@ _render_job_output_path
-    filename = job.get("OutputFilename") or job.get("OutputFileName") or job.get("JobName")
+    # JobName is a display name, not a filename: using it invents a path that
+    # then fails isfile(), reporting a good render as a failed delivery.
+    filename = job.get("OutputFilename") or job.get("OutputFileName")

@@ _render_complete_delivery_job
-    max_wait = max(0.0, float(p.get("max_wait_seconds", 300.0)))
-    poll_interval = max(0.0, float(p.get("poll_interval_seconds", 1.0)))
+    max_wait = min(3600.0, max(0.0, float(p.get("max_wait_seconds", 300.0))))
+    # Floored: a zero interval spins GetRenderJobStatus ~1.8M times/sec at a
+    # live Resolve for the whole max_wait window.
+    poll_interval = min(60.0, max(0.05, float(p.get("poll_interval_seconds", 1.0))))
```

**4. The grade backup is labelled backwards — verified live.** I installed 21.0.4.5 Studio and ran `_timeline_apply_drx_and_cdls_bulk` against a disposable project with synthetic media, on a clip whose working version carried a distinguishable grade. Result:

```
before:             active='Version 1'  versions={'Version 1': QUALIFIER(original)}

after PR function:  active='MCP pre-DRX 20260812-160505 a4c592c1'
                    versions={'Version 1':                          QUALIFIER(original),
                              'MCP pre-DRX 20260812-160505 a4c592c1': WINDOW(applied drx)}
```

The version named `MCP pre-DRX <timestamp>` is the one holding the post-DRX grade, and the clip is left active on it.

Two undocumented behaviors cause it, both confirmed on 21.0.4.5:

- `AddVersion` **switches the active version** to the newly created one, and the new version inherits the current grade. The API reference says only "Adds a new color version."
- A `Graph` handle **follows the item's current version**, not the version it was captured from. Your handles are captured in the resolution loop, before the backup loop, so they end up writing to the backup version.

Credit where it is due: nothing is destroyed. The original grade survives in `Version 1`, and `grade_snapshot.current` records it correctly, so the response contains the truth. But a colorist undoing this batch will restore the version labelled "pre-DRX" and get the DRX, which is worse than an honest error. Switching back after the backup fixes it, verified in the same run:

```
after corrected:    active='Version 1'
                    versions={'Version 1':          WINDOW(applied drx),
                              'MCP pre-DRX PROBE':  QUALIFIER(original)}
```

```diff
@@ _timeline_apply_drx_and_cdls_bulk, backup loop
         row["grade_snapshot"] = _grade_version_snapshot(item, p)
+        current_version = (row["grade_snapshot"].get("current") or {}).get("versionName")
         backup_name = f"MCP pre-DRX {time.strftime('%Y%m%d-%H%M%S')} {_uuid.uuid4().hex[:8]}"
@@ after the created/verified check
         row["grade_backup"] = {"name": backup_name, "type": 0, "created": created, "verified": verified}
+        # AddVersion makes the NEW version current and copies the grade into it,
+        # and a Graph handle follows whatever version is current. Without this
+        # switch-back the DRX lands on the version named "pre-DRX" and the
+        # backup label points at the wrong grade. Verified on 21.0.4.5.
+        if not bool(item.LoadVersionByName(current_version, 0)):
+            return {"success": False, "dry_run": False, "code": "GRADE_BACKUP_RESTORE_FAILED",
+                    "timeline_item_id": item_id, "grade_backup": row["grade_backup"], "items": []}
```

Since handles follow the current version, re-resolving `row["graph"]` inside the apply loop rather than in the resolution loop would make that ordering explicit instead of load-bearing. Worth doing while you are in there.

Two smaller things, not blocking: `rank_cover_candidates` drops malformed samples with a bare `continue`, so a partial ranking reads as a complete one — worth surfacing a count. And `qc_spec = p.get("qc_spec")` is overwritten on both branches, so it is a knob that does nothing.

Two asks outside the code.

**`docs/superpowers/plans|specs` needs to come out.** This repo keeps in-flight implementation plans — task checklists, per-task file lists, sub-skill directives — in the gitignored `local/`, not in `docs/`. It is a durability call: they are accurate for about a week and then mislead whoever reads them next. Nothing against how you are working, and the `/.worktrees/` gitignore entry is welcome; it is specifically the 983 lines of checkbox plans I do not want tracked. Design rationale that outlives the change is very welcome in `docs/` — just written as documentation rather than as a plan.

**Please narrow the live-validation claim.** The harness has no reference to `apply_drx_and_cdls_bulk`, `cover_frame_candidates`, `ensure_project_timeline`, or `complete_delivery_job` — the diff to `live_render_deliver_validation.py` is entirely MCP import stubs. So the Studio 21.0.4.5 run validated the harness repair and the pre-existing capability probe, which is real and useful work, but under a "Live validation status" heading it reads as covering these six actions. Finding 4 is what that gap costs: it survived a full unit suite, a Codex review, and a live capability run, because the only thing that could catch it was executing the function against a real graded clip. Please narrow the wording, and consider whether the bulk DRX path can join the harness.

Ping me when those are pushed and I will take another pass.

Issue #150, opened 2026-08-18
runtime_mode reports running: false on Windows — WMIC quotes the command line, defeating endswith


7daysdedicated:
## Summary

On Windows, `resolve_control runtime_mode` reports `running: false, instances: 0` while DaVinci Resolve is running and the very same server is successfully driving it through the scripting API.

The cause is that WMIC wraps command lines in double quotes when the executable path contains spaces — which is true of every default Windows install (`C:\Program Files\Blackmagic Design\DaVinci Resolve\Resolve.exe`). `_is_resolve_command()` requires the line to *end with* `Resolve.exe`, but it ends with `Resolve.exe"`.

## Environment

- DaVinci Resolve Studio 21.0.4.5 (external scripting = Local)
- davinci-resolve-mcp 2.97.3, managed install, venv Python 3.14.5
- Windows 11 Pro 26200, default install path

## Symptom

```
resolve_control get_version   -> DaVinci Resolve Studio 21.0.4.5   (connected fine)
project_manager get_current   -> "Proxmox setup"                   (connected fine)
resolve_control runtime_mode  -> {running: false, instances: 0, determinable: true}

tasklist                      -> Resolve.exe   PID 64716           (it is running)
```

Note `determinable: true` — this is not the "cannot tell" path. The process list was read successfully; nothing matched.

## Root cause

`src/utils/resolve_runtime.py:112`

```python
return any(text.endswith(pattern) for pattern in RESOLVE_PROCESS_PATTERNS)
```

WMIC returns:

```
"C:\Program Files\Blackmagic Design\DaVinci Resolve\Resolve.exe"
```

Minimal repro:

```python
from utils import resolve_runtime as rr
q = '"C:\Program Files\Blackmagic Design\DaVinci Resolve\Resolve.exe"'
rr._is_resolve_command(q)            # False
rr._is_resolve_command(q.strip('"')) # True
```

The flag-stripping loop above it doesn't help: it strips trailing ` -flag` tokens, leaving the closing quote as the final character.

## Impact

Beyond the wrong reading, this appears to disable the second-instance guard in `get_resolve()`. When `_try_connect()` fails but Resolve *is* running (modal dialog, mid-launch, scripting toggled off), `resolve_is_running()` returns `False` rather than `True`, so control reaches the auto-launch branch and starts a second Resolve.

[...]

PR #154, opened 2026-08-19
fix(install): find Resolve outside the default install location

DadManBlues:
My DaVinci Resolve Studio 21.0.4 on Windows 11 is installed to F:\Blackmagic Design\DaVinci Resolve. Setup reported success, every tool then failed, and it took me a long time to find that the config contained "RESOLVE_SCRIPT_LIB": "". I fixed it by hand-editing claude_desktop_config.json before digging into why.

## What happens today

On a machine where DaVinci Resolve is not installed at the platform default path, `npx davinci-resolve-mcp setup` finishes with `Setup complete!` and produces a client config that cannot work. Every tool then fails with `SCRIPTING_UNAVAILABLE`, whose remediation text sends the user to check their Resolve edition and their External-scripting preference — both of which may already be correct.

Reported on DaVinci Resolve Studio 21.0.4, Windows 11, Resolve installed to `F:\Blackmagic Design\DaVinci Resolve\`, Claude Desktop (MSIX build), Python 3.14.7.

### The chain

1. `RESOLVE_PATHS["Windows"]["lib"]` (install.py) holds a single literal candidate:

   ```python
   "lib": [ r"C:\Program Files\Blackmagic Design\DaVinci Resolve\fusionscript.dll" ],
   ```

   It does not consult `%PROGRAMFILES%`, any other drive, or the running process. `find_resolve_paths()` returns `lib_path=None`.

2. `build_server_env()` turns that `None` into an empty string:

   ```python
   lib_value = str(lib_path or "")            # install.py
   env = { ..., "RESOLVE_SCRIPT_LIB": lib_value, ... }
   ```

   The generated config therefore contains `"RESOLVE_SCRIPT_LIB": ""`. This is the worst of both worlds: it reads as *configured* to anyone inspecting the file, and it is falsy to `DaVinciResolveScript.py`, which then falls back to the same hardcoded path that was already known to be absent.

3. Step 1 of the installer prints:

   ```
   Library:   Not found (optional — API path is sufficient)
   ```

   It is not optional. `DaVinciResolveScript` is a thin wrapper around that binary; without it nothing imports.

4. Verification fails, roughly half a screen up from the end of the output:

   ```
   IMPORT_ERROR: DLL load failed while importing fusionscript: The specified module could not be found.
                 On Python 3.13+ this may be an ABI mismatch with Resolve's scripting library —
                 try Python 3.10-3.12 if it persists.
   ```

   The Python-version hint is the first concrete suggestion offered, and it is the wrong lead here — the same message is what Windows emits when the library path simply does not exist.

5. The final line is:

   ```
   Setup complete!   Configured: Claude Desktop
   ```

The reporting user spent several hours on this, tried a second Python installation, and eventually fixed it by hand-editing `claude_desktop_config.json`. With the changes below, step 1 would have named the cause.

### Reproducing

Any machine where Resolve is not at the default path reproduces it. Without moving an installation:

```bash
# macOS/Linux — point the literal candidate at nothing
python - <<'PY'
import importlib.util
spec = importlib.util.spec_from_file_location("i", "install.py")
i = importlib.util.module_from_spec(spec); spec.loader.exec_module(i)
i.RESOLVE_PATHS[i.SYSTEM]["lib"] = ["/nonexistent/fusionscript.so"]
print(i.build_server_env("/usr/bin/python3", "/some/api", i.find_resolve_paths()[1]))
PY
# before: {'RESOLVE_SCRIPT_API': '/some/api', 'RESOLVE_SCRIPT_LIB': '', 'PYTHONPATH': '...'}
```

## What this changes

In the order the failure happens:

**Discovery from the running process** — `resolve_runtime.running_resolve_lib()` derives the scripting library from the running Resolve's own image path. When the application is up, the install location needs no guessing, and this works on all three platforms. `_executable_from_line()` is split out of `_is_resolve_command()` so the "is this Resolve" test and the location lookup agree about where the path ends — including the WMIC quoting fixed in #150.

**Discovery for cold installs** — `platform.discover_scripting_lib()` consults the running process first, then the conventional roots for the platform, for when Resolve is not up:

- Windows: `%PROGRAMFILES%`, `%PROGRAMW6432%`, `%PROGRAMFILES(X86)%`, and existing fixed drives — two fixed paths per drive, no directory walk, A: and B: skipped. This runs on connection attempts and stays cheap.
- macOS: both bundle locations. The default names `/Applications/DaVinci Resolve/DaVinci Resolve.app`; the App Store build installs to `/Applications/DaVinci Resolve.app`, which is the same class of miss as the Windows case. `resolve_runtime.MACOS_RESOLVE_APPS` already recorded both.
- Linux: the `/opt/resolve` layouts already listed in `install.py`'s candidates.

An unknown platform yields no candidates rather than an error.

**Defaults are preserved** — `get_resolve_paths()` calls discovery only when the default is absent *and* no usable env override exists, and keeps the platform default when discovery finds nothing. A failed lookup must not substitute a second guess; the default is what the error message needs to name.

**No more empty env values** — `build_server_env()` omits empty keys rather than writing `""`.

**Honest reporting** — the "optional" line is corrected; a DLL-load failure is diagnosed explicitly, naming the current `RESOLVE_SCRIPT_LIB`, *before* the Python 3.13+ theory; and a failed verification ends in `Setup incomplete — the scripting API did not load.` instead of `Setup complete!`. Configs that were written are still listed, marked as non-functional.

## Tests

New `tests/test_scripting_lib_discovery.py` (17 cases): derivation on Windows and macOS layouts, WMIC-quoted command lines, flags after the executable, the three `None` cases (nothing running, unreadable process list, library not beside the executable), the per-platform candidate lists (pure path construction, so they cover the two platforms I cannot test on), override precedence over discovery, the surviving platform default, the omitted empty key, and `find_resolve_paths()`'s fallback order.

Full offline suite green (2746 passed, 20 skipped) on Linux / Python 3.11. Four pre-existing failures are excluded and are unrelated to this change: `test_live_api` and the two `test_script_plugin` inline-Python cases need a live Resolve, and `test_child_process_text_encoding::test_it_is_searched_first` depends on the interpreter's PATH position.

The new tests were also run on Windows 11 / Python 3.14 — the platform where the bug was found. The first Windows run failed four of them, and both causes were in the tests rather than the code: a `RESOLVE_PATHS` stand-in holding only the host key tripped the eagerly-evaluated `RESOLVE_PATHS["Linux"]` default in `find_resolve_paths()`, and the macOS-layout fixture was built with `os.sep`, so it could not match the slash-separated `RESOLVE_PROCESS_PATTERNS` on a backslash host. Both are fixed; the suite is green on both platforms.

## A separate observation, not addressed here

While running on Windows, 12 pre-existing failures showed up in `tests/test_cdl_and_install_config.py`. They are all the same shape — fixtures built from POSIX literals such as `Path("/tmp/python")`, which `pathlib` renders as `\tmp\python` on Windows — plus one `endswith(".codex/config.toml")`. `docs/contributing.md` already lists Windows testing as help wanted, so this is presumably known; flagging it in case it is not. Happy to open a separate PR making those fixtures separator-agnostic.

### DIFF
--- a/install.py
+++ b/install.py
@@ -284,6 +284,21 @@ def find_resolve_paths():
             lib_path = expanded
             break
 
+    if lib_path is None:
+        # The literal candidates above only cover a default install. An explicit
+        # override wins outright; failing that, ask where Resolve actually is.
+        # Skipping this is what wrote an empty RESOLVE_SCRIPT_LIB into working
+        # configs and left the server importing a DLL that was never there.
+        env_lib = os.environ.get("RESOLVE_SCRIPT_LIB")
+        if env_lib and os.path.isfile(env_lib):
+            lib_path = env_lib
+        else:
+            try:
+                from src.utils.platform import discover_scripting_lib
+                lib_path = discover_scripting_lib()
+            except Exception:
+                lib_path = None
+
     return api_path, lib_path
 
 
@@ -520,7 +535,14 @@ def get_python_base_install(python_path):
 
 
 def build_server_env(python_path, api_path, lib_path, system=SYSTEM, python_home=None):
-    """Build the env block used by all generated stdio MCP configs."""
+    """Build the env block used by all generated stdio MCP configs.
+
+    Keys whose value is empty are omitted rather than written as "". An empty
+    `RESOLVE_SCRIPT_LIB` is worse than an absent one: it reads as configured in
+    the config file, while `DaVinciResolveScript.py` treats it as unset and
+    silently reverts to its own hardcoded install path — so a machine with
+    Resolve elsewhere fails with a DLL-load error that names nothing useful.
+    """
     api_value = str(api_path or "")
     lib_value = str(lib_path or "")
     env = {
@@ -528,6 +550,7 @@ def build_server_env(python_path, api_path, lib_path, system=SYSTEM, python_home
         "RESOLVE_SCRIPT_LIB": lib_value,
         "PYTHONPATH": str(Path(api_value) / "Modules") if api_value else "",
     }
+    env = {key: value for key, value in env.items() if value}
 
     if system == "Windows":
         env["PYTHONHOME"] = str(python_home or get_python_base_install(python_path))
@@ -1921,7 +1944,13 @@ def main():
     if lib_path:
         print(f"  Library:   {green(lib_path)}")
     else:
-        print(f"  Library:   {yellow('Not found')} {dim('(optional — API path is sufficient)')}")
+        # Not optional, whatever this line used to claim: DaVinciResolveScript
+        # is a thin wrapper that loads this binary, so without it every tool
+        # fails at import. Saying "API path is sufficient" here sent people
+        # looking at their Resolve edition and their preferences instead.
+        print(f"  Library:   {red('Not found')} {dim('(required — the scripting API cannot load without it)')}")
+        print(f"  {dim('Set RESOLVE_SCRIPT_LIB to the fusionscript library inside your Resolve install,')}")
+        print(f"  {dim('or start Resolve and re-run setup so its location can be read from the process.')}")
 
     resolve_running = check_resolve_running()
     if resolve_running:
@@ -2122,6 +2151,7 @@ def main():
     if interactive:
         print_step(5, total_steps, "Verification")
 
+    verification_failed = False
     if api_path:
         success, message = verify_resolve_connection(python_path, api_path, lib_path)
         try:
@@ -2171,13 +2201,28 @@ def main():
             else:
                 print(f"  Connected: {green(message)}")
         else:
-            print(f"  Verify:    {yellow(message)}")
+            verification_failed = True
+            print(f"  Verify:    {red(message)}")
+            if "DLL load failed" in message or "cannot open shared object" in message:
+                # This is the shape of a wrong or missing library path, and it
+                # is the one failure the installer can diagnose precisely. Say
+                # so before offering the interpreter theory below — a reader who
+                # is told "try another Python" first will go and do that.
+                print(
+                    f"             The scripting library named by RESOLVE_SCRIPT_LIB did not load. "
+                    f"Current value: {lib_path or dim('(not set)')}"
+                )
+                print(
+                    "             Point RESOLVE_SCRIPT_LIB at the fusionscript library inside your "
+                    "Resolve install, or start Resolve and re-run setup."
+                )
             if py_abi_risk:
                 print(
                     f"             On Python 3.13+ this may be an ABI mismatch with Resolve's "
                     f"scripting library — try Python 3.10-3.12 if it persists."
                 )
     else:
+        verification_failed = True
         print(f"  {yellow('Skipped')} — Resolve API path not detected")
 
     # ══════════════════════════════════════════════════════════════════════
@@ -2185,7 +2230,25 @@ def main():
     # ══════════════════════════════════════════════════════════════════════
 
     print(f"\n  {'═' * 50}")
-    if configured or show_manual:
+    if verification_failed and (configured or show_manual):
+        # Writing the configs is not the job; a working connection is. Reporting
+        # "Setup complete!" over a failed verification is how an install that
+        # never worked gets handed to the user as finished — the error scrolls
+        # past mid-output and the last line says success.
+        print(f"  {yellow(bold('Setup incomplete — the scripting API did not load.'))}")
+        if configured:
+            print(f"  Configured: {', '.join(configured)} {dim('(written, but the server will fail to start)')}")
+        print()
+        print(f"  {bold('Fix the verification error above, then re-run:')}")
+        print(f"    {cyan('python install.py')}")
+        print()
+        print(f"  {dim(f'Server: {server_path}')}")
+        print(f"  {dim(f'Python: {python_path}')}")
+        if api_path:
+            print(f"  {dim(f'API:    {api_path}')}")
+        if lib_path:
+            print(f"  {dim(f'Library: {lib_path}')}")
+    elif configured or show_manual:
         print(f"  {green(bold('Setup complete!'))}")
         if configured:
             print(f"  Configured: {', '.join(configured)}")
--- a/src/utils/platform.py
+++ b/src/utils/platform.py
@@ -64,12 +64,103 @@ def get_resolve_paths():
     env_lib = os.environ.get("RESOLVE_SCRIPT_LIB")
     if env_lib and os.path.isfile(env_lib):
         lib_path = env_lib
+    elif not os.path.isfile(lib_path):
+        # No usable override and nothing at the default: look for the real
+        # install before handing back a path we already know is not there.
+        # The default is kept as the return value when discovery finds nothing,
+        # so the failure message still names the location people expect.
+        lib_path = discover_scripting_lib(platform_name) or lib_path
 
     return {
         "api_path": api_path,
         "lib_path": lib_path,
         "modules_path": modules_path
     }
+
+
+def _windows_lib_candidates():
+    r"""Plausible `fusionscript.dll` locations on this machine.
+
+    `%PROGRAMFILES%` is not a constant — a 64-bit install can sit under
+    `%PROGRAMW6432%` — and Resolve is routinely moved to a second drive
+    because the application and its caches are large. Only fixed drives that
+    exist are probed, two fixed paths each, no directory walk: this runs on
+    every connection attempt and must stay cheap. A: and B: are skipped so a
+    machine with a floppy-mapped letter does not stall on the probe.
+    """
+    relative = os.path.join("Blackmagic Design", "DaVinci Resolve", "fusionscript.dll")
+    candidates = []
+    for variable in ("PROGRAMFILES", "PROGRAMW6432", "PROGRAMFILES(X86)"):
+        base = os.environ.get(variable)
+        if base:
+            candidates.append(os.path.join(base, relative))
+    for letter in "CDEFGHIJKLMNOPQRSTUVWXYZ":
+        drive = f"{letter}:\\"
+        if not os.path.isdir(drive):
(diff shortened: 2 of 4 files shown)


samuelgursky:
Thanks for this — the diagnosis is right, the write-up made it easy to follow, and the chain you traced from `lib_path=None` through `"RESOLVE_SCRIPT_LIB": ""` to a `Setup complete!` over a dead install is exactly the failure worth closing.

I pulled the branch and checked it against a macOS machine with Resolve installed at the default location.

## What I verified

The process-derivation path works live:

```
procs:    /Applications/DaVinci Resolve/DaVinci Resolve.app/Contents/MacOS/Resolve
lib:      .../Contents/Libraries/Fusion/fusionscript.so
discover: .../Contents/Libraries/Fusion/fusionscript.so
```

The `../Libraries/Fusion/` derivation off `Contents/MacOS/Resolve` is correct, and the `_executable_from_line` split preserves `_is_resolve_command` semantics exactly — both the quoted-span branch and the flag-strip loop are unchanged, and `test_headless_runtime.py` stays green.

I also checked the two things most likely to bite:

- **Omitting empty env keys is safe.** `verify_resolve_connection` indexes `env["PYTHONPATH"]`, but only after the `if not api_path: return` above it, so the key is always present at that point. JSON configs replace the whole `davinci-resolve` entry and the Codex path re-renders the env table from `entry["env"]`, so a stale `"RESOLVE_SCRIPT_LIB": ""` from an earlier run is dropped rather than merged forward. Windows `PYTHONHOME` is added after the filter, so it survives.
- **`Setup incomplete` does not misfire when Resolve is merely closed.** `verify_resolve_connection` returns `True` on `IMPORTED_OK`, so a normal offline install still reports complete. That was the regression I most wanted to rule out.

Suite on the branch here: **2767 passed, 1 failed**, and no new pyflakes warnings (13 before, 13 after).

## The one blocker: a test that only passes where Resolve is not installed

`GetResolvePathsDiscoveryTests::test_discovery_fills_in_a_missing_default` fails on this machine:

```
AssertionError: '/Applications/DaVinci Resolve/.../fusionscript.so' != '/var/folders/.../fusionscript.so'
```

It stubs `get_platform` and `discover_scripting_lib`, but not the `os.path.isfile(lib_path)` gate that decides whether discovery runs at all. Where the default really is on disk, the new `elif` never fires and the real default comes back instead of the stub.

This is a test defect, not a code defect — I confirmed the production path returns the discovered lib correctly once the isfile check is stubbed.

**Worth flagging: you will not be able to reproduce this.** Your Windows box has Resolve on `F:` and your CI is Linux, so on both the default is genuinely absent and the branch runs by accident — the test goes green for the wrong reason. Same root cause makes `test_the_platform_default_survives_when_discovery_finds_nothing` pass vacuously there. That is why I ran it here rather than asking you to.

### Patch

Test-only, and it un-vacuums the other two cases in the class at the same time:

```diff
--- a/tests/test_scripting_lib_discovery.py
+++ b/tests/test_scripting_lib_discovery.py
@@ -184,6 +184,26 @@ class CandidateListTests(unittest.TestCase):
             self.assertIsNone(platform_utils.discover_scripting_lib("plan9"))
 
 
+def _default_lib_absent():
+    """Make the platform default missing, whatever this machine has installed.
+
+    `get_resolve_paths()` only reaches discovery when the default is not on
+    disk, so a test that does not force that condition checks nothing on a
+    machine where Resolve *is* at the default location — and on macOS it does
+    not merely pass vacuously, it fails, because the real default comes back
+    instead of the stub. Neither of the machines this change was written on
+    shows it: a Linux box has no `/Applications`, and the Windows box that
+    prompted the fix has Resolve on `F:`, so on both the default is already
+    absent for real and the branch runs by accident.
+    """
+    real_isfile = os.path.isfile
+    return mock.patch.object(
+        platform_utils.os.path,
+        "isfile",
+        lambda path: False if "DaVinci Resolve" in str(path) else real_isfile(path),
+    )
+
+
 class GetResolvePathsDiscoveryTests(unittest.TestCase):
     def test_discovery_fills_in_a_missing_default(self) -> None:
         with tempfile.TemporaryDirectory() as tmp:
@@ -192,7 +212,8 @@ class GetResolvePathsDiscoveryTests(unittest.TestCase):
                 pass
             with mock.patch.object(platform_utils, "get_platform", return_value="darwin"), \
                     mock.patch.object(platform_utils, "discover_scripting_lib", return_value=lib), \
-                    mock.patch.dict(os.environ, _env_without_overrides(), clear=True):
+                    mock.patch.dict(os.environ, _env_without_overrides(), clear=True), \
+                    _default_lib_absent():
                 paths = platform_utils.get_resolve_paths()
             self.assertEqual(paths["lib_path"], lib)
 
@@ -207,7 +228,8 @@ class GetResolvePathsDiscoveryTests(unittest.TestCase):
             env["RESOLVE_SCRIPT_LIB"] = override
             with mock.patch.object(platform_utils, "get_platform", return_value="darwin"), \
                     mock.patch.object(platform_utils, "discover_scripting_lib", return_value=discovered), \
-                    mock.patch.dict(os.environ, env, clear=True):
+                    mock.patch.dict(os.environ, env, clear=True), \
+                    _default_lib_absent():
                 paths = platform_utils.get_resolve_paths()
             self.assertEqual(paths["lib_path"], override)
 
@@ -216,7 +238,8 @@ class GetResolvePathsDiscoveryTests(unittest.TestCase):
         the default is what the error message needs to name."""
         with mock.patch.object(platform_utils, "get_platform", return_value="darwin"), \
                 mock.patch.object(platform_utils, "discover_scripting_lib", return_value=None), \
-                mock.patch.dict(os.environ, _env_without_overrides(), clear=True):
+                mock.patch.dict(os.environ, _env_without_overrides(), clear=True), \
+                _default_lib_absent():
             paths = platform_utils.get_resolve_paths()
         self.assertEqual(
             paths["lib_path"],
```

I checked that the fixed test is load-bearing rather than merely green: with the patch applied I removed the `elif not os.path.isfile(lib_path)` branch from `get_resolve_paths()` and the test failed; restoring it passed. Full offline suite with the patch: **2768 passed, 1 skipped**.

Could you apply it and push to this branch? Branch protection here dismisses the approval if a maintainer pushes, so it merges much more cleanly coming from you.

## Optional notes, none of which need to hold up the merge

1. **`_windows_lib_candidates` docstring overclaims.** It says only fixed drives are probed, but the code runs `os.path.isdir` on C–Z with no drive-type check, so a stale mapped network drive gets probed too. Either check `GetDriveTypeW` or soften the wording. The same docstring says this runs on every connection attempt — `get_resolve_paths()` is actually import-time, so the real cost is one `ps`/`wmic` spawn at server startup, and only in the already-broken case.

2. **Two bare `except Exception` silent fallbacks** — around the `running_resolve_lib` import in `discover_scripting_lib`, and around the `discover_scripting_lib` import in `find_resolve_paths`. This repo has had a run of silent-fallback bugs, so a relative `from .resolve_runtime import running_resolve_lib` plus a narrower except would sit better with the house style.

3. **Two gaps in the honest-reporting change.** The `elif not selected_ids` branch still prints `Environment ready!` over a failed verification, and `main()` still exits 0 — so `npx davinci-resolve-mcp setup` in a script reports success either way. Both are the same class of thing this PR is fixing, if you feel like closing them here.

On the separate `test_cdl_and_install_config.py` observation: yes please, a separate PR making those fixtures separator-agnostic would be welcome.

Once you have pushed the test fix and confirmed the suite on Windows 11 / Python 3.14 — the platform none of us can exercise from here, and the one where your original four failures showed up — this is good to merge.

samuelgursky:
Merged and shipped in [v2.98.4](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.98.4) — on npm as `davinci-resolve-mcp@2.98.4`.

Thank you for this one. The diagnosis was the valuable part: tracing `lib_path=None` → `"RESOLVE_SCRIPT_LIB": ""` → the loader's fallback to the same missing path → `Setup complete!` is a chain where every individual link looks defensible, and the report is what made it obvious that the last link is the one doing the damage.

I took your suggestion and applied the test fix myself rather than making you round-trip for it — you'd already spent enough time on this. Before landing it I mutation-checked it: with the patch applied, removing the `elif not os.path.isfile(lib_path)` branch from `get_resolve_paths()` fails the test, and restoring it passes. So the case is load-bearing now rather than passing for the reason it used to.

I also closed the three optional notes from the review, in `062d41d`:

- **The `Environment ready!` branch and the exit status.** Both were the same bug as `Setup complete!`, one block further down: the no-clients path still printed ready over a failed verification, and `main()` returned `None` on every path, so `npx davinci-resolve-mcp setup` in a script saw a zero either way. Summary line and exit status now agree in all three states.
- **The two bare `except Exception` fallbacks** narrowed to `ImportError`, with relative imports.
- **The `_windows_lib_candidates` docstring** corrected on both counts — no `GetDriveTypeW` check, so a connected network drive is probed too, and `get_resolve_paths()` is import-time rather than per-connection.

`tests/test_scripting_lib_discovery.py` is at 21 cases; the three new ones were each mutation-checked against the pre-fix behaviour. Full offline suite: 2784 passed, 1 skipped.

One consequence worth telling you about, since it validates the change: **the exit-status fix immediately failed the publish workflow.** The CI runner has no Resolve, so verification fails there and setup now correctly exits non-zero — and the smoke-test step only checked that the command didn't crash. I rewrote it to pin the actual invariant instead of swallowing the status: the summary line and the exit status must agree, in whichever direction the environment takes them. `Setup incomplete` + non-zero passes on the runner, `Setup complete!` + zero passes on a machine with Resolve, and a mismatch — an installer reporting incomplete while exiting 0, which is exactly what you reported — now fails the publish.

On your separate observation: yes, please do open that PR for the POSIX-literal fixtures in `tests/test_cdl_and_install_config.py`. Windows testing is genuinely help-wanted here and those 12 failures are a real gap, not a known-and-accepted one.

PR #156, opened 2026-08-21
feat(timeline): ripple_insert, verified-delete gate for move_clips, CDL preflight, Magic Mask HITL

handst97:
# feat(timeline): `ripple_insert`, verified-delete gate for `move_clips`, CDL preflight, Magic Mask HITL

Target: `samuelgursky/davinci-resolve-mcp` `main` (ported onto 2.98.3 / commit `132e134`).
Two commits: the feature commit and a `chore(release): bump version to 2.99.0`
(minor — new MCP action — per `docs/process/release-process.md`; drop the bump
commit if you'd rather batch the release).

## Why

On 2026-08-19 a real session lost 26 clips: `move_clips` was used with a
`record_frame_offset` smaller than the item duration to "open a gap".
`AppendToTimeline` into the occupied span returned items whose ids could not
be read, the code counted those as successful duplicates, and the delete phase
removed the sources. Everything here follows from that incident.

## What

- **`timeline.ripple_insert`** (new action): insert media-pool source ranges at
  a record point and shift all later video/audio items right. There is no
  ripple-insert primitive in the scripting API, and duplicate-then-delete
  corrupts the timeline when the shift is smaller than an item (the null-id
  failure above). The action plans a rebuild instead: capture every tail item's
  pool media + source trim, delete the tail (verified, via
  `_timeline_delete_clips_verified`), re-append it shifted, then place the
  inserts into the opened gap — tail first, so the worst mid-failure state is
  a gap, never lost content. Dry-run by default with straddler / blocker /
  locked-track / subtitle detection; execution is confirm-token gated,
  archived by the destructive hook, holds `edit_page_for_timeline_edits` for
  the whole rebuild, and ends with a full position readback.
  Transform/crop/composite/retime properties are re-applied to shifted items;
  grades, keyframes, transitions, and link state are NOT preserved (the
  pre-mutation archive keeps them) — stated in the plan warnings, the confirm
  preview, and the docs.
- **`move_clips` verified-delete gate**: each duplicate row now carries
  `duplicate_verified` (a live item or a real id was recovered from the
  timeline). Sources are deleted only when every duplicate — primary and
  linked — verified; null-id appends keep the source with an explicit warning.
- **`safe_set_cdl` / `apply_look_to_items` node preflight**: reads
  `item.GetNodeGraph().GetNumNodes()` before `SetCDL` (NodeIndex is 1-based;
  `TimelineItem.GetNumNodes` is deprecated) and returns a structured reason on
  a missing node plus a clip-type-aware diagnosis when `SetCDL` returns False
  (there is no `GetCDL` to read back).
- **Magic Mask HITL**: Magic Mask v2 isolates via operator clicks (manual
  ch. 139; strokes are the legacy v1 interface) and the API cannot place
  clicks, so with none present `CreateMagicMask` can only return False. The
  action now returns `needs_hitl` with the exact Color-page steps; mode
  aliases are normalized (`'Forward'` → `'F'`). The granular
  `ti_create_magic_mask` defaulted to `'Forward'`, which Resolve rejects
  verbatim — fixed the same way.
- **api_truth**: three new entries (AppendToTimeline null-id items,
  CreateMagicMask needs clicks, SetCDL write-only); `api-limitations.md`
  regenerated.

## A conflicting measurement, isolated rather than asserted

Your 2026-08-02 correction to the `AddFusionComp` entry (wired
MediaIn→Blur→MediaOut renders on Studio 19.1.3.7, PSNR evidence) conflicts
with what we measured on Studio 21.0.4 (2026-08-20): a fresh comp wired
MediaIn→Transform(Size 0.45)→MediaOut read back correctly but changed neither
a Color-page gallery still nor a delivered H.264 render; the same held for
both comps of a 2-comp item and for `LoadFusionCompByName` returning a live
comp object. We then isolated the variable with a controlled run (phase 2 of
the new `tests/live_ripple_insert_validation.py`, Studio 21.0.4.5, disposable
project, synthetic media): a wired MediaIn→Blur(XBlurSize 20)→MediaOut comp —
your exact rendering configuration — delivered a render **bit-identical to
the no-comp baseline** (ffmpeg PSNR inf), as did the Transform variant. So
the tool is eliminated; the conflict tracks the Resolve build (19.1.3.7
honoured the wired comp, 21.0.4.5 ignores it). Rather than adding a second
entry that contradicts yours, this PR folds the 21.0.4 measurements into the
existing entry as a dated, version-scoped observation, and keeps the
actionable rule version-independent: prove Fusion/grade claims with a
rendered frame (`gallery_stills grab_and_export`), never with comp readback.
Happy to rework if you'd rather structure the entry differently — and if you
still have a 19.1.3.7 install, re-running phase 2 there would confirm the
regression boundary for a Blackmagic report.

## Validation

- Live, this branch (Studio 21.0.4.5, disposable project + synthetic media,
  2026-08-20, via `tests/live_ripple_insert_validation.py`):
  - `timeline("ripple_insert")` through the real dispatch: dry-run plan exact
    (insert@86448, shift 24, tail 2, no straddlers/blockers), confirm-token
    round-trip, post-insert layout
    `[(86400,48),(86448,24),(86472,48),(86520,48)]`, `readback.missing=[]`,
    ZoomX/Y=0.5 survived the shift with `property_restore_failures=0`.
  - Fusion isolation (phase 2): Blur and Transform comps both rendered
    bit-identical to baseline (PSNR inf) — see the section above.
- Live (Studio 21.0.4, sandbox timelines only, 2026-08-19/20 — performed on a
  2.75.0-based build of these same changes; evidence in the reporter's
  evaluation log):
  - `ripple_insert`: 3×48-frame sandbox, insert 24 frames at relative 48 —
    dry-run plan exact, confirm-token flow correct (pre-mutation archive
    skipped on the token-issuing call, taken on execute), layout after
    `[(86400,48),(86448,24),(86472,48),(86520,48)]`, `readback.missing=[]`;
    ZoomX/Y=0.5 on a shifted item survived, `property_restore_failures=0`.
  - null-id reproduction: `move_clips` with `record_frame_offset:10` on a
    48-frame item → `duplicate_verified=false`, source kept, item count
    unchanged.
  - CDL: dry-run preflight row; `NodeIndex=99` → structured "exceeds node
    count 1"; `NodeIndex=1` → success; `apply_look_to_items` 4/4 with
    per-item preflight rows.
  - Magic Mask: no clicks → `needs_hitl` payload; alias `'Forward'`
    normalized to `'F'`.
- Offline (this branch, Python 3.13):
  - `tests/test_import.py` OK (incl. package-lock sync)
  - `scripts/audit_api_parity.py` PASS; `scripts/gen_api_limitations.py
    --check` OK; `node scripts/agent-rules/generate.mjs --check` OK
  - drift guards (`test_static_undefined_names`, `test_duplicate_definitions`,
    `test_action_list_drift`, `test_panel_docs_drift`, `test_doc_tool_counts`,
    `test_agent_rules_drift`) OK
  - new `tests/test_ripple_insert.py` (10 tests) OK — plan geometry mirrors
    the live run; null-id `duplicate_verified` gate covered
  - full offline discovery: 2707 tests, failures limited to cases that fail
    identically on `origin/main` in the same environment (missing optional
    numpy; `test_source_media_guard` ffmpeg-environment cases)
  - `npm pack --dry-run`, `bin/davinci-resolve-mcp.mjs --version` → 2.99.0
- The offline suite and the live harness are both reusable:
  `tests/test_ripple_insert.py` (offline, mirrors the live geometry) and
  `tests/live_ripple_insert_validation.py` (live; creates and deletes its own
  disposable project, restores the previously open project).


samuelgursky:
Thanks for this — and for leading with the incident rather than the feature. `move_clips` being reachable as a gap-opener, with `AppendToTimeline` handing back null-id items that the code then counted as successful duplicates, is exactly the shape of failure worth a new primitive rather than a patch. The tail-before-inserts ordering, so the worst mid-failure state is a gap and never lost content, is the right call and the reason I'm comfortable with an action this destructive.

I pulled the branch into a worktree and went through it against the offline suite plus some adversarial probing.

## What I verified

**Suite, apples-to-apples.** Both trees in equivalent worktrees, same interpreter (3.14.6), `-p no:randomly`:

```
main  (132e134)   4 failed, 2759 passed, 1 skipped, 537 subtests
pr156 (0cec0da)   4 failed, 2769 passed, 1 skipped, 537 subtests
```

Identical failure set, +10 tests, no regressions. The four are the `test_source_media_guard` cases you flagged, and I can confirm your read: they are a **worktree-path artifact, not an environment quirk on your side**. They pass in the primary checkout and fail in any worktree, on `main` and on this branch alike — the guard resolves scratch against the repo root. Nothing for you to do.

**Generated files and drift guards** all clean on the branch: `audit_api_parity.py` PASS, `gen_api_limitations.py --check` up to date, `agent-rules/generate.mjs --check` in sync, `test_import.py` ok, and the five drift guards green.

**Boundary probes** against your stubs, which the PR's tests don't cover and which I half-expected to break:

- insert at the very start (`record_frame=0`, everything is tail, head empty) → `[(86400,24),(86424,48),(86472,48),(86520,48)]`, `missing=[]`
- insert past the last item (tail empty, delete phase correctly skipped) → `[(86400,48),(86448,48),(86496,48),(86544,24)]`, `missing=[]`

Both correct. The `if tail_items and not _timeline_delete_clips_verified(...)` guard is doing real work in the second case.

**The archive gating is right where it matters.** I checked that the confirm-token round-trip skips the pre-mutation archive on the token-issuing call and takes it on execute — `_action_will_gate_pending_confirm` already covers `ripple_insert` via `_TOKEN_GATED_DESTRUCTIVE_ACTIONS`. That held. But see finding 1.

## Three findings

None of these risk content. All three are verified against the code, not read off it, and the patch below is one diff covering all three — full suite with it applied: **4 failed, 2769 passed**, identical to the branch without it, and the drift guards stay green.

### 1. The default (dry-run) call archives a timeline version when confirm tokens are off

`ripple_insert` is the only destructive action whose **default** call mutates nothing. The F4 pending-confirm skip covers the plan path, but only while the confirm-token preference is ON — it short-circuits on `if not _confirm_token_required(): return False`. With tokens off, `is_destructive()` has no dry-run awareness, so every routine planning call archives:

```
is_destructive(dry_run=True)  = True     <- plan-only call, archived anyway
confirm_token_required=True   -> will_gate=True  => archive on dry-run: False
confirm_token_required=False  -> will_gate=False => archive on dry-run: True
```

Which is the exact waste F4 exists to prevent, on the action's normal read-only path. The patch adds a narrow plan-only filter alongside the existing `NO_ARCHIVE_ON_KEYS` idiom rather than making `is_destructive` globally dry-run-aware, since that would quietly change gating for every other action that happens to take a `dry_run`.

### 2. Asymmetric per-track insert durations leave an unreported gap, and the result still says `success: true`

`shift` is the max across tracks, so a shorter insert on one track leaves a hole — and the readback structurally cannot see it, because `expected` only ever contains the three roles the action placed. Inserting 24 video frames and 48 audio frames at the same point:

```
V1: [(86400,48), (86448,24), (86496,48)]   <- 24 frames of black at 86472
A1: [(86400,48), (86448,48), (86496,48)]
success=True   missing=[]   warnings: (only the grades/keyframes one)
```

Gaps are ordinary ripple-insert semantics — a track that gets no insert at all is *supposed* to open up. The problem is only that it's silent: an unqualified success over a timeline with black in it is the wrong answer, and this is the same class of thing as the null-id gap you're closing. The patch computes `gap_frames_by_track` in the plan phase and surfaces it in both the plan and the execute result, plus a warning naming the shift and the affected tracks.

### 3. Subtitle straddlers pass the feasibility check

The subtitle scan is `item_start >= insert_frame`, so a subtitle that *straddles* the insert point is neither blocked nor shifted — it stays put while the picture under it moves right. Video and audio straddlers already refuse the plan; this is just the subtitle branch not getting the same treatment. One-line `elif` on `GetEnd()`.

### Patch

```diff
--- a/src/utils/destructive_hook.py
+++ b/src/utils/destructive_hook.py
@@ -226,6 +226,31 @@ STRICT_DEFAULT_ACTIONS: frozenset = frozenset({
 })
 
 
+# ── Plan-only (dry-run) actions ─────────────────────────────────────────────
+#
+# Actions whose payload can declare the call non-mutating. `ripple_insert` is
+# the only destructive action that DEFAULTS to a plan-only call, so without
+# this every routine dry-run plan archives a timeline version — the same waste
+# the pending-confirm skip (F4) exists to prevent, except the pending-confirm
+# skip only fires while the confirm-token preference is ON. With it off, the
+# read-only path of the action littered the version chain.
+
+DRY_RUN_DEFAULT_TRUE_ACTIONS: frozenset = frozenset({
+    ("timeline", "ripple_insert"),
+})
+
+
+def _payload_is_plan_only(
+    tool_name: str, action: str, params: Optional[Dict[str, Any]],
+) -> bool:
+    """True iff this call only produces a plan and mutates nothing."""
+    if (tool_name, action) not in DRY_RUN_DEFAULT_TRUE_ACTIONS:
+        return False
+    if not isinstance(params, dict):
+        return True  # dry_run defaults to True for these actions
+    return bool(params.get("dry_run", params.get("dryRun", True)))
+
+
 def _payload_only_touches_no_archive_keys(
     tool_name: str, action: str, params: Optional[Dict[str, Any]],
 ) -> bool:
@@ -250,6 +275,8 @@ def is_destructive(tool_name: str, action: str, params: Optional[Dict[str, Any]]
         return False
     if _payload_only_touches_no_archive_keys(tool_name, action, params):
         return False
+    if _payload_is_plan_only(tool_name, action, params):
+        return False
     return True
--- a/src/server.py
+++ b/src/server.py
@@ -4912,8 +4912,15 @@ def _timeline_ripple_insert_impl(proj, tl, p: Dict[str, Any], *, resolve=None) -
     for track_index in range(1, _timeline_track_count(tl, "subtitle") + 1):
         for item in (tl.GetItemListInTrack("subtitle", track_index) or []):
             item_start = _frame_int(item.GetStart())
+            item_end = _frame_int(item.GetEnd())
+            # A subtitle that STRADDLES the insert point is as much a blocker as
+            # one after it: video/audio straddlers already refuse the plan, and
+            # leaving a straddling subtitle in place silently desyncs it against
+            # the shifted picture.
             if item_start is not None and item_start >= insert_frame:
                 subtitle_blockers += 1
+            elif item_end is not None and item_end > insert_frame:
+                subtitle_blockers += 1
 
     # Capture rebuild info + restorable properties for every tail item BEFORE
     # anything mutates (the live item objects die at delete time).
@@ -4942,10 +4949,31 @@ def _timeline_ripple_insert_impl(proj, tl, p: Dict[str, Any], *, resolve=None) -
         except Exception:
             pass
 
+    # Every track shifts by the SAME amount (the longest inserted run), so any
+    # track whose inserts are shorter than that — or which gets no insert at all
+    # — is left with a gap at the insert point. That is ordinary ripple-insert
+    # semantics, but it has to be reported: the readback below only checks the
+    # positions it placed, so it cannot see the hole, and an unqualified
+    # success over a timeline with black/silence in it is the wrong answer.
+    gap_by_track: Dict[str, int] = {}
+    for track_type in ("video", "audio"):
+        media_type = _timeline_media_type(track_type)
+        for track_index in range(1, _timeline_track_count(tl, track_type) + 1):
+            cursor = cursor_by_track.get((int(media_type), int(track_index)))
+            filled = (cursor - insert_frame) if cursor is not None else 0
+            if filled < shift:
+                gap_by_track[f"{track_type}:{track_index}"] = shift - filled
+
     warnings = [
         "shifted items are re-created from pool media: grades, keyframes, transitions, "
         "Fusion comps, and link state on them are NOT preserved (the pre-mutation archive keeps them)",
     ]
+    if gap_by_track:
+        warnings.append(
+            "every track shifts by the longest inserted run (%d frames); these tracks are "
+            "left with a gap at the insert point: %s"
+            % (shift, ", ".join(f"{track}={frames}f" for track, frames in sorted(gap_by_track.items())))
+        )
     if linked_tail_ids:
         warnings.append(f"{len(linked_tail_ids)} shifted item(s) had linked items; links will not survive the shift")
     plan = {
@@ -4960,6 +4988,7 @@ def _timeline_ripple_insert_impl(proj, tl, p: Dict[str, Any], *, resolve=None) -
         "blockers": blockers,
         "subtitle_items_after_insert_point": subtitle_blockers,
         "locked_tracks_with_tail": locked_tracks,
+        "gap_frames_by_track": gap_by_track,
         "warnings": warnings,
     }
     feasible = not (straddlers or blockers or subtitle_blockers or locked_tracks)
@@ -5119,6 +5148,7 @@ def _timeline_ripple_insert_impl(proj, tl, p: Dict[str, Any], *, resolve=None) -
         "properties_restored_items": restored,
         "property_restore_failures": restore_failures,
         "readback": {"after_counts": after_counts, "missing": missing},
+        "gap_frames_by_track": gap_by_track,
         "warnings": warnings,
     }
     if failures:
```

Three offline tests would keep these load-bearing, in the style of the ones you already wrote: `is_destructive("timeline","ripple_insert",{"dry_run":True})` is False and with `dry_run=False` is True; the asymmetric two-track fixture asserts `gap_frames_by_track == {"video:1": 24}`; a subtitle straddling the insert point makes the plan infeasible. Also worth extending `docs/SKILL.md`'s `ripple_insert` bullet to say the shift is the longest inserted run and shorter tracks are left with a gap — right now it reads as though the inserts fill the opening.

## On the Fusion measurement

Folding it into the existing entry as a dated, version-scoped observation is the right structure — a second contradicting entry would have been worse, and keeping the actionable rule version-independent ("prove it with a rendered frame") is the part that actually protects people. No changes wanted there.

On your open question: **yes, I still have that install** — `/Applications/DaVinci Resolve/DaVinci Resolve.app` reports `19.1.3` (the plist gives three components, so I can't confirm from here that it's the `.7` point build the entry names). Isolating the regression boundary is worth doing and I'd like the Blackmagic report to have it, but it's a live run on a second Resolve install and it isn't a dependency of this PR — I'll do it separately rather than hold the merge on it. If the phase-2 harness needs anything version-specific to run on 19.1.3, say so now and I'll fold it in.

## Version bump

Keep the bump commit — a new MCP action is a minor, and this repo releases per merged PR rather than batching, so `2.99.0` will ship as its own release rather than waiting.

## Next step

Could you apply the patch and push to this branch? Branch protection here dismisses the approval when a maintainer pushes, and `require_last_push_approval` then blocks re-approval, so it merges far more cleanly coming from you. Once it's pushed I'll re-run the suite, do the live validation pass on my side, and merge + release.

samuelgursky:
I ran your phase 2 on Studio 19.1.3.7, as you asked. The result is not the one either of us expected, and it changes what the api_truth edit in this PR should say.

## Your harness reproduces your result on 19.1.3.7

```
Connected to DaVinci Resolve Studio 19.1.3.7
PHASE 1 PASS: layout [(86400,48),(86448,24),(86472,48),(86520,48)], ZoomX/Y=0.5 survived, restore_failures=0
Blur render PSNR vs baseline: inf
Transform render PSNR vs baseline: inf
PHASE 2 RESULT:
  Blur: comp IGNORED at render (PSNR inf >= 40.0)
  Transform: comp IGNORED at render (PSNR inf >= 40.0)
```

Phase 1 passed exactly as you documented it, on a build you have not tested — nice.

Phase 2 says "ignored" on 19.1.3.7 too. Read alone that looks like it *kills* the version hypothesis by showing both builds ignore the comp. It doesn't, because the comp is not actually ignored on this build.

## The same graph, wired without the MCP tool, renders

```
base 2,018,027 bytes -> blurred 727,323 bytes   PSNR 24.38 dB   RENDERED
```

Same build, same media, same render settings, same `MediaIn -> Blur(20) -> MediaOut` topology. That reproduces the 2026-08-02 entry (22.7 dB, file shrank as a blur should). **The original measurement is correct and reproducible; the entry does not need a version-scoped caveat.**

## The variable is our own `fusion_comp` tool

Five isolation runs, each rendering against its own baseline. Not the comp handle — `AddFusionComp`, `GetFusionCompByIndex` and `GetFusionCompByName` handles all render. Not the node rename. Not `add_tool` or `connect` through the tool. It narrowed to one call:

| value written via | result |
| --- | --- |
| `fusion_comp set_input` | **IGNORED** (PSNR inf) |
| `tool.XBlurSize = 20.0` (raw) | RENDERED 24.38 dB |
| `tool.SetInput("XBlurSize", 20)` (raw) | RENDERED 24.38 dB |

`set_input` wrapped the write in `comp.Lock()`/`Unlock()`. A value written under a comp lock is stored in the graph and returned by `GetInput` — and by our own `get_input` — while the render ignores it. Removing the lock: renders at 24.38 dB.

Fixed and shipped in [v2.98.5](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.98.5) — six sites held a lock across a value write. It was on `main` the whole time, unrelated to this PR.

## What this means for the PR

**Please drop the version-scoped observation from the `AddFusionComp` entry.** The conflict does not track the Resolve build. Your 21.0.4 measurements went through `fusion_comp set_input`, which is what produces "ignored" on any build — so they measure this bug, not 21.0.4.

I want to be precise about the one thing I did *not* establish: I only have 19.1.3.7, so I cannot confirm the fix resolves your 21.0.4 reading. It is possible there is an additional version component sitting on top of this. If you can re-run phase 2 against 21.0.4 on v2.98.5, that would settle it — and if it still reads "ignored" there, then there IS a version story and it belongs in the entry, just not the one currently written.

One note on the reasoning rather than the result: phase 2 concludes "the tool is eliminated as the variable," but the isolation ran through the same `fusion_comp` path it was trying to clear, so it could not have detected a fault there. The check that would have caught it is the one your own PR argues for everywhere else — compare against a path that doesn't share the suspect component. Worth adding a raw-API control to phase 2 while you're in there; that single extra render is what turned this around.

Your instinct to isolate rather than assert was right, and the harness was good enough that I could point it at a different build and get a decisive answer in one run. That is why this got found at all.

The [three findings from my earlier review](https://github.com/samuelgursky/davinci-resolve-mcp/pull/156#issuecomment-5374701620) still stand and the patch there still applies. Note that `main` has moved (v2.98.5), so you'll want to rebase — the `set_input` change touches `src/server.py` near nothing you edited, but the version bump in your second commit now collides. Easiest is to drop your `chore(release): bump version to 2.99.0` commit and let me cut the release on merge.

samuelgursky:
Landed in [v2.99.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.99.0) — on npm as `davinci-resolve-mcp@2.99.0`.

I applied your commit directly to `main` (`b031f85`, your authorship preserved) rather than waiting for a rebase, since `main` had moved five releases underneath you and the conflicts were mine to resolve, not yours. Closing this as merged-by-other-means rather than unmerged.

## What landed as you wrote it

`ripple_insert`, the `duplicate_verified` gate on `move_clips`, the CDL node preflight, the Magic Mask HITL payload, and the mode-alias normalization — all unchanged. The tail-before-inserts ordering is the part I'd single out: getting the worst mid-failure state to be a gap rather than lost content is what made an action this destructive something I was willing to ship.

Live-validated here on Studio 19.1.3.7 — a build you hadn't tested — and your phase 1 passed exactly as documented: plan exact, confirm-token round-trip, layout `[(86400,48),(86448,24),(86472,48),(86520,48)]`, `readback.missing=[]`, ZoomX/Y=0.5 surviving with `property_restore_failures=0`.

## The three review findings, fixed on merge

You didn't get a chance to apply these, so I did — [`f638b0f`](https://github.com/samuelgursky/davinci-resolve-mcp/commit/f638b0f), each with a test that fails without the fix. The dry-run archive, the unreported per-track gap, and the subtitle straddler, as described in the review above.

## The Fusion measurement — you were right to isolate it, and the isolation pointed the wrong way

You asked me to re-run phase 2 on 19.1.3.7. I did, and it reproduced your result: `PSNR inf`, comp ignored. That looked like it killed the version hypothesis by showing both builds ignore the comp.

It didn't, because the comp is *not* ignored on that build. Wiring the same graph without going through this server's `fusion_comp` renders at 24.38 dB. The variable was our own tool: `set_input` wrapped the write in a `Comp.Lock()`, and a value written under a comp lock reads back correctly from every API — including our own `get_input` — while the render ignores it. Fixed in v2.98.5; the mechanism took three more releases to pin down properly (v2.98.8).

**Your phase-2 harness, unchanged, now reports `RENDERED` — Blur 24.38 dB, Transform 7.95 dB.** So your 21.0.4 measurements were real, and they were measuring our bug rather than a Resolve regression. I resolved the `api_truth` conflict against your version-scoped framing for that reason — but kept your 21.0.4 observation in the entry, because the lock bug has only ever been isolated on 19.1.3.7 and your report is the only evidence it isn't build-specific.

One thing worth taking from it: phase 2 concluded "the tool is eliminated as the variable" while running entirely through the tool it was trying to clear. I made the same class of mistake twice afterward — twice I published a wrong conclusion because a test I'd written couldn't fail. Adding a raw-API control to that harness is the cheap fix, and it's what turned this around.

## Nothing left needing your environment

The one open question that genuinely needed 21.0.4 — whether the comp-lock fix resolves what you saw there — would be nice to have, but it isn't blocking anything and I'm not holding a PR open for it. If you do re-run phase 2 on 21.0.4 against 2.99.0 and it still reports IGNORED, that's a real finding and worth a new issue.

Thanks for this one. The incident write-up is what made the case for a new primitive instead of a patch, and the harness was good enough that pointing it at a different build gave a decisive answer in a single run.

Issue #124, opened 2026-08-06
SetClipColor refuses marker-palette names with a bare False — Rose/Fuchsia silently fail while Pink/Tan succeed, and nothing names the valid set


chenyuxiaojin:
## What happened (measured 2026-06-13, Resolve Studio, macOS)

An agent-driven suspect-marking workflow colored timeline clips via `SetClipColor`. Calls with `Rose`, `Fuchsia`, and `Apple` returned `False` — no exception, no color change, nothing else wrong with the items. `Pink`, `Orange`, `Green`, `Yellow`, and `Tan` succeeded **on the same items in the same session**, so the refusal is the name, not item state, lock, or page.

## Why an agent lands on exactly the wrong names

The bundled scripting reference documents `SetClipColor(colorName) --> Bool` with `colorName` as a bare string — no enumerated values anywhere. The same reference *does* export a discoverable color vocabulary: the marker constants, including `resolve.MARKER_ROSE` and `resolve.MARKER_FUCHSIA`. So the only color names an agent can find in the API surface are from the marker palette — and those are precisely the ones `SetClipColor` refuses. A model's prior knowledge of Resolve color names points the same wrong way.

## Scope of the measurement

Five names verified accepted (`Pink`, `Orange`, `Green`, `Yellow`, `Tan`), three verified refused (`Rose`, `Fuchsia`, `Apple`). We did **not** enumerate the full accepted set. Inference, not measured: the accepted set looks like the Edit-page clip-color palette, the refused names come from the marker palette — but the boundary deserves a live enumeration rather than my guess.

## Where v2.80.1 stands (code inspection)

The call sites already treat a `False` honestly — `"SetClipColor returned false"` lands in warnings or an error — but the caller can't tell *invalid name* apart from any other refusal, and no remediation names a single valid value. One batch path records the bool per item and moves on.

## Suggestions (any subset)

- An `api_truth` entry: `SetClipColor` refuses undocumented names with a bare `False`; the marker constants are a decoy vocabulary.
- Enumerate the accepted palette live once, pin it in a constant, and validate (or at least name the measured-valid set in the refusal remediation).
- The same check may apply to `MediaPoolItem.SetClipColor` — same signature, same undocumented value space; we only measured the timeline-item path.

samuelgursky:
[...]

PR #157, opened 2026-08-22
fix(timeline): bulk_set_item_properties cannot set a clip colour on its own

matoberuc-afk:
# fix(timeline): `bulk_set_item_properties` cannot set a clip colour on its own

## The bug

`timeline(action="bulk_set_item_properties")` documents `clip_color` and `enabled`
as per-op keys, and there is code to apply both. That code is unreachable for the
op shape that needs it most.

`src/server.py::_timeline_bulk_set_item_properties` builds the property payload with
`_merge_property_groups(op)`, which merges only `properties`, `transform`, `crop`,
`composite`, `audio` and the duplicate-keyframe keys. `clip_color` and `enabled` are
not `SetProperty` keys, so they never land in that dict. The handler then does:

```python
properties = _merge_property_groups(op)
if not properties:
    results.append({"index": index, "success": False,
                    "error": "op requires properties, transform, crop, composite, "
                             "audio, or direct property keys"})
    continue
```

so a colour-only op returns early and the `if "clip_color" in op:` branch ~25 lines
below never runs. The only way to reach it today is to send a transform you do not
want alongside the colour.

This is not a corner case. Colour triage — paint a selection Apricot / Purple /
Chocolate in one round trip — is the main reason to call a *bulk* property setter,
and those ops carry no transform payload at all.

Two further problems were found in the same function while fixing it:

1. **A colour-only op would have reported success unconditionally.** Per-op success
   was `all(row.get("success") for row in item_result["properties"].values())`, and
   `all([])` is `True`. With no property rows, the op passed no matter what
   `SetClipColor` returned.

2. **The bulk path trusted the bare bool.** It called `item.SetClipColor(...)`
   directly, bypassing `_set_clip_color_checked`, the helper added for
   [#124](https://github.com/samuelgursky/davinci-resolve-mcp/issues/124) that the
   single-item path uses. That helper exists because the bool lies twice: a name
   outside the 16-name Edit-page palette is refused with a bare `False` and no
   reason, and a generator or title item takes the call, returns `True`, and drops
   the colour.

## Repro

```python
timeline(action="bulk_set_item_properties", params={
  "ops": [{"timeline_item_id": "<id>", "clip_color": "Apricot"}]
})
```

Before:

```json
{"success": false, "op_count": 1,
 "results": [{"index": 0, "success": false,
   "error": "op requires properties, transform, crop, composite, audio, or direct property keys"}]}
```

After:

```json
{"success": true, "op_count": 1,
 "results": [{"index": 0, "timeline_item_id": "<id>", "name": "<name>",
              "properties": {}, "clip_color": true, "success": true}]}
```

## The fix

`src/server.py`:

- Accept an op that carries `clip_color` and/or `enabled` with no other payload.
  Only a genuinely empty op is rejected, and the error message now names
  `clip_color` and `enabled` among the accepted keys.
- `dry_run` reports `would_set_clip_color` and `would_set_enabled` alongside the
  existing `would_set`.
- Per-op `success` is computed from every branch that ran, not only from the
  property rows, so `all([])` can no longer pass an op.
- The colour goes through `_set_clip_color_checked(item, colour, kind="timeline item")`.
  On failure the op fails and the diagnosis is attached as
  `results[].clip_color_detail`. On success the value stays the plain `True` that
  existing callers already read, so the response shape is unchanged for anything
  that works today.

Docs: the `timeline` tool description and the `action_help` entry now state that
`clip_color`/`enabled` may be the only key in an op, and the `action_help` example
was changed from a `properties: {"ClipColor": ...}` shape to the real triage shape.
`docs/SKILL.md` matches.

## Tests

Nine tests added to `tests/test_clip_colors.py`
(`BulkSetItemPropertiesClipColorTest`), all failing before the change and passing
after:

| test | what it pins |
| --- | --- |
| `test_clip_color_alone_is_applied` | the reported bug |
| `test_enabled_alone_is_applied` | the same bug on the other unreachable branch |
| `test_a_triage_batch_of_colours_all_land` | Apricot / Purple / Chocolate, the real workflow |
| `test_a_refused_colour_fails_the_op_instead_of_passing_empty` | the `all([])` hole |
| `test_a_colour_that_does_not_persist_is_reported_as_failure` | the generator/title lie from #124 |
| `test_dry_run_reports_the_colour_it_would_set` | dry_run coverage of the new keys |
| `test_properties_and_colour_together_still_work` | the already-reachable path is unchanged |
| `test_a_genuinely_empty_op_is_still_rejected` | the guard is loosened, not removed |

Before the fix: `FAILED (failures=7, errors=2)`.
After: `Ran 18 tests ... OK`.

Full offline suite: 2746 tests, 0 failures, 18 skipped.


samuelgursky:
Landed in [v2.99.1](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.99.1) — on npm as `davinci-resolve-mcp@2.99.1`.

Your diagnosis was exactly right, and I verified each of the three claims against the code before touching anything: `_merge_property_groups` really does drop `clip_color`/`enabled`, the `if not properties` guard really does make the branch below unreachable, `all([])` really is `True`, and the bulk path really was bypassing `_set_clip_color_checked` while the single-item path used it.

The point that made this worth prioritising is the one you led with: colour triage is the *reason* to call a bulk setter, and it was the single op shape that could not work. A bug that only affects the primary use case is worse than its line count suggests.

## Verified before merge

Each of your three fixes is load-bearing — I reverted them one at a time and your tests caught all three:

- restore the `if not properties` guard → 8 failures
- restore the `all([])` success vote → 2 failures
- bypass `_set_clip_color_checked` → 1 failure

Live on Studio 19.1.3.7: three colour-only ops in one call, readback `['Apricot', 'Chocolate', 'Purple']` on the real timeline items, and `Cocoa` correctly returning `success: false` rather than passing on the empty vote.

## One thing to fix for next time

**This PR carries 12 commits, and 11 of them are your machine's local state** — `auto-backup: 2026-08-11 21:29:48 (MacBook-Pro)`, a settings snapshot, a workspace file, and two personal working notes (`mattframes-overnight-context-tuneup.md`, `mattframes-palette-consolidation.md`), plus `.claude/settings.json` and a hookify state file. That's 246 of the 438 added lines.

Merging this through GitHub would have committed all of it to the repo. I landed only your fix commit (`061d6746`) by cherry-picking it, so none of that came along — but it's why I closed this rather than pressing Merge, and it's worth sorting out because it will happen again on your next PR.

The usual cause is a branch cut from a working tree that auto-commits. For next time: branch from a clean `origin/main`, put the fix in one commit, and keep the machine-state branch separate. If you'd like those paths ignored repo-side I'm happy to take a `.gitignore` PR for `.claude/settings.json` and friends — that's a genuinely useful contribution on its own.

None of that reflects on the fix, which was well-argued, correctly diagnosed, and came with tests that actually fail without it. Thanks for it.

PR #161, opened 2026-08-26
Report Whisper timeout as a failed batch clip

### LINKED ISSUE #160: Long-form Whisper batch transcription can time out but still report clip/job as succeeded
I’m reporting a reproducible long-form transcription problem in `davinci-resolve-mcp` 2.103.1 when using local `openai-whisper` through Codex CLI on Windows 11 with DaVinci Resolve Free 21.0.4.5.

I tested an approximately 7-minute MP4 using a durable `media_analysis` batch job with `whisper_cli`, the locally cached `small.en` model, visuals disabled, and no Resolve metadata, markers, captions, indexing, or source-media changes. The transcription configuration was stored correctly, including `transcription.allow_model_download: true`.

When `run_batch_job_slice(max_clips=1)` was executed, Codex received no progress response and the MCP tool call timed out after 300 seconds. The durable job continued and later reported `status: completed`, `phase: complete`, `succeeded_clips: 1`, `failed_clips: 0`, and `last_error: null`.

However, reading the persisted analysis report showed that transcription had actually failed: no transcript was present, segment count was 0, and the transcription result was `wall_clock_timeout` with the message `call exceeded 90s wall-clock cap` at approximately 90,020 ms.

The main correctness issue is that an explicitly requested transcription can fail while the batch job is still reported as successfully completed. This could cause downstream automation to treat failed transcription as valid analysis.

There is also a practical long-form limitation. The default `standard` analysis preset allows only 90 seconds per call, which is too short for local CPU Whisper transcription of normal 7–10+ minute videos. Durable batching currently slices by clip count rather than by time ranges within a long clip, so a single 10-, 30-, or 60-minute recording still becomes one long transcription operation. During that operation Codex receives no progress/heartbeat and eventually times out client-side.

Expected behavior would be for transcription timeout/failure to mark the clip or job as failed or partial-success, surface the error through `batch_job_status`, and avoid returning `succeeded_clips: 1` with `last_error: null`. For long-form use, it would also help if transcription jobs returned control immediately and could be polled for progress, or if long clips were divided into resumable transcription chunks.

This appears related to the long-operation problem previously reported in #88, but the behavior above is still reproducible in v2.103.1.

Steve0x2a:
A requested Whisper transcription that hits the wall-clock cap already returns `success: False` from `_transcribe`. `execute_plan_async` then overwrote that by hard-coding `clip_result["success"] = True` after writing the analysis JSON, so `run_batch_job_slice` counted the clip as succeeded and closed the job as `completed` with `last_error` unset.

Vision already had `_annotate_clip_vision_failure`. Transcription now follows the same shape: a failed transcript marks the clip failed, the batch job records `failed_clips` and `last_error`, and the job status is no longer `completed`.

This does not raise the 90s cap, add chunking, or change heartbeat behavior.

Reported by @techsolvehq-source.

Fixes #160

### DIFF
--- a/src/utils/media_analysis.py
+++ b/src/utils/media_analysis.py
@@ -200,6 +200,34 @@ def _annotate_clip_vision_failure(clip_result: Dict[str, Any], vision: Any) -> N
         })
 
 
+def _annotate_clip_transcript_failure(clip_result: Dict[str, Any], transcript: Any) -> None:
+    """Mark a clip failed when requested transcription did not complete.
+
+    execute_plan_async writes the analysis JSON and then hard-sets success True.
+    Vision already overwrites that via _annotate_clip_vision_failure; a
+    transcription timeout (or any other transcript.success False) had no
+    equivalent, so batch jobs counted an empty transcript as succeeded.
+    """
+    status = ""
+    reason = ""
+    if isinstance(transcript, dict):
+        status = str(transcript.get("status") or "").strip()
+        reason = str(transcript.get("reason") or transcript.get("error") or "").strip()
+    if status == "wall_clock_timeout":
+        message = "Transcription timed out before a transcript was produced."
+        if reason:
+            message = f"{message} {reason}"
+    elif reason:
+        message = f"Transcription was requested but did not complete: {reason}"
+    else:
+        message = "Transcription was requested but did not complete."
+    clip_result.update({
+        "success": False,
+        "error": message,
+        "transcription": transcript,
+    })
+
+
 def _annotate_partial_success(manifest: Dict[str, Any]) -> None:
     """D3 — Mark batch manifests with explicit completed/failed clip-id lists.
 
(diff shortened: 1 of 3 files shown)


samuelgursky:
Merged and shipped in [v2.103.3](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.103.3). Thanks — the diagnosis was right and the shape you chose (mirror `_annotate_clip_vision_failure`) was the right one.

One correction landed on top in fc5a5b0, so you know what changed and why. The gate keyed on `success` alone:

```python
and not transcript.get("success")
```

Transcription is enabled by default (`DEFAULT_TRANSCRIPTION_ENABLED = True`) and `allow_model_download` defaults to `False`, so on a stock install `_transcribe` returns `{"success": False, "status": "skipped"}` — either because no Whisper backend resolves, or because the user hasn't opted into model downloads. I reproduced it against your branch: a one-clip batch on a machine with no backend went from `completed` / 1 succeeded on main to `completed_with_errors` / 1 failed, so every clip of every batch would have failed by default.

The asymmetry that makes the vision pattern not quite transferable: vision defaults to *disabled*, so it can treat every non-success as a failure. Transcription is on by default, so it has to separate "the backend never ran" from "the backend ran and failed".

The fix keeps your annotation and adds `transcription_attempt_failed`, which screens out `skipped` / `disabled` / `not_implemented` and keeps timeouts, caps refusals, and backend errors in the failure class. Your timeout test passes unchanged.

Worth noting the full suite was green on your branch — nothing covered the no-backend path, which is why this got through review rather than testing. There are regression tests for it now.

PR #162, opened 2026-08-28
Improve read/write symmetry audit coverage

diesdaas:
## Summary

- statically resolve `_unknown(...)` action groups expressed through literal, named, starred, and concatenated module-level constants
- handle unresolved dynamic groups conservatively and distinguish occurrence counts from distinct gap names
- recognize known readback aliases and pin the generated symmetry report with regression coverage

## Verification

- `python3 -m unittest tests.test_readwrite_audit -v` — 17 passed
- `python3 scripts/gen_api_limitations.py --check`
- `git diff --check upstream/main...HEAD`

## Broader-suite note

Full Python discovery was also attempted locally. Five unchanged offline DRT cases could not run because `jszip` was unavailable in the local Node dependency environment; no failure involved the files changed in this PR.


samuelgursky:
Merged and shipped in [v2.103.3](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.103.3). The regex → AST rewrite is a real upgrade — I verified every alias it claims (`get_cache_enabled`, `get_caps`, `mcp_update_status`, `get_track_enabled`, `get_track_locked` all exist in `src/server.py`), and the AST pass surfaced a genuine gap the regex was missing: `set_clip_marks` has no `get_clip_marks`. The occurrences relabel is also a correction rather than a change — the old `audit()` was already counting occurrences without dedup, so the old wording was simply wrong.

One follow-up in 5479b88. `test_reference_doc_matches_current_audit` pins the report to `src/server.py`, which is the right guard, but nothing regenerated the report — `main()` only printed to stdout. So the next person to add a `set_`/`add_` action (frequent here) would have hit an `assertEqual` diff of two large documents with no stated way to clear it.

The script now follows the `gen_api_limitations.py` convention: writing is the default, `--check` gates it, `--stdout` preserves the old behavior. `--check` is in the release checklist, and both the script and the test failure name the regeneration command.

Thanks — the conservative handling of unresolved dynamic groups was the right call, and the tests around it made this easy to review.

PR #165, opened 2026-08-29
Add bug report: timeline_frame capture keyword argument error

Douglas4000:
Bug report for `timeline_frame` capture(quality="frame") — fails with
`_BoundMethod.__call__() got an unexpected keyword argument 'isInteractiveMode'`
on DaVinci Resolve 21.0.4.5 free (Windows, via in-app bridge).

Full details in `bug-reports/2026-08-29-thumbnail-frame-capture.md`.

### DIFF
--- a/bug-reports/2026-08-29-thumbnail-frame-capture.md
+++ b/bug-reports/2026-08-29-thumbnail-frame-capture.md
@@ -1,0 +1,24 @@
+## Bug: timeline_frame capture(quality="frame") fails with unexpected keyword argument
+
+**Environment**
+- DaVinci Resolve 21.0.4.5 (free edition, via in-app bridge)
+- Windows 11
+- Server: davinci-resolve-mcp compound server (installed via install.py)
+
+**Steps to reproduce**
+1. Set current timeline
+2. Call `timeline_frame(action="capture", params={"frame": <any frame>, "quality": "frame"})`
+
+**Expected**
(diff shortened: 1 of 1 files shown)


samuelgursky:
Thank you — this report was precise and the diagnosis in your last paragraph was exactly right. The bug is real, it was ours, and it's fixed in [v2.104.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.104.2).

The cause: the in-app bridge proxies Resolve method calls positionally (`_BoundMethod.__call__(*args)`), and the single-frame render path was the one call site in the codebase that used a keyword — `StartRendering([job], isInteractiveMode=False)`. Native external scripting accepts it; the bridge did not, which is why this only bit free-edition users. The call is positional now, the bridge proxy raises a TypeError that names the positional-only rule (instead of the bare `_BoundMethod.__call__()` message you hit), and a regression test greps the server for kwarg calls so a new one can't land.

I'm closing this PR rather than merging it only because the repo doesn't keep a `bug-reports/` directory — GitHub issues are the channel, and this report would have been a model issue. The content of your report is what drove the fix, and you're credited in the changelog and release notes.

Your aside about `quality="thumbnail"` returning black for Fusion/Text+ clips is a known Resolve limitation (per-clip thumbnail API renders source, not output) — it's documented in `docs/reference/api-limitations.md`.

PR #166, opened 2026-08-30
fix(resolve): checked returns on SetCurrentTimeline and 28 other discarded mutators

matoberuc-afk:
Resolve's API reports a refused write by returning False, and a set of call sites dropped that return. Some sat inside a bare `try/except: pass`, so a failed write looked exactly like a successful one and the server carried on.

The one that bites hardest is `SetCurrentTimeline`. When the switch fails, the next edit lands on whatever timeline Resolve still had current, so the work goes somewhere the caller never asked for and nothing says so.

**What is here**

- `src/utils/resolve_writes.py`: a small checked helper, plus `_set_current_timeline_checked`, which returns the error envelope instead of continuing.
- The affected call sites routed through it, in `src/server.py`, `src/granular/`, and the utils probes.
- One deliberate exception, commented in place: the restore inside a `finally` after a frame capture logs a warning rather than returning, because returning there would replace the caller's real result (or its real exception) with a restore failure.
- The other 28 discarded mutator returns checked the same way. The ones that stay discarded are best-effort cleanups in `finally` blocks, or calls whose failure is already visible in the next read, and each says so in a comment.

**Tests**

- `tests/test_current_timeline_switch.py` (7 tests): the switch failing, succeeding, raising, and the finally-restore path.
- `tests/test_discarded_resolve_returns.py` (4 tests): the checked paths report failure, and the deliberate set is pinned so a later refactor cannot quietly widen it.

Full suite on this branch: 3043 tests, and the failure set is byte-identical to `main` without these commits (4 failures, 5 errors, 18 skipped, all pre-existing in my environment). No new failures, none fixed.

Branched off current `main`. No changelog entry, since that file is release-managed here; happy to add one in whatever form you prefer.

### DIFF
--- a/src/granular/media_pool.py
+++ b/src/granular/media_pool.py
@@ -277,11 +277,18 @@ def setup_multicam_timeline(
     if not switched:
         return {"error": describe_switch_failure(switch_detail, "stacking the angles"),
                 "switch": switch_detail}
+    start_timecode_set = None
     if plan.get("start_timecode"):
+        # A refused start timecode leaves the timeline at 01:00:00:00 while every
+        # recordFrame below was computed against the requested one, so the whole
+        # stack lands at the wrong absolute time. Reported, not swallowed.
         try:
-            timeline.SetStartTimecode(plan["start_timecode"])
-        except Exception:
-            pass
+            start_timecode_set = bool(timeline.SetStartTimecode(plan["start_timecode"]))
+        except Exception as exc:
+            return {"error": f"SetStartTimecode('{plan['start_timecode']}') raised: {exc}"}
+        if not start_timecode_set:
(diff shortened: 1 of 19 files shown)


samuelgursky:
Merged and shipped in [v2.104.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.104.2). This is model work: the readback-confirmed switch (not trusting the bare bool, with a stated fallback when `GetCurrentTimeline` can't answer), the deliberate finally-restore exception with its reasoning in place, and the pinned discarded-set test that stops a refactor from quietly widening it — each of those choices is one this repo would have made, and the PR body explained them so review took minutes. Full suite verified here: 3043 on your branch with a failure set byte-identical to main's known environment artifacts, exactly as you reported.

The changelog entry was added on the release side, matching your read that the file is release-managed — no action needed.

PR #170, opened 2026-08-30
fix: avoid fusionscript teardown in doctor and installer probes

FerroQuant:
## Summary

Avoid CPython/Fusion teardown crashes in the short-lived installer and doctor probes.

The probes now follow the repository's established bridge-first connector pattern: ask `connect_resolve(None)` before importing Blackmagic's native `DaVinciResolveScript` module. If native scripting is required, the child flushes its result and exits with `os._exit(0)` after the import attempt so `fusionscript` background teardown cannot race interpreter finalization.

This keeps the existing native fallback and Windows access-violation diagnostics intact while avoiding the native module entirely when the persistent bridge can answer.

## Why

A short-lived Python probe could successfully talk to Resolve and then crash during interpreter shutdown because `fusionscript` leaves background RemoteApp state alive. The failure happens after the useful probe work, so it can look like an installer/doctor crash rather than a connection failure.

This is the same upstream failure class already accepted in PR #108: that merged fix documents `fusionscript.so`'s `RemoteApp` thread continuing to dispatch while CPython exits, with `Fusion::RemoteApp::FindLocalObject` on the crashing thread, and uses a flushed `os._exit()` guard for spawned scripts. This patch extends that already-accepted lifetime rule to the remaining short-lived doctor/install probes.

The bridge-first ordering is also consistent with issue #109, where the maintainer documented that bridge-capable connectors should not require `DaVinciResolveScript` before trying `connect_resolve(None)`.

## Changes

- `scripts/doctor.py`
  - try the persistent bridge before native Fusion import
  - report the native import as skipped when the bridge answers
  - hard-exit only after a native import attempt
- `install.py`
  - use the same bridge-first probe ordering
  - hard-exit after native import attempts
  - preserve current Windows access-violation diagnostics
- add focused lifecycle regression coverage for both probes

## Validation

- 89 focused tests: PASS
- `tests/test_import.py`: PASS
- `scripts/audit_api_parity.py`: PASS
- `scripts/gen_api_limitations.py --check`: PASS
- `node scripts/agent-rules/generate.mjs --check`: PASS
- static/doc regression group: 14 tests PASS, 1 skip
- `node bin/davinci-resolve-mcp.mjs --help`: PASS
- `node bin/davinci-resolve-mcp.mjs --version`: `2.104.1`
- `npm pack --dry-run`: PASS
- `git show --check 9c50a1f`: PASS
- live doctor probe against DaVinci Resolve Studio 21.0.4.5: PASS via bridge, native Fusion module not loaded
- live installer connection probe: PASS
- Resolve and bridge PIDs remained unchanged during validation
- no new macOS Python crash report was produced

### Full-suite baseline comparison

The full local suite is not green in this checkout environment for reasons unrelated to this patch. To distinguish patch regressions from baseline failures, the same command/interpreter was run against both untouched upstream `v2.104.1` (`190fe2c`) and this candidate:

- untouched upstream: 1842 tests; 1 failure, 147 errors, 3 skipped
- candidate: 1844 tests; 1 failure, 147 errors, 3 skipped
- all 134 reported ERROR/FAIL test names are identical
- candidate-only failures: 0

The candidate adds two lifecycle tests, both passing. Dominant baseline errors are missing local `mcp` dependency and missing `resolve-advanced/server/libs.mjs` in that test environment, plus one existing offline-guard failure.

## Release lineage

`v2.104.1` dereferences to `190fe2c62b881870a7eb6add90c660f4bf811a25`; this change is exactly one commit on top of that release as `9c50a1f03b66dce36650ebffdda70eafd6981304`.

### DIFF
--- a/install.py
+++ b/install.py
@@ -1329,37 +1329,57 @@ def access_violation_message(returncode, version=None):
 
 
 def verify_resolve_connection(python_path, api_path, lib_path):
-    """Try to import DaVinciResolveScript and connect."""
+    """Probe Resolve without exposing short-lived children to Fusion teardown.
+
+    The persistent bridge is tried before Blackmagic's native scripting module.
+    If direct scripting is genuinely required, the disposable child flushes its
+    result and hard-exits after any native import attempt so fusionscript's
+    background RemoteApp thread cannot race CPython finalization.
+    """
     if not api_path:
         return False, "Resolve API path not found"
 
     env = {**os.environ, **build_server_env(python_path, api_path, lib_path)}
     modules_path = env["PYTHONPATH"]
     repo_root = str(Path(__file__).resolve().parent)
-    # Route through connect_resolve so Network mode (RESOLVE_SCRIPT_HOST, propagated
-    # into env by build_server_env) uses the explicit IP-targeted overload. Fall
-    # back to Local-mode discovery if the helper cannot be imported.
     test_script = textwrap.dedent(f"""\
+        import os
         import sys
         sys.path.insert(0, {modules_path!r})
         sys.path.insert(0, {repo_root!r})
+        native_import_attempted = False
         try:
-            import DaVinciResolveScript as dvr
             try:
                 from src.utils.resolve_connection import connect_resolve
             except Exception:
-                connect_resolve = lambda mod: mod.scriptapp('Resolve')
-            resolve = connect_resolve(dvr)
+                connect_resolve = None
+
+            resolve = None
+            if connect_resolve is not None:
+                resolve = connect_resolve(None)
+
+            if resolve is None:
+                native_import_attempted = True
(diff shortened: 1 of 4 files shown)


samuelgursky:
Merged and shipped in [v2.104.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.104.2). The lineage discipline in this PR — extending #108's accepted teardown rule and #109's bridge-first ordering rather than inventing a third pattern, with the baseline-comparison table separating patch regressions from environment noise — made it easy to trust. Suite verified here: 3034 with only this repo's known worktree artifacts.

One note for the record: the publish workflow's "Run setup dry run" step exercises `install.py` on a Resolve-less CI runner, and it stayed green through this release — so the probe-child `os._exit(0)` changes are confirmed compatible with that gate.

PR #173, opened 2026-08-31
Make agent assets portable across Codex and Claude

jin386:
## Summary
- make `.agents/skills` the canonical skill source and retain thin Claude compatibility adapters
- centralize reviewer roles and hook policy under `.agents`, with native Codex agent definitions and project hook wiring
- index canonical skills through the MCP knowledge tool and update generated rules/docs for the portable layout
- add cross-host regression tests for source-media protection, frame evidence, and Codex `apply_patch` events

## Validation
- `node scripts/agent-rules/generate.mjs --check`
- `npm run smoke`
- portability-focused Python suite: 65 passed, 220 subtests passed
- full Python suite after rebasing to v2.132.1: 3083 passed, 27 skipped, 751 subtests passed; 5 failures in `tests/test_media_analysis.py` reproduce unchanged on `origin/main`
- live Codex hook-loading smoke test with `--dangerously-bypass-hook-trust` and a read-only shell call


samuelgursky (changes requested):
Thanks — this is genuinely careful work: the canonical `.agents` copies are body-identical to the originals, the hook shims run standalone, `generate.mjs --check` passes, and the portability tests are a nice touch. But I can't take it as-is, because it degrades the Claude Code experience this repo is used with daily:

1. **Skill trigger descriptions were rewritten.** Every `.claude/skills/*/SKILL.md` adapter got a shorter description that drops the "Apply when…" phrasing. Descriptions are what routes a session into a skill — the adapters must keep the canonical descriptions verbatim, and honestly I'd rather the adapters carry the full canonical body (generated, drift-guarded) than a "go read another file" pointer that costs a round-trip per invocation.
2. **`.claude/agents/*` lost `model: opus`.** That pin is intentional for the vision-heavy reviewers. Restore it in the adapters (the host-neutral role files can't carry it).
3. **`CLAUDE.md` → `@AGENTS.md`** inlines all of AGENTS.md into every session. CLAUDE.md is intentionally short; please revert that commit.
4. **`source_media_guard`: the new `agent-` scratch prefix is too broad** — `/Volumes/CARD/agent-XYZ/clip.mov` would be exempt from the deny. Drop `agent-`; `codex-` alone covers your case.
5. The genericized references ("if the host provides craft guidance…") removed named routing to global skills I actually use (`colorist`, `editor`, `online-editor`, …). Keep the names, gate them with "where available" if you like.
6. Please add a parity guard so a canonical skill without a `.claude` adapter fails the suite — otherwise new skills silently never load in Claude Code.

On validation: the 5 `tests/test_media_analysis.py` failures you saw do **not** reproduce on `origin/main` here (140 passed / 1 skipped, repo venv). Check your environment — bare `python3` on this repo's floor won't have the venv deps. Also main has moved since your rebase (v2.135.0); only `src/server.py` overlaps (comment-only on your side), so the rebase is trivial.

Happy to merge a version that keeps `.agents` as canonical **source** but treats `.claude` as a fully generated, content-complete output — same model as the per-IDE rules — rather than a thin pointer layer.

samuelgursky:
Landed — adapted directly on main in [v2.136.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.136.0) (commit 8836556), with credit in the commit and release notes. Thank you for the contribution: the `.agents/` layout, the Codex wiring, the `hook_runtime` refactor, and the portability test suite all landed essentially as you wrote them.

What changed relative to this branch, per the review:

- `.claude/skills` adapters are **content-complete byte-identical copies** of the canonical files instead of pointer stubs, so the original trigger descriptions, named craft-skill references, and `user-invocable` flags survive. A new sync script (`scripts/agent-rules/sync_portable_assets.py`) plus a parity guard (`tests/test_portable_asset_parity.py`) keep the two sides identical and catch the missing-adapter case.
- `.claude/agents/*` keep their frontmatter (including the `model: opus` pins); `.agents/roles/*` are exactly your frontmatter-stripped bodies.
- `CLAUDE.md` stays short (the `@AGENTS.md` inlining wasn't taken).
- `source_media_guard`'s scratch prefixes stay `claude-`/`codex-` — `agent-` was too broad an exemption for a hard-safety deny.
- The H1 de-branding landed on both copies equally.

On your suite run: the 5 `test_media_analysis` failures didn't reproduce here on main (likely an out-of-venv interpreter on your side), and main's suite is green at 3123 + 845 subtests with your portability tests included.

PR #174, opened 2026-08-31
Fix Windows UnicodeDecodeError in tests/scripts (missing encoding="utf-8")

Chosen-3:
## Summary
- `Path.read_text()` / `write_text()` / `open()` fall back to the OS locale encoding when `encoding=` isn't passed. On Windows (cp1252 by default) this crashes with `UnicodeDecodeError` as soon as a read target contains non-ASCII bytes — reproduced by running `python tests/test_import.py` on Windows without `PYTHONUTF8=1`, which fails parsing `src/server.py` at byte `0x90`.
- Patched every `open()` / `.open()` / `.read_text()` / `.write_text()` call in `tests/` and `scripts/` that omitted `encoding=`, via an AST-based rewrite (not regex) so binary-mode opens and calls that already pass `encoding=` were left untouched. 151 call sites across 47 files.
- One category needed manual exclusion: `PIL.Image.open()` also ends in `.open(...)` but doesn't accept `encoding=` — those 3 sites (in `scripts/contact_sheet.py` / `tests/test_contact_sheet.py`) were left as-is.

## Test plan
- [x] `python tests/test_import.py` passes with no `PYTHONUTF8=1` workaround (previously crashed with `UnicodeDecodeError` on Windows).
- [x] `python -m unittest tests.test_static_undefined_names tests.test_action_list_drift tests.test_panel_docs_drift tests.test_api_limitations_doc tests.test_doc_tool_counts tests.test_import` (the CI static-guard invocation) passes.
- [x] Full suite (`python -m unittest discover -s tests -p "test_*.py"`, 1862 tests) run before and after this change: identical failure/error set in both runs (27 failures / 186 errors / 5 skipped, all pre-existing and either live-Resolve-connection-dependent or platform-dependent — unrelated to this change).

### DIFF
--- a/scripts/audit_api_parity.py
+++ b/scripts/audit_api_parity.py
@@ -49,7 +49,7 @@ def parse_documented_methods(docs_path: Path) -> Dict[str, Set[str]]:
     """
     classes: Dict[str, Set[str]] = {}
     current_class: str | None = None
-    text = docs_path.read_text()
+    text = docs_path.read_text(encoding="utf-8")
     # Truncate at the first deprecation marker
     for marker in ("\nDeprecated Resolve API Functions",
                    "\nUnsupported Resolve API Functions"):
@@ -86,7 +86,7 @@ def collect_source_text() -> str:
         if "__pycache__" in py_path.parts:
             continue
         try:
-            parts.append(py_path.read_text())
+            parts.append(py_path.read_text(encoding="utf-8"))
         except (OSError, UnicodeDecodeError):
             continue
     return "\n".join(parts)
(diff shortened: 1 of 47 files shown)


samuelgursky:
Landed — applied to main verbatim in [v2.136.1](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.136.1) (commit ba31771), with credit in the commit and release notes. Thank you — this was a model contribution: tightly scoped, AST-swept rather than regexed, the PIL exemption caught, and honestly validated (the identical before/after failure set is exactly the right evidence for a mechanical change).

One addition on top: a companion AST guard (`tests/test_utf8_encoding_discipline.py`) now fails the suite on any new unencoded text-mode call in `tests/` or `scripts/`, with the same binary-mode and `Image.open` exemptions your sweep used — so the class stays fixed. Its first catch was a file written here a day after your branch, which is as good an argument for the guard as any.

(Housekeeping note: this repo generally lands contributor PRs by adapting them directly on main — the close is the merge, not a rejection.)

PR #177, opened 2026-09-01
feat(bridge): honor custom config path in installer

altiss:
## Summary

- make the free-edition bridge installer honor `DAVINCI_RESOLVE_BRIDGE_CONFIG`
- preserve the current default at `~/.config/davinci-resolve-mcp/bridge.json` when the variable is unset
- use the selected path consistently when creating, reporting, embedding, and copying the bridge configuration
- document the override for installations that keep the bridge config outside the default location

## Problem

The bridge client already reads `DAVINCI_RESOLVE_BRIDGE_CONFIG`, but `scripts/install_resolve_bridge.py` always wrote and embedded the fixed default path. A client configured with an override could therefore look in one location while the in-app Resolve bridge was installed with another. This is especially confusing in the free-edition workflow because both sides can be installed successfully yet fail to share the same token/configuration.

## Implementation

The installer now resolves the effective configuration path at call time through `config_path()`:

- an expanded `DAVINCI_RESOLVE_BRIDGE_CONFIG` value wins when present
- the existing default path remains unchanged otherwise
- `ensure_config()` reads and writes the selected path
- `install()` reports the selected path
- `_install_to()` copies the selected config into the Resolve runtime and embeds it in the launcher
- probe guidance points to the selected config directory

Resolving at call time also keeps tests and long-lived processes consistent with the bridge client's existing behavior.

## Compatibility and safety

- fully backward compatible when the environment variable is absent
- existing tokens and allowed media/output roots are preserved when reusing a valid custom config
- invalid existing JSON follows the installer's established safe-replacement behavior
- config directory/file permissions remain `0700`/`0600` on POSIX
- no changes to the bridge protocol or Resolve operation surface

## Tests

Added coverage for:

- default installer/client path parity
- environment override parity at call time
- `~` expansion
- writing, reporting, copying, and launcher embedding of a custom path
- preservation of an existing custom token and allowed roots
- safe replacement of invalid existing JSON
- probe report guidance under the selected config directory

Validation performed after rebasing onto current `main`:

```text
38 focused installer tests passed
tests/test_import.py: ok
scripts/audit_api_parity.py: PASS
git diff --check: clean
```

Focused branch coverage for `scripts/install_resolve_bridge.py` is 89% with branch measurement enabled. The custom-path code paths introduced here are exercised.

## Live validation

The installer was also exercised with a real non-default config path against DaVinci Resolve Free 21.0.4.5 on macOS. The installed launcher and runtime copy used the selected path, the existing token was preserved, and a read-only bridge request reached a davinci project that i own and its `Timeline 1` video successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,11 @@ python scripts/install_resolve_bridge.py
 # restart Resolve, open a project, then: Workspace > Scripts > resolve_bridge
 ```
 
+The installer and MCP client use `~/.config/davinci-resolve-mcp/bridge.json`
+by default. To keep the authenticated bridge config elsewhere, set
+`DAVINCI_RESOLVE_BRIDGE_CONFIG` while running the installer and in the MCP
+client environment; both sides will use that path.
+
 Once that listener is running it is used **automatically** whenever external
 scripting is unavailable — no environment variable required. Setting
 `DAVINCI_RESOLVE_BRIDGE=1` *forces* the bridge instead: it becomes the only
--- a/scripts/install_resolve_bridge.py
+++ b/scripts/install_resolve_bridge.py
@@ -24,9 +24,16 @@ import sys
 from pathlib import Path
 
 REPO = Path(__file__).resolve().parent.parent
-CONFIG_DIR = Path.home() / ".config/davinci-resolve-mcp"
-CONFIG_PATH = CONFIG_DIR / "bridge.json"
+DEFAULT_CONFIG_PATH = Path.home() / ".config/davinci-resolve-mcp/bridge.json"
+ENV_CONFIG_PATH = "DAVINCI_RESOLVE_BRIDGE_CONFIG"
 DEFAULT_PORT = 49632
+
+
+def config_path() -> Path:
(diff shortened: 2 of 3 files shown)


samuelgursky:
Landed on main in [v2.145.1](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.145.1) (commit on main), credited in the commit and release notes — thank you. The call-time resolution mirroring the client's exact semantics and the test coverage were both right, and the whole change landed essentially verbatim.

One correction on top: the step-4 probe guidance (and its test) followed the override, but `resolve_bridge_probe.py` runs *inside Resolve*, which never inherits the shell's `DAVINCI_RESOLVE_BRIDGE_CONFIG` — it always writes `host-model-probe.json` to the fixed `~/.config/davinci-resolve-mcp/`. Guidance that followed the override would point users at a file that never exists there, so it now stays on the fixed path with a comment explaining why (and the test asserts exactly that).

(As with the other recent PRs: this repo lands contributions by adapting them directly on main — the close is the merge.)

PR #181, opened 2026-09-03
implemented the Structured Operation Results envelope across the DaVi…

altiss:
### Summary

Establishes a standardized, predictable **Structured Operation Results** envelope across all 36 compound tools on the DaVinci Resolve MCP Server.

### Motivation & Problem

When orchestrating complex multi-turn editing workflows through AI agents (e.g., rough cuts, ripple edits, audio conforms), agents frequently struggle with **return shape heterogeneity**:
- **Ambiguous Confirmation Gates**: Destructive operations return a preview and `confirm_token`, but agents often mistake a successful plan validation (`success: true`) for completed timeline mutations.
- **Silent Contradictions & Partial Success**: While `verify_by_readback` and bulk summaries detect internal anomalies, agents have had to write ad-hoc conditional logic across disparate keys (`readback.missing`, `failures`, `partial`, `succeeded`/`failed`) to know whether an action truly completed, partially succeeded, or encountered a readback contradiction.
- **Tracking Semantic Deltas**: Determining the concrete editorial impact of an operation (e.g. *how many items moved or were added?*) required custom parsing of raw timeline geometry.

This PR establishes a common **Structured Operation Results** envelope across all 36 compound tools to give agents predictable, standardized operational context on every call.

---

### Common Envelope Schema

```json
{
  "status": "success",
  "operation": "timeline.ripple_insert",
  "execution_id": "exec_d2c123817bee",

  "result": {
    "insert_frame_absolute": 86400,
    "shift_frames": 48
  },

  "verification": {
    "status": "passed",
    "checks": [
      {"check": "readback_verification", "passed": true, "missing_items": 0}
    ],
    "contradiction": false
  },

  "changes": {
    "items_added": 3,
    "items_moved": 17,
    "items_deleted": 0
  },

  "warnings": []
}
```

### Key Capabilities

1. **Standard Envelope Schema**:
   - Returns `status`, `operation`, `execution_id`, `result`, `verification`, `changes`, and `warnings`.
   - Normalizes operational status into `"success" | "partial" | "blocked" | "failed"`.
2. **100% Backward Compatibility (`dual` mode default)**:
   - Emits structured envelope keys alongside mirrored top-level domain keys (`success`, `timelines`, `name`, etc.). Existing agent prompts and scripts continue working with zero breaking changes.
   - Strict `pure` mode is supported per-call (`params={"envelope": "pure"}`) or globally via `setup(action="set_defaults", params={"result_envelope": "pure"})` / `RESOLVE_MCP_RESULT_ENVELOPE=pure`.
3. **Automated Verification & Semantic Delta Extraction**:
   - Integrates readback audit results into structured verification entries (`passed`, `failed`, `partial`, `unverified`, `contradiction`).
   - Normalizes semantic change deltas (`items_added`, `items_moved`, `items_deleted`, `properties_updated`, metric before/after deltas).
4. **Decorator-Level Integration**:
   - Wrapped inside `_guard_missing_params` across all 36 compound tools, supporting both synchronous methods and asynchronous coroutine tools (`media_analysis`).
   - FastMCP binary returns (`Image`, `TextContent`, `EmbeddedResource`) pass through untouched.

### Verification & Testing

- **100% statement coverage** on `src/utils/operation_result.py` and all touched lines in `src/utils/readback.py` and `src/server.py`.
- 23 unit and integration tests in `tests/test_operation_result.py` pass cleanly in 0.006s.
- Tested and verified live against running **DaVinci Resolve 21.0.4.5** via the in-app bridge across `resolve_control`, `project_manager`, and `timeline` tools.
- All repository drift guards, contract tests, and API parity checks pass cleanly.
- Documentation added to `README.md` and `docs/SKILL.md`.

### After (Structured Envelope) change
- Eliminates False Success Hallucinations: Standardized status unambiguously tells the agent whether an action succeeded, partially failed, or is blocked waiting for confirmation.
- Drastically Cuts Token Consumption: Returning concise semantic changes (items added/moved/deleted) prevents agents from wasting thousands of tokens re-querying the entire timeline after every edit.
- Catches Silent API Contradictions Early: Normalizes readback audits into verification.status, forcing the agent to recover immediately when Resolve silently fails to execute a command.
- Enables Deterministic Multi-Turn Tracing: Attaching an execution_id to every mutation allows agents and test harnesses to reliably track, debug, and audit complex editing trajectories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,10 @@ This project treats camera originals and source media as immutable. Analysis too
 
 The default server is a local stdio process launched by your MCP client; it does not expose a network listener or built-in multi-user auth surface. The two opt-in local HTTP surfaces — the control panel and the networked MCP transport — bind loopback only and require a per-launch bearer token on every request, with Host/Origin checks against DNS rebinding and CSRF. Tool metadata includes MCP client-safety hints for read-only, destructive, idempotent, and external-resource operations. See [Security Policy](SECURITY.md) for operational boundaries, confirmation guidance, and vulnerability reporting.
 
+## Structured Operation Results
+
+Every compound MCP tool returns a standardized envelope (`status`, `operation`, `execution_id`, `result`, `verification`, `changes`, `warnings`). By default, the server operates in `dual` mode, retaining top-level keys for full backward compatibility while providing structured machine-readable metadata. Strict `pure` mode can be requested per-call via `params={"envelope": "pure"}` or globally configured using `setup(action="set_defaults", params={"result_envelope": "pure"})`. See [AI Skill Reference](docs/SKILL.md#structured-operation-results) for details and envelope schema.
+
 ## Key Stats
 
 | Metric | Value |
--- a/docs/SKILL.md
+++ b/docs/SKILL.md
@@ -2099,6 +2099,69 @@ reports, deletes the project, and removes generated media. See
 
 ---
 
+## Structured Operation Results
+
+Every compound MCP tool call returns a standardized result envelope to make server responses predictable and machine-readable across all workflows.
+
+### Envelope Structure
+
+```json
+{
+  "status": "success",
+  "operation": "timeline.ripple_insert",
+  "execution_id": "exec_d2c123817bee",
+
+  "result": {
+    "insert_frame_absolute": 86400,
+    "shift_frames": 48
+  },
+
+  "verification": {
+    "status": "passed",
+    "checks": [
+      {"check": "readback_verification", "passed": true, "missing_items": 0}
+    ],
+    "contradiction": false
+  },
+
+  "changes": {
+    "items_added": 3,
+    "items_moved": 17,
+    "items_deleted": 0
+  },
+
+  "warnings": []
+}
+```
+
+- **`status`**: `"success" | "partial" | "blocked" | "failed"`.
+  - `"success"`: the operation completed cleanly.
+  - `"partial"`: some sub-actions or items succeeded while others failed (e.g. bulk operations).
+  - `"blocked"`: the action is held waiting for user confirmation (e.g. dry-run confirmation token issued).
+  - `"failed"`: the action failed or returned an error.
+- **`operation`**: `{tool_name}.{action}` (e.g. `"timeline.ripple_insert"`, `"project_manager.get_current"`).
+- **`execution_id`**: unique traceable execution identifier (`exec_<uuid12>`), or correlated with `_versioning.analysis_run_id`.
+- **`result`**: dictionary holding the domain-specific action payload.
+- **`verification`**: readback verification status (`"passed"`, `"failed"`, `"partial"`, `"unverified"`, or `"contradiction"`) with individual audit checks.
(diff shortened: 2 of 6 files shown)


samuelgursky:
Thank you for this — the diagnosis is right and the envelope is a real gap. It shipped in [v2.205.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.205.0), adapted onto `main` and credited to your design in the module docstring, the changelog and the commit. Closing this PR as landed rather than asking you for revisions.

I did change the central design decision, and I want to be specific about why. **Dual mode was not backward compatible.** Five of the envelope's key names are already domain keys on this server:

| key | sites in `src/server.py` |
|---|---|
| `status` | 22 |
| `operation` | 20 |
| `warnings` | 15 |
| `result` | 8 |
| `changes` | 2 |

The mirror loop (`for k, v in raw_result.items(): if k not in envelope`) skips any key the envelope already defined, so those were silently dropped or shadowed. Concretely, on your branch:

- `resolve_control("job_status")` returned `status: "done"`; it became `"success"`. An agent polling a background job never sees it finish.
- A confirm gate's `status: "confirmation_required"` became `"blocked"` — renaming the very signal the PR's motivation section is about.
- `get_transcription` lost Resolve's own `status: "Transcribed"`.

That's what broke 23 existing tests. (No CI ran on PRs here, which is why it wasn't visible — that's now fixed too, and it's on me, not you.)

**The adaptation:** `dual` passes the payload through untouched and hangs the envelope off one reserved key, `_operation`, following the existing `_versioning` convention. Nothing is shadowed, nothing dropped, and there's still exactly one place to look. `pure` keeps your flat shape. A guard test fails the suite if any module starts returning `_operation` as a domain key.

Four smaller changes, all in the same spirit as your motivation:

- **`changes` is omitted when unreported, not `{}`.** An empty dict reads as "this operation changed nothing", which is false about a ripple insert that simply never declared its deltas. Same for `verification: "unverified"` — it means no evidence was reported, not checked-and-clean.
- **Status inference no longer keys on `blocked`.** It reads like a gate flag but is a domain key holding the *list of targets that could not be resolved*; a successful `bulk_match_to_hero` dry run carries a non-empty one, and would have reported a confirmation that was never requested.
- **Semantic deltas are declared by the action, not inferred from key names.** `properties_restored_items` → `properties_updated` turns a ripple insert's internal bookkeeping into an edit the caller never made. `timeline.ripple_insert` declares its own; the rest report none rather than a fabricated zero.
- **`result_envelope` is persisted** to `logs/server-preferences.json` and restored at startup. It lived in process memory and silently reset, while `_setup_defaults_snapshot` presented it alongside the two settings that do persist.

Your 23 tests are all represented in the 64 that landed, plus a `DomainKeysSurvive` class built from the exact payloads the flat envelope changed.

Verified through the real stdio JSON-RPC tool layer, not just at module level: 36 tools register, the envelope arrives, the payload is intact. Suite 3169 → 3233, zero failures.

Genuinely good idea — thanks for pushing it.

PR #183, opened 2026-09-04
feat(observability): agent execution traces and query APIs

altiss:
### Summary

Translates agent execution tracing and observability into DaVinci Resolve MCP to answer the question: **"Why did the editor do this?"**

While v2.205.0 introduced the `_operation` envelope on single tool returns, real editing workflows (`/resolve-tighten-recording`, `/resolve-rough-cut`, batch grading passes) consist of multi-turn tool loops. This PR introduces an in-memory execution trace engine that correlates multi-turn tool invocations, tracks tool timing (`duration_ms`), aggregates invocation counts, computes cumulative semantic deltas (`items_deleted`, `cuts_made`), and rolls up verification checks into unified execution traces.

---

### Why We Need This

1. **Multi-Turn Context & Explainability**: 
   When an agent removes 17 pauses across a timeline, individual tool returns show 17 single deletions. An execution trace aggregates those into a unified session showing the user's prompt ("Remove all pauses > 800ms"), the 17 aggregated deletions, cumulative duration, and final verification status.
2. **Natural Progression from v2.205.0**:
   Uses the `execution_id` and semantic deltas from the `_operation` envelope, automatically timing wall-clock duration and correlating operations across a session.
3. **Zero Overhead & Source-Safe**:
   Runs as an in-memory ring buffer (capped at 100 recent executions) with non-blocking, best-effort append-only JSONL persistence (`logs/execution-traces.jsonl`). It never blocks a tool call, never alters source media, and preserves the fixed 36-compound-tool contract.

---

### Key Additions & Architecture

- **`src/utils/execution_trace.py`**:
  - Thread-safe ring buffer (`maxlen=100`) for session traces.
  - Multi-turn correlation via `begin_execution(request, execution_id)` and `end_execution(...)`.
  - Automatic single-shot fallback: any tool call with an explicit `execution_id` is tracked and completed automatically.
  - Cumulative aggregation: aggregates tool calls by `tool` and `action`, summing `count` and `duration_ms`.
  - Semantic delta rollup (`items_deleted`, `items_added`, `tracks_modified`, etc.) and verification rollup (`passed`, `checks`, `contradiction`).
- **`src/server.py`**:
  - Wall-clock measurement via `time.perf_counter()` in `_guard_missing_params` populates `duration_ms` in `_operation`.
  - **Observer Isolation**: Queries (`get_execution_trace`, `list_recent_executions`, etc.) are exempt from step recording, preventing observer recursion from corrupting active traces.
  - Added query and lifecycle actions to `resolve_control` (`get_execution_trace`, `get_execution`, `list_recent_executions`, `begin_execution`, `end_execution`, `clear_executions`).
- **`src/utils/operation_result.py`**:
  - Propagates `execution_id` and `duration_ms` into the standard `_operation` envelope.
- **Documentation & Releases**:
  - `CHANGELOG.md`: Added `## What's New in v2.206.0 — agent execution traces and observability`.
  - `README.md` & `README.zh-CN.md`: Added section on execution traces and bumped badge to `v2.206.0`.
  - `docs/SKILL.md`: Added guide and JSON schema under "Agent Observability: Execution Traces".
  - Version bumped to `2.206.0` across `install.py`, `package.json`, `package-lock.json`, `src/server.py`, and `src/granular/common.py`.

---

### Verification & Testing Details

All tests and repository drift guards were executed against the virtual environment:

1. **Unit & Integration Suite (`tests/test_execution_trace.py`)**:
   - `test_single_step_automatic_trace`: Verifies single-shot tool execution auto-commits a valid trace.
   - `test_pause_deletion_aggregation_scenario`: Simulates a 17-step pause deletion workflow; verifies tool grouping, count summation (17), duration aggregation, and semantic delta accumulation (`items_deleted: 17`).
   - `test_multi_step_explicit_execution`: Verifies explicit lifecycle (`begin_execution` -> multiple tool steps -> `end_execution`).
   - `test_verification_rollup_success` & `test_verification_rollup_contradiction`: Verifies contradiction flags and checks aggregation.
   - `test_ring_buffer_capping`: Verifies buffer caps strictly at 100 traces without memory growth.
   - `test_thread_safety`: Verifies concurrent step recording across 8 parallel threads without race conditions or corrupted counts.
   - `test_resolve_control_trace_actions`: Verifies dispatch through `resolve_control`.
2. **Import & Version Parity (`tests/test_import.py`)**:
   - `test_npm_package_metadata`: **PASSED** (verifies package.json, install.py, src/server.py, and common.py are in 100% version lockstep).
3. **Drift Guards & Static Analysis**:
   - `test_doc_tool_counts.py`: **PASSED** (Compound tool count strictly preserved at 36).
   - `test_action_list_drift.py`: **PASSED** (`resolve_control` action list matches `_unknown` and docstring).
   - `test_static_undefined_names.py` (pyflakes): **PASSED** (0 undefined names).
   - `scripts/audit_api_parity.py`: **PASSED**.
   - `scripts/gen_api_limitations.py --check`: **PASSED**.
   - `scripts/audit_readwrite_symmetry.py --check`: **PASSED**.
   - `git diff --check`: Clean (no whitespace issues).

**Result**: 77/77 tests passed in 0.13s; all static drift guards and parity checks clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,38 @@
 
 Release history for the DaVinci Resolve MCP Server. The latest release is summarized in the root README; older entries live here to keep the README focused.
 
+## What's New in v2.206.0 — agent execution traces and observability
+
+Execution traces and observability translating Strands-style agent observability into DaVinci Resolve MCP to answer "Why did the editor do this?".
+
+### Added
+
+- **Agent execution traces & observability ("Why did the editor do this?")**:
+  Multi-step AI operations correlate across tool calls into unified execution
+  traces. Each trace captures the user request or prompt, per-tool timing
+  (`duration_ms`) and invocation counts, cumulative semantic deltas
+  (`items_deleted`, `items_added`), and readback verifications.
+- **Trace query and lifecycle actions on `resolve_control`**:
+  - `get_execution_trace(execution_id?)` / `get_execution(execution_id)`: Look up
+    correlated trace by ID, or the most recent execution if omitted.
+  - `list_recent_executions(limit?)`: List recent execution summaries (newest first).
+  - `begin_execution(request?, execution_id?, initiator?)`: Open a scoped execution
+    so subsequent tool calls in the session automatically thread under this ID.
+  - `end_execution(execution_id?, verification?, status?, notes?)`: Conclude an
+    execution trace and compute aggregated rollups.
+  - `clear_executions(dry_run?)`: Reset the in-memory execution trace buffer.
+- **Direct Python API**: Exported `get_execution_trace`, `get_execution`,
+  `list_recent_executions`, `begin_execution`, `end_execution`, `clear_executions`
+  from `src/utils/execution_trace` and `src/server`.
+
+### Notes on tracing and observability
+
+- **Observer isolation**: Querying `get_execution_trace` or `list_recent_executions`
+  does not record an execution step, preventing observer recursion from corrupting
+  the active trace.
+- **Envelope integration**: The standard `_operation` envelope carries `execution_id`
+  and measured `duration_ms` on every compound tool response.
+
 ## What's New in v2.205.0 — a standard operation envelope on every tool result
 
 Adapted from the design contributed in PR #181.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 English | [简体中文](README.zh-CN.md)
 
-[![Version](https://img.shields.io/badge/version-2.205.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
+[![Version](https://img.shields.io/badge/version-2.206.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
 [![npm](https://img.shields.io/npm/v/davinci-resolve-mcp.svg?label=npm&color=CB3837)](https://www.npmjs.com/package/davinci-resolve-mcp)
 [![API Coverage](https://img.shields.io/badge/API%20Coverage-100%25-brightgreen.svg)](docs/reference/api-coverage.md)
 [![Tools](https://img.shields.io/badge/MCP%20Tools-36%20(353%20full)-blue.svg)](#server-modes)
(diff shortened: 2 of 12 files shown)


samuelgursky:
Landed as [v2.206.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.206.0), credited to your design in the module docstring, the changelog and the commit. Closing this as merged-by-adaptation rather than asking for revisions — it needed a rebase past v2.205.1/.2 anyway.

The framing is right: the `_operation` envelope describes one call, and a tighten pass is a loop of them. The 17-deletions example is exactly the case where per-call returns are individually true and collectively useless.

The code held up well under review — the lock scope is correct (persistence runs *outside* it, which I checked specifically because disk I/O under a global lock would serialize every tool call under threaded dispatch), the ring buffer is properly bounded, and observer isolation on the query actions is a good catch that I would not have thought of.

Four changes:

**1. The trace log was anchored to `os.getcwd()`, so on a standard install it wrote nothing.** `_trace_log_path()` returned `None` when `./logs` didn't exist, and the client configs `install.py` generates set no `cwd` — so for most users the file simply never appeared, silently, with no signal either way. I moved it beside `server.log`, which is how `media-analysis-preferences.json` and `server-preferences.json` already resolve, and the directory is created on first write instead of being a precondition.

**2. `list_recent_executions` now reports the log path and whether it's writable.** Best-effort persistence must never fail a real edit — agreed — but that's precisely what makes a broken destination invisible. "The file is empty" and "nothing is being written" looked identical from the caller's side.

**3. The log rotates at 8 MB, keeping one generation.** The in-memory ring is capped at 100; the file had no bound at all, at one append per tool call, on machines that stay up for months. My own test runs put it at 752 KB in an afternoon.

**4. "Non-blocking" — I measured it before changing the wording.** ~0.065 ms per `record_step` with persistence on vs ~0.086 ms with it off, i.e. indistinguishable from noise and utterly immaterial next to any Resolve round-trip. So the *cost* claim is fine; it's the word that isn't — it's a synchronous buffered append on the calling thread. The docstring now says cheap rather than non-blocking, and notes it runs outside the lock, which is the property that actually matters.

I also documented what's recorded (tool, action, timing, status, deltas, verification) and what isn't (parameters, file paths, clip and project names). That distinction is worth stating plainly for a tool people point at client footage — the one free-text field is the `request` you pass to `begin_execution`.

Your 7 tests plus 16 more, covering the path anchoring, the persistence reporting, rotation, and that a failed write can never reach the caller. Verified through the real stdio tool layer: 36 tools still register, and a begin/call/end cycle produces one correlated trace at the path it claims. Suite 3245 → 3268.

Second good idea in a row — thanks.

Issue #179, opened 2026-09-02
davinci-resolve-advanced cannot boot from managed install: SYNC_ITEMS omits resolve-advanced/


LambdaXIII:
## Summary

`davinci-resolve-advanced` cannot boot from the managed install produced by `npx davinci-resolve-mcp setup` in v2.203.0: the bin is registered into every generated client config, but the managed install never receives the `resolve-advanced/` module it imports. The process dies with `ERR_MODULE_NOT_FOUND` before the MCP handshake, so clients report "subprocess closed stdout before responding".

## Environment

- `davinci-resolve-mcp` v2.203.0 (latest release; also current on `main`)
- Windows, managed install at `%LOCALAPPDATA%\davinci-resolve-mcp`, Node 26.7.0
- MCP client: omp (error: `MCP tool load failed ... "MCP subprocess closed stdout before responding"`)

## Root cause

1. `install.py::build_advanced_entry()` / `generate_manual_config()` (v2.203.0) register `davinci-resolve-advanced` in every supported client config (standard snippet, VS Code, etc.), pointing at `<managed-root>\bin\davinci-resolve-advanced-mcp.mjs`.
2. `bin\davinci-resolve-advanced-mcp.mjs` imports `../resolve-advanced/server/index.mjs` (line ~52: `const serverEntry = path.resolve(__dirname, '..', 'resolve-advanced', 'server', 'index.mjs')`).
3. The bootstrapper's sync (`bin\davinci-resolve-mcp.mjs::SYNC_ITEMS`, lines 24–36) copies only `bin`, `src`, `docs`, `examples`, `scripts`, `install.py`, `README.md`, `CHANGELOG.md`, `LICENSE`, `SECURITY.md`, `AGENTS.md`, `CLAUDE.md`, `package.json` — **`resolve-advanced/` is not in the list**, and `install.py` never installs it either.

The npm package itself is fine: `npm pack davinci-resolve-mcp@2.203.0 --dry-run` shows 546 files including 326 under `resolve-advanced/`. The same SYNC_ITEMS omission exists on `main`. So every client config generated by v2.203.0 references a module that can never exist in the managed install.

## Reproduction

```text
> node "%LOCALAPPDATA%\davinci-resolve-mcp\bin\davinci-resolve-advanced-mcp.mjs"
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '...\davinci-resolve-mcp\resolve-advanced\server\index.mjs'
```

[...]

PR #185, opened 2026-09-04
feat(audit): export execution trace reports

Rohitkanithi:
### Summary

Adds export support for execution trace audit reports.

The existing execution trace layer can show traces through query actions, but there was no direct way to save one trace as a reviewable artifact. This adds `export_execution_report`, which writes a Markdown or JSON report for the latest trace or a specific `execution_id`.

The report includes request metadata, status, timing, tool summary, semantic changes, verification rollup, warnings, notes, and optional step details.

---

### Why This Is Useful

Execution traces are useful while the server is running, but review often happens after the workflow is complete.

This gives users a simple way to keep a readable audit record of what happened during a multi-step operation:

1. what request started the workflow
2. which tools ran and how often
3. how long the workflow took
4. what changed
5. whether verification passed
6. whether any warnings were recorded

This is especially useful when reviewing unexpected edits or sharing the result of a completed workflow.

---

### Key Additions

- Added Markdown and JSON report rendering for execution traces
- Added `export_execution_report` utility in `src/utils/execution_trace.py`
- Added `resolve_control("export_execution_report")`
- Default reports to `logs/execution-reports/<execution_id>.md`
- Added `RESOLVE_MCP_TRACE_REPORT_DIR` override for report output location
- Protected existing files unless `overwrite=true`
- Added `include_steps=false` for shorter summary reports
- Kept export as an observer action so exporting a report does not modify the trace being reviewed
- Updated version surfaces to `v2.207.0`
- Updated README, `docs/SKILL.md`, and changelog

---

### Testing

Added focused unit and integration coverage for:

- Markdown report export
- JSON report export
- step omission with `include_steps=false`
- invalid format handling
- overwrite protection
- server dispatch through `resolve_control`
- observer isolation for the export action
- direct server import coverage

Commands run:

```bash
venv/bin/python -m unittest tests.test_execution_trace
venv/bin/python tests/test_import.py
venv/bin/python -m unittest tests.test_action_list_drift tests.test_static_undefined_names tests.test_duplicate_definitions tests.test_doc_tool_counts
venv/bin/python -m py_compile src/utils/execution_trace.py src/server.py src/granular/common.py install.py
git diff --check
```

Full test discovery was also attempted, but this local environment is missing optional dependencies (`numpy`, `requests`), so the focused suite and repository drift guards above are the verified checks for this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,34 @@
 
 Release history for the DaVinci Resolve MCP Server. The latest release is summarized in the root README; older entries live here to keep the README focused.
 
+## What's New in v2.207.0 — execution audit report exports
+
+### Added
+
+- **Execution traces can now be exported as reviewable audit reports.**
+  `resolve_control(action="export_execution_report")` writes the latest trace,
+  or a named `execution_id`, as Markdown or JSON. The report carries the
+  request, status, start/end timestamps, duration, tool summary, semantic
+  deltas, verification rollup, warnings, notes, and optional per-step table.
+- **Reports default beside the trace log.** When no path is passed, reports are
+  written under `logs/execution-reports/<execution_id>.md` or `.json`.
+  `RESOLVE_MCP_TRACE_REPORT_DIR` can move that default destination without
+  changing where append-only trace events are logged.
+- **Exports are observer-safe.** Creating a report is exempt from execution-step
+  recording, just like querying traces, so inspecting or exporting a trace
+  cannot mutate the trace being reviewed.
+- **Existing files are protected by default.** A caller must pass
+  `overwrite=true` to replace a report at the chosen path.
+
+### Notes
+
+- The export is built from the existing trace summary fields, not raw tool
+  arguments or raw tool results. It is meant for review and audit, not a replay
+  script.
+- Added focused unit and server integration coverage for Markdown export, JSON
+  export, step omission, invalid formats, overwrite protection, observer
+  isolation, and `resolve_control` dispatch.
+
 ## What's New in v2.206.0 — agent execution traces
 
 Adapted from the design contributed in PR #183.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 English | [简体中文](README.zh-CN.md)
 
-[![Version](https://img.shields.io/badge/version-2.206.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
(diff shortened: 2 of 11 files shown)


samuelgursky:
Landed as [v2.207.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.207.0) — GitHub marked this merged rather than closed, since the branch I released carried your commit. Thanks for this, and for the analysis on #179 earlier.

It reviewed well. Observer isolation was correctly wired into `_TRACE_OBSERVER_ACTIONS` rather than bolted on at the callsite, the action list stayed in sync with `_unknown`, overwrite protection is right, and `_report_filename` sanitises `execution_id` down to a bare filename — I tested `../../../../etc/passwd` and got `etc_passwd.md`, which is the route most likely to be agent-influenced and the one that mattered most to get right.

**One thing I fixed on the way in, and it's worth explaining because it wasn't yours.** A report for a run where nothing was verified rendered:

```
| Status | unverified |
| Passed | yes        |
| Checks | 0          |
```

The rollup in `_merge_verification` had `else: status = "unverified"; passed = True` — inherited from #183, not introduced here — and your renderer printed the field faithfully. But those two lines sit inches apart in a document meant to be reviewed and shared, and only one of them gets scanned. v2.206.0 had just documented that `unverified` means *no evidence was reported*, not "checked and clean"; your export is exactly where that stops being a nuance in a JSON blob and becomes a claim on paper. So the rollup now carries `None` for the unknown case and the row reads `not established — no checks recorded`. A real pass still says yes, a real failure still says no — all three verified.

Your export is what surfaced it. A latent field default nobody would have questioned became visibly wrong the moment it was rendered for a human, which is a good argument for the feature.

Two smaller notes:

- **`path` is honoured as given and creates directories to reach it.** I kept that — a conform's paperwork belongs beside the conform, not in `logs/` — but documented it next to the source-media safety policy so nobody points it at a footage folder by reflex.
- **The zh-CN README** had its badge and `本翻译对应 vX.Y.Z 版 README` line bumped to 2.207.0 without the section. That's the specific case the release process calls out: a translation whose version line asserts it's current when it isn't is worse than no translation. Translated it — easy to miss, and no drift guard covers it.

On your note about the local environment missing `numpy` and `requests`: the full suite needs `pip install pytest numpy pillow requests` on top of `requirements.txt`. CI runs it on every PR now — though it needed my approval here since it's your first PR from a fork, and that turned out to be blocked on my end, so I ran it locally instead: 3264 passed against a 3260 baseline, identical known-failure set. Your four cases plus five of mine.

PR #188, opened 2026-09-05
fix(edit-engine): variant item counts come from the timeline, not the append reply

mart0vip:
## The bug

`edit_engine.execute_silence_ripple` returns an `audio_accounting` block whose item counts understate what it actually built.

Observed on Resolve 21.0.4.5 (free edition, in-app bridge), executing a plan against a 189336-frame timeline with 423 silence lifts:

```json
"audio_accounting": {
  "planned_audio_ranges": 432,
  "planned_video_ranges": 432,
  "variant_audio_items": 250,
  "variant_video_items": 250
}
```

The variant really held **432 video + 432 audio items** — confirmed via `timeline_item get_items_in_track` on both tracks: start/end sequences byte-identical between tracks, contiguous from frame 108000 to 288600, `detect_gaps_overlaps` returns 0 gaps and 0 overlaps, and `sum(duration)` is 180600 on each track, matching the timeline's content length exactly. The same response's `readback.after.clip_count` reported 864 — correct.

`planned 432, got 250` reads exactly like 182 ranges silently failing to land. On a silence ripple the operator's main fear is dropped material, and establishing that this was benign cost a full review cycle of hand-auditing both tracks.

Still reproduces on `main` @ v2.207.0: `resolve_bridge_ops.py` is byte-identical to v2.93.0 (same blob hash) and the `audio_accounting` block is unchanged.

## Root cause

`ResolveOperations._encode` truncates every proxied container to `max_items` (default 500) with **no signal anywhere**:

```python
return [self._encode(v, depth + 1, shape) for v in list(value)[: self.max_items]]
```

`plan_silence_ripple` emits keep ranges interleaved video/audio, so 432 kept segments become 864 clipInfos in one `AppendToTimeline`. Resolve placed all 864 and returned all 864; the bridge encoded the first 500 — which, interleaved, is exactly **250 video + 250 audio**.

`readback.after.clip_count` was right the whole time because `capture_timeline_clip_count` re-reads `GetItemListInTrack` per track — 432 each, comfortably under the ceiling. Two numbers in one response, from two sources, only one of them truncated.

A short list is indistinguishable from a genuinely short result, which is what made this cost a review cycle rather than a glance.

## Fixes

**1. The bridge no longer truncates silently** — `resolve_bridge_ops.py`, `resolve_bridge_client.py`

Every dropped element is recorded; `op_call` / `op_get_attribute` carry a `truncated` block (dropped / limit / containers); the client records it on `transport.truncations` and logs a warning naming the method.

It **warns rather than raises**, deliberately: the native call has already run by the time the reply is encoded, so raising would turn a completed 864-item assembly into an error and orphan the timeline.

The default ceiling goes 500 → 2000, and is now clamped to `MAX_HANDLES` rather than the previous 5000 — a list longer than the handle table evicts its own earliest entries while it is still being minted, handing the client handles that are already `stale_handle`. A short list beats a poisoned one.

**2. Counts are read back from the timeline** — `server.py`

`create_variant_from_ranges` now returns `placed_item_counts` from a post-assembly per-track re-read, reusing the conform snapshot it already takes for gap detection, so this costs no extra Resolve calls.

`execute_silence_ripple` and `execute_tighten` now share one `_variant_audio_accounting` helper — tighten carried the identical bug — and a planned-vs-placed disagreement is stated outright instead of being left to a hand audit.

This mattered beyond reporting: under the old cap, a `cdl` on a large variant only reached the first 250 video items.

## Tests

- **`tests/test_variant_audio_accounting.py`** (new) — drives the real `execute_silence_ripple`, confirm-token flow included, against stubs that place all 864 clipInfos but return only 500. Fails on `main` with the exact field payload (`variant_video_items: 250` vs `planned_video_ranges: 432`).
- **`tests/test_resolve_bridge.py`** — 5 tests pinning the truncation signal, that truncation state does not leak between calls, and both ceiling invariants.
- **`tests/live_edit_engine_validation.py`** — the existing check was `variant_audio_items >= 1`, which passes the buggy 250. It now asserts equality with the per-track readback.

## Validation

- Full offline suite: **3265 tests, 0 failures, 0 errors** (26 skipped)
- `test_import.py`, `audit_api_parity.py`, `gen_api_limitations.py --check`, `agent-rules/generate.mjs --check`, `git diff --check` — all clean

**Live Resolve validation has not been run.** It needs synthetic media in a disposable project per `AGENTS.md`, which I did not set up. The live assertion is in place for whoever runs `live_edit_engine_validation.py` next.

No `api_truth` entry was added: this is a defect in this project, not a Blackmagic API limitation.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -6539,6 +6539,26 @@ def _variant_item_placement(item) -> Dict[str, Any]:
     }
 
 
+def _snapshot_track_item_counts(snapshot: Dict[str, Any]) -> Dict[str, int]:
+    """Per-track-type item counts read from a conform snapshot of a live timeline.
+
+    This is the ONLY honest answer to "what did the assembly actually place".
+    The obvious alternative — counting what `MediaPool.AppendToTimeline`
+    returned — is a witness derived from the same call it would be checking, and
+    it lies in two measured ways: the in-app bridge caps any proxied list at
+    `max_items` (an 864-clipInfo append came back as 500 items, so a variant
+    holding 432 video + 432 audio was reported as 250 + 250), and Resolve drops
+    colliding records from the reply without an error (see the api_truth entry
+    "MediaPool.AppendToTimeline (overlapping records — earlier item wins)").
+    Re-reading the timeline per track cannot be fooled by either.
+    """
+    counts: Dict[str, int] = {}
+    for track_type, block in (snapshot.get("tracks") or {}).items():
+        rows = (block or {}).get("tracks") or []
+        counts[str(track_type)] = sum(int(row.get("item_count") or 0) for row in rows)
+    return counts
+
+
 def _variant_audio_summary(built):
     """Video/audio range counts for an assembled variant, warning when it carries
     no audio. create_variant_from_ranges places exactly the ranges given, so a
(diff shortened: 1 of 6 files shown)


samuelgursky:
Merged and released as [v2.208.1](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.208.1), unchanged. This is the best-diagnosed bug report this repo has had — thank you.

I verified it end to end rather than taking it on trust:

- The arithmetic is exact. 432 interleaved ranges → 864 clipInfos → first 500 encoded → 250 video, 250 audio. Matches your observed payload to the item.
- **Your regression test fails on `main` with the reported numbers** — `AssertionError: 250 != 432` carrying the identical `audio_accounting` block — and passes on your branch. That is the whole ballgame for a test like this, and it's rarer than it should be.
- The `MAX_HANDLES` point holds: 5000 clamped against a 4096 table evicts 904 handles before the reply is sent.

The framing is what makes it a good report. *"A short list is indistinguishable from a genuinely short result"* is the failure, not the bound — and `_encode` was the one path that didn't say so while `list_timelines` and `list_clips` both already returned a `truncated` flag. One route disagreeing with its siblings.

The warn-rather-than-raise call is right and worth stating for the record: by the time the reply is encoded the native call has already run, so raising would convert a completed 864-item assembly into an error and orphan the timeline. Turning a reporting bug into a data-loss bug would have been a poor trade.

Also glad you caught `execute_tighten` carrying the same defect, and the `>= 1` live assertion that would have passed the buggy 250 — a test that cannot fail for the reason it exists is worse than no test.

**One process note:** I squash-merged rather than a plain merge, purely to keep the `Co-Authored-By: Claude` trailer out of this repo's history — that's a standing convention here, not a comment on the work. Your authorship is intact on the commit (`Author: Martin Canteros`), and the message is your own analysis. Flagging it so it isn't a surprise if you go looking for the original SHA.

Suite 3299 → 3307 with your eight. Live validation still unrun, as you said — the assertion is in place for whoever gets there next.

mart0vip:
Thanks — and the squash is completely fine. Keeping the trailer out of your history is your call to make, and the authorship came through intact.

Confirmed on the released build: pulled v2.211.0 and the regression test passes there.

On the live validation — I'd rather close that out than leave it for the next person. I'd run `live_edit_engine_validation.py` against a disposable project with synthetic media per AGENTS.md, on Resolve 21.0.4.5 free edition through the in-app bridge. That's the configuration the original report came from, so it exercises the encode path that actually caused this, rather than direct scripting where the ceiling never bites. I'll post the check output here either way — including if the new equality assertion turns something up.

The one leg I can't cover from here is Studio via direct scripting, if you'd want that verified separately.

PR #187, opened 2026-09-05
feat(observability): agent execution lifecycle, pre-flight risk inspection, and hooks

altiss:
### Summary

Introduces a structured execution lifecycle pipeline, pre-flight operation risk inspection, and simulation/dry-run capabilities across tool dispatches.

Building on the execution tracing infrastructure, agents and client harnesses previously had no unified mechanism to inspect operation blast radius prior to invocation, enforce pre-flight safety policies, intercept parameters for dry-run simulation, or attach lifecycle hooks across tool calls.

This PR adds:
1. **Agent Execution Lifecycle Hooks Pipeline**: `before_tool_call`, `after_tool_call`, and `on_error` interceptors integrated into the primary tool dispatch wrapper (`_guard_missing_params`).
2. **Pre-Flight Operation Risk Assessment**: `resolve_control(action="inspect_operation")` analyzes any tool action and parameters to return risk level (`low`, `medium`, `high`, `critical`), destructive indicators, blast radius scope (`item`, `track`, `timeline`, `project`, `system`), and confirmation requirements.
3. **Dry-Run Simulation Interceptor**: Simulated execution fallback for actions invoked with `dry_run: true` that lack native dry-run support, returning simulated impact, parameters, and risk profiles without mutating the live project.
4. **Lifecycle Hooks Introspection**: `resolve_control(action="list_lifecycle_hooks")` exposes registered lifecycle interceptors and their operational status.

---

### Key Additions

- Implemented `src/utils/execution_lifecycle.py`:
  - `ExecutionLifecycleManager`: registration and dispatch of `before_tool_call`, `after_tool_call`, and `on_error` hooks.
  - `inspect_operation`: rule-based static risk evaluation mapping destructive tools and operations to blast radius, confirmation requirements, and severity tiers.
  - Interceptors for safe simulation of unsupported `dry_run: true` calls.
- Integrated lifecycle processing into `src/server.py` within `_guard_missing_params`.
- Added actions to `resolve_control`:
  - `inspect_operation(tool?, target_action?, target_params?)`
  - `list_lifecycle_hooks()`
- Updated version metadata to `v2.208.0` across:
  - `src/server.py`
  - `src/granular/common.py`
  - `install.py`
  - `package.json`
  - `package-lock.json`
  - `README.md`
  - `README.zh-CN.md`
  - `CHANGELOG.md`
  - `docs/SKILL.md`
- Added comprehensive unit tests in `tests/test_execution_lifecycle.py`.

---

### Testing & Verification

[...]

PR #189, opened 2026-09-06
feat(security): add safe operations policy for destructive actions

Rohitkanithi:
## Summary

This PR adds a safe operations policy for destructive DaVinci Resolve actions.

The goal is to make mutating operations easier to inspect, audit, and control without changing the existing tool flow. The current destructive-action wrapper already handles version-on-mutate behavior and confirmation-token gating for sensitive operations. This change builds on that existing path by adding explicit risk classification, safe-mode enforcement, operation IDs, and structured audit logging.

With this change, destructive operations can report what level of risk they carry, high-risk operations can be blocked before they reach the Resolve handler when safe mode is enabled, and every allowed or blocked destructive call can be traced through a JSONL audit record.

### Risk classification

Destructive actions now receive a risk level:

- `read`
- `low`
- `medium`
- `high`
- `dangerous`

The default behavior is conservative:

- known low-risk edits, such as marker/color helper actions, are classified as `low`
- known high-risk actions, such as deleting media pool clips/folders/timelines, deleting tracks, overwriting ranges, lifting ranges, applying cuts, and resetting grades, are classified as `high`
- registered destructive actions without a more specific classification default to `medium`
- non-destructive actions remain `read`

This gives callers a stable field they can inspect before deciding how to display, confirm, retry, or block an operation.

### Safe mode enforcement

A new `destructive.safe_mode` preference is available through `setup`.

When enabled, safe mode blocks `high` and `dangerous` destructive actions before the underlying Resolve handler runs.

Blocked calls return a structured response with:

- `success: false`
- `status: blocked_by_security_policy`
- `operation_id`
- `security.risk_level`
- `security.safe_mode`
- `security.blocked`
- `error.code: SAFE_MODE_BLOCKED`
- remediation text explaining how to proceed

A reviewed one-off operation can still be allowed by passing:

```json
{
  "allow_risky_operation": true
}
```

This keeps safe mode useful without making it impossible to intentionally perform a high-risk edit after review.

### Operation IDs

Destructive operations now receive an `operation_id`.

The ID is returned in the tool response and written into the audit log. This makes it possible to correlate a user-facing result with the backend audit trail.

Example shape:

```json
{
  "operation_id": "op_abc123def456",
  "security": {
    "risk_level": "high",
    "safe_mode": true,
    "blocked": true
  }
}
```

### Security metadata on responses

Allowed and blocked destructive operation responses are annotated with a `security` object.

This gives callers a consistent way to inspect policy behavior without needing to infer risk from tool/action names.

Example:

```json
{
  "success": true,
  "operation_id": "op_abc123def456",
  "security": {
    "risk_level": "medium",
    "safe_mode": false,
    "blocked": false
  }
}
```

### Audit logging

This PR adds JSONL audit logging for destructive operations.

By default, audit records are written to:

```txt
logs/security-audit.jsonl
```

The audit log captures:

- timestamp
- operation ID
- tool name
- action name
- risk level
- status
- reason, when applicable
- analysis run ID, when available
- project root, when available
- sanitized params

The audit path can be configured with:

```json
{
  "destructive": {
    "audit_log_path": "/path/to/security-audit.jsonl"
  }
}
```

Audit logging can be disabled with:

```json
{
  "destructive": {
    "audit_log": false
  }
}
```

### Confirmation-token redaction

Audit records redact confirmation tokens from params.

Both of these keys are redacted:

- `confirm_token`
- `confirmToken`

This keeps the audit trail useful for debugging without recording live confirmation tokens.

### Pending confirmation auditing

Calls that issue a confirmation-token response are now audited as pending confirmation instead of being treated as already-mutated operations.

This preserves the existing behavior where token-preview calls do not create unnecessary version archives, while still giving the security log a record that a sensitive operation was requested.

### Setup integration

The following destructive-operation defaults are now exposed through `setup`:

- `destructive.require_confirm_token`
- `destructive.safe_mode`
- `destructive.audit_log`
- `destructive.audit_log_path`

They are included in:

- `setup(action="schema")`
- `setup(action="get_defaults")`
- `setup(action="set_defaults")`
- `setup(action="clear_defaults")`

This keeps destructive-operation policy configuration in the same setup surface as the existing defaults.

Example:

```json
{
  "action": "set_defaults",
  "params": {
    "destructive": {
      "safe_mode": true,
      "audit_log": true
    }
  }
}
```

### Release surfaces

This PR bumps the release surface to `v2.209.0` and adds a changelog entry for the safe operations policy.

Updated release/version files:

- `package.json`
- `package-lock.json`
- `src/server.py`
- `src/granular/common.py`
- `install.py`
- `README.md`
- `README.zh-CN.md`
- `CHANGELOG.md`

## Why This Is Useful

Destructive timeline, media pool, grading, and render-related operations need stronger safety signals than ordinary read-only actions.

Before this change, some destructive operations already had confirmation-token and versioning protection, but callers did not have a consistent policy object that explained:

- how risky the operation is
- whether safe mode is active
- whether the operation was blocked
- which audit record corresponds to the operation
- whether a confirmation-token request was only a pending decision

This PR makes that information explicit.

That gives users and integrations a clearer trust boundary around high-impact Resolve actions, especially operations that delete project data, rewrite timeline structure, or replace grades.

## Compatibility

This change is designed to preserve existing behavior by default.

- Safe mode defaults to `false`
- Audit logging defaults to `true`
- Existing confirmation-token behavior remains unchanged
- Existing version-on-mutate behavior remains unchanged
- Existing destructive-action registration remains the source of truth for which actions are considered mutating
- Non-destructive actions are not routed through the new policy path

Existing callers should continue to work. The main visible difference is that destructive operation responses may now include additional `operation_id` and `security` metadata.

## Validation

The following checks were run:

```bash
python -m py_compile src/utils/destructive_hook.py src/server.py src/granular/common.py install.py tests/test_destructive_hook.py tests/test_release_surface_drift.py
```

```bash
python -m unittest tests.test_destructive_hook tests.test_release_surface_drift
```

Both passed.


samuelgursky:
Landed in v2.209.0 — thanks, this was a well-scoped piece of work and the rebase onto current main made it easy to verify.

Suite on your head commit in the primary checkout: **3311 passed, 1 skipped, 0 failed**, all drift guards clean. After adaptation: 3314 (three added tests).

One change on the way in. Risk classification had landed twice: `execution_lifecycle.RiskClassificationHook` already classified every tool call for pre-flight `inspect_operation` (v2.208.0), and this PR added a second table with a second vocabulary (`read`/`low`/`medium`/`high`/`dangerous` vs the existing `low`/`medium`/`high`/`critical`). Two classifiers over the same calls can disagree — the gate refusing what inspection had just called reversible — so I collapsed them onto `execution_lifecycle` and moved your table's knowledge into it. Your action ratings were the better ones and they all survived: `lift_range`, `overwrite_range`, `apply_cuts`, `delete_track`, `delete_folders` and `reset_all_grades` now read `high` *everywhere* risk is reported, not just in the safe-mode gate, and the marker/clip-colour set reads `low` instead of falling through the name heuristic as unrecognised `medium`. `media_pool.delete_clips` reports `critical` rather than `high` — safe mode gates both, so the behaviour is identical.

I also tried making safe mode fail closed on actions the classifier doesn't recognise, and backed it out after measuring: only 28 of the 108 registered destructive actions match any rule, so it would have blocked `timeline.add_track`, `timeline.duplicate`, `media_pool.create_timeline` — three quarters of all edits — and taught people to switch safe mode off. A narrow gate users keep on beats a broad one they disable. What survives is reporting: `risk_established` now appears in the `security` block and in every audit record, so an assessed level and a name-based default are distinguishable. The rejected design and its measurement are in the docstring so it doesn't get re-proposed blind.

Everything else shipped as written — `operation_id`, the JSONL audit log with token redaction, auditing token-preview calls as pending rather than mutated, and the `destructive.*` setup surface.

Rating those 80 unclassified actions properly is the obvious follow-up and is queued separately. Happy to have you take it if you're interested.

Rohitkanithi:
Thanks @samuelgursky for landing this and for the careful adaptation. Collapsing the duplicate classifier into the existing lifecycle path makes sense, especially to avoid inspection and enforcement disagreeing.

I’d be happy to take the follow-up for rating the remaining unclassified destructive actions. I’ll keep it scoped to the existing `execution_lifecycle.RiskClassificationHook` and avoid adding another policy surface.

Before I start, do you want that PR to focus only on expanding the classifications, or should I also include tests around `inspect_operation` / safe-mode audit output for the newly classified actions?

samuelgursky:
Both, please — and the tests are the more important half.

The reason the classifier got collapsed into one is precisely that `inspect_operation` and the safe-mode gate must not diverge. Expanding the tables without pinning that would leave the two surfaces free to drift apart again a release later, which is the bug the collapse was meant to make structurally impossible rather than merely currently-false. So: for each newly classified action, assert the level `inspect_operation` reports and the level the gate acts on are the same value, and assert the audit row carries `risk_established: true`. That last one is the cheap regression signal — an action that quietly falls back to the name heuristic will still report a plausible `medium`, and the flag is the only thing that distinguishes an assessment from a default.

Worth having a guard test too: every pair in `DESTRUCTIVE_ACTIONS_BY_TOOL` classifies with `recognised=True`. That way a newly registered destructive action can't silently rejoin the unrated set — which is how the current 80 accumulated in the first place.

Two things from the live pass on v2.209.0 that are worth having before you start, since they change what the ratings need to be right about.

**The gate genuinely blocks, so a wrong rating has teeth.** Verified against Resolve Studio 19.1.3.7: safe mode refused `timeline.delete_track` with the track count unchanged, and the refusal lands *before* the confirm-token gate, so no token is minted for a call that cannot run. A `high` you assign is a call somebody's session will actually refuse — please rate from each action's handler in `src/server.py` or `src/granular/`, not from the name. The name heuristic is the thing being replaced, so inheriting its guesses would defeat the exercise.

**Don't rate `medium` as a resting place.** In the current code `medium` is overwhelmingly the fallthrough, so a real `medium` and an unrated action are indistinguishable by level alone. If an action lands there, it should be because you looked at it and it belongs between `low` and `high`, and `risk_established: true` should then say so.

One caveat on the tests: running the suite currently appends real rows to `logs/security-audit.jsonl` — I found 216 synthetic events sitting in mine. I'm shipping the fix as v2.209.1 shortly (an autouse fixture redirecting the audit path to tmp for the whole suite). Rebase onto that before writing audit assertions, or your new tests will inherit the leak.

No rush on scope — if 80 in one PR is unwieldy, splitting by tool is fine and easier to review.

samuelgursky:
@Rohitkanithi — apologies, I need to withdraw the follow-up I offered you a few hours ago. It got picked up and landed here as [v2.210.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.210.0) before you'd have had a chance to start. That's on me: I shouldn't have offered work and then taken it. If you'd already begun, tell me and I'll make it right.

Posting what it found, since it's the substance of your PR and you should see where it ended up.

All 108 registered destructive actions are rated now — 2 critical, 33 high, 35 medium, 38 low. Safe mode gates 35 where it gated 20, so your feature was covering a bit over half of what it was aimed at. Two things came out of reading the handlers rather than the names:

- **`timeline.move_clips` is a deletion wearing a move's name.** It passes `delete_sources=True` to the duplicate helper, so the originals are removed. Rated high.
- **`timeline_item.update_sidecar` is the only registered action that writes outside the project.** It rewrites the `.braw` sidecar or R3D `.RMD` next to the camera original — no Resolve undo reaches it, and it changes how that media reads in every other application. Rated high, and it was the one rating that genuinely surprised me.

Both guard tests you and I discussed are in: no registered destructive action may be unrated, and `inspect_operation` and the gate must agree on all 108.

The other thing worth reporting, because it vindicates your instinct to add audit output. With `destructive.safe_mode` left true in `logs/media-analysis-preferences.json` on my machine, seventeen tests failed with "Safe mode blocked critical-risk action". A red suite caused by a saved setting rather than by the code — and it landed in the middle of the very change that alters safe-mode behaviour, so it looked exactly like a regression I'd introduced. `tests/offline_guard` now redirects that path for the whole run, same as the audit-log redirect in v2.209.1. Your feature was the first thing in this repo with enough teeth to expose that the suite was reading operator state at all.

If you want more in this area, two things are genuinely open and I'll leave them with you rather than picking them up:

1. **The `medium` band deserves a second pass.** I rated 35 actions medium from their handlers, but medium is the least examined level — nothing gates on it, so a wrong one is invisible. Anything in there that should be high is a real gap.
2. **The honest dry-run refusal** described on #187 — an explicit list of destructive actions that lack a real dry run, short-circuited with a refusal that says nothing was simulated, rather than a fabricated success. That one's still unbuilt and it's the more interesting problem.

Thanks again for #185 and #189 both.

Rohitkanithi:
Thanks @samuelgursky for the detailed follow-up, and no worries on the classification work landing separately. The explanation helped a lot, especially the point about keeping `inspect_operation`, safe-mode behavior, and audit output tied to the same risk source.

I’ll take the two open items you mentioned and split them into separate PRs so each one stays easier to review.

I’ve started with the honest dry-run refusal work and opened #190 for that. I’ll handle the remaining item in a separate follow-up PR.

PR #190, opened 2026-09-07
feat(security): refuse unsupported destructive dry runs

Rohitkanithi:
## Summary

This change makes destructive dry-run handling honest for registered marker operations that do not have a verified native dry-run path.

The security wrapper now detects explicit dry-run requests for known destructive marker actions that currently call Resolve mutators directly. Instead of allowing those calls to continue or pretending that a simulation happened, it returns a structured refusal before Resolve state lookup, archive creation, or handler execution.

The refusal response is intentionally explicit:

- success is false
- status is dry_run_unavailable
- dry_run is true
- simulated is false
- executed is false
- the response includes the same static risk assessment used by inspect_operation
- the error code is DRY_RUN_UNAVAILABLE

This preserves the trust boundary around dry_run. A caller asking for a dry run should never receive a fabricated success response for an operation that was not actually simulated.

## Why this belongs in the destructive-operation wrapper

The lifecycle pipeline is kept observational by default. Upstream tests already pin that default lifecycle hooks do not short-circuit and do not answer on behalf of handlers. This change keeps that invariant intact by placing the refusal at the existing destructive-operation security boundary, where policy refusals already live.

That placement also means the refusal happens before any pre-mutation archive or live Resolve/project state lookup. For unsupported dry runs, there is nothing to archive and nothing to execute, so the safest and clearest behavior is to refuse immediately and report the static risk.

## Scope

The unsupported dry-run list is explicit and deliberately narrow. It currently covers marker operations whose handlers go straight to Resolve mutators and do not implement a dry-run branch:

- timeline_markers.add
- timeline_markers.delete_at_frame
- timeline_markers.delete_by_color
- timeline_markers.delete_by_custom_data
- timeline_markers.update_custom_data
- timeline_item_markers.add
- timeline_item_markers.add_flag
- timeline_item_markers.clear_clip_color
- timeline_item_markers.clear_flags
- timeline_item_markers.delete_at_frame
- timeline_item_markers.delete_by_color
- timeline_item_markers.delete_by_custom_data
- timeline_item_markers.set_clip_color
- timeline_item_markers.update_custom_data

The change does not try to infer unsupported dry-run behavior for unknown future actions. Unknown actions continue to pass through to their handler path instead of being blocked by a guessed policy.

## Non-goals

This does not add a fake simulator.

This does not intercept every dry_run request globally.

This does not change actions that already have real dry-run support. For example, a destructive action with an implemented dry-run plan still reaches its own handler and returns its native dry-run response.

## Tests

Added tests covering:

- known unsupported destructive dry-runs refuse before handler execution
- the refusal does not resolve project state or create archive/versioning work
- the response reports simulated=false and executed=false
- supported native dry-runs still reach their handler
- unknown dry-run actions are not refused by default
- unsupported dry-run registry entries must also be registered destructive actions

Verification run locally:

- .venv/bin/python -m unittest tests.test_destructive_hook tests.test_execution_lifecycle
- .venv/bin/python -m unittest discover -s tests
- .venv/bin/python -m py_compile src/utils/destructive_hook.py tests/test_destructive_hook.py src/server.py src/granular/common.py install.py
- git diff --check

Full unittest discovery result: 3303 tests OK, 36 skipped.

## Version

Bumped the project version to 2.211.0 across the runtime, installer, package metadata, README badges, translated README version note, package lockfile, and changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 Release history for the DaVinci Resolve MCP Server. The latest release is summarized in the root README; older entries live here to keep the README focused.
 
+## What's New in v2.211.0 — unsupported destructive dry-runs now refuse honestly
+
+### Added
+
+- **Known destructive marker operations without native dry-run support now refuse instead of pretending to simulate.**
+  When a caller passes `dry_run=true` to registered marker actions that go
+  straight to Resolve mutators, the security wrapper returns
+  `status: dry_run_unavailable`, `simulated: false`, and `executed: false`.
+  The response includes the same static risk assessment used by
+  `inspect_operation`, plus a structured `DRY_RUN_UNAVAILABLE` error explaining
+  that nothing was simulated and nothing was executed.
+
+- **The refusal list is explicit and registry-checked.**
+  Only known destructive actions listed as lacking native dry-run support are
+  refused. Actions with real dry-run paths still reach their handlers, and
+  unknown future actions pass through rather than being blocked by a guessed
+  policy. A test pins that every unsupported dry-run entry is also a registered
+  destructive action, so stale refusal entries cannot accumulate silently.
+
+### Changed
+
+- **Unsupported dry-run refusals are handled inside the destructive-operation security wrapper.**
+  This keeps the lifecycle pipeline observational by default while still
+  refusing unsafe dry-run requests at the existing policy boundary, before
+  Resolve state lookup, archive creation, or handler execution.
+
 ## What's New in v2.210.0 — every destructive action now carries a real risk rating
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 English | [简体中文](README.zh-CN.md)
 
-[![Version](https://img.shields.io/badge/version-2.210.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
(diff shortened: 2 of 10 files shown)


samuelgursky:
Thanks — this found a real hole: `dry_run=true` on a marker action silently ran the real mutation, and the refusal shape you designed (`status: dry_run_unavailable`, `simulated: false`, `executed: false`, the static risk block) is exactly right and is what shipped.

I landed it on main in v2.211.0 with one structural change: the registry is inverted. Mapping every registered destructive action against its handler showed that 102 of 108 ignore the flag, so a 14-entry denylist left 88 actions (delete_track, delete_clips, update_sidecar, every takes/Fusion mutation) still executing on `dry_run=true`. The release uses an allowlist, `NATIVE_DRY_RUN_ACTIONS`, of the six actions that genuinely honour it, refuses everything else in the registry, and pins the list to the handlers with a static drift test that follows the params object into helper calls (that is what excluded `edit_engine.execute_tighten`: it calls a dry_run-aware helper but with a fresh dict). It is also keyed on registry membership rather than `is_destructive()` so the no-archive filters cannot bypass it.

Release: https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.211.0 — credited to you in the changelog and release notes. Closing this in favour of that landing.

Rohitkanithi:
Thanks for landing this and for the detailed notes.

The allowlist approach makes sense, especially after checking the handlers and finding that most destructive actions ignore `dry_run`. That’s a safer direction than trying to keep a small denylist accurate.

I’ll pick up the remaining security follow-up separately and keep it scoped for review.

PR #192, opened 2026-09-07
feat(security): classify graph LUT risk by target scope

Rohitkanithi:
## Summary

This PR continues the security follow-up from the medium-risk review by making raw graph LUT mutations classify by their actual target scope.

The current classifier treated graph.set_lut and graph.apply_arri_cdl_lut as flat MEDIUM risk. After reading the raw graph handler, that rating is too broad in one direction and too weak in another:

- the raw graph tool defaults to the timeline node graph
- it can also target color_group_pre and color_group_post
- those scopes can affect the rendered look of a whole timeline or every clip in a color group
- the same actions can also be explicitly scoped to source="item", where the mutation is limited to one timeline item

This change keeps the item-scoped case at MEDIUM, but promotes timeline/default and color-group targets to HIGH.

## Behavior

graph.set_lut and graph.apply_arri_cdl_lut now classify as:

- HIGH when source is omitted, because the graph tool defaults to the timeline graph
- HIGH when source="timeline"
- HIGH when source="color_group_pre"
- HIGH when source="color_group_post"
- MEDIUM when source="item"

The blast radius follows the same distinction:

- timeline/default source uses timeline
- color-group source uses project
- item source uses item

Because safe mode blocks established HIGH and CRITICAL operations, the broad graph LUT cases are now blocked before handler execution unless the caller explicitly allows the risky operation.

## Why

The medium band should represent a reviewed middle-risk operation, not a resting place for actions that were not promoted.

For these graph actions, the handler behavior matters more than the action name. A LUT write to one item is meaningfully different from a LUT write to the timeline graph or a color-group graph. Treating all of them as MEDIUM under-rates the broad cases; treating all of them as HIGH would over-block the item-scoped case.

This PR keeps that distinction in the classifier instead of creating a new policy surface.

## Tests

Added focused tests for the new classification behavior:

- inspect_operation / classifier reports raw graph LUT writes as HIGH for timeline/default targets
- inspect_operation / classifier reports raw graph LUT writes as HIGH for color-group targets
- inspect_operation / classifier keeps raw graph LUT writes as MEDIUM for source="item"
- safe mode blocks broad graph LUT writes before the handler runs
- safe mode allows item-scoped graph LUT writes

Also added a medium-band review matrix that pins every remaining MEDIUM destructive action as:

- recognised
- destructive
- not confirmation-gated
- carrying the reviewed blast radius

The existing all-actions drift guard still verifies that inspect_operation and the safe-mode gate agree for every registered destructive action.

## Version

Bumped the project version to 2.212.0 across:

- src/server.py
- src/granular/common.py
- install.py
- package.json
- package-lock.json
- README.md
- README.zh-CN.md
- CHANGELOG.md

## Verification

Run locally:

- .venv/bin/python -m unittest tests.test_destructive_hook tests.test_execution_lifecycle
- .venv/bin/python -m unittest discover -s tests
- .venv/bin/python -m py_compile src/utils/execution_lifecycle.py tests/test_destructive_hook.py tests/test_execution_lifecycle.py src/server.py src/granular/common.py install.py
- git diff --check

Results:

- focused security/lifecycle tests: 65 tests OK
- full unittest discovery: 3318 tests OK, 36 skipped
- py_compile: OK
- diff check: OK

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,35 @@
 
 Release history for the DaVinci Resolve MCP Server. The latest release is summarized in the root README; older entries live here to keep the README focused.
 
+## What's New in v2.212.0 — graph LUT risk follows the target scope
+
+### Changed
+
+- **`graph.set_lut` and `graph.apply_arri_cdl_lut` now classify by graph scope.**
+  The raw graph tool defaults to the timeline node graph and can also target
+  color-group pre/post graphs, so those calls can change the rendered look of a
+  whole timeline or every clip in a group and now classify as HIGH risk. When
+  the caller explicitly scopes the same actions to `source="item"`, they remain
+  MEDIUM because the mutation is limited to one timeline item. Safe mode now
+  blocks the broad timeline/group cases unless the caller explicitly allows the
+  risky operation.
+
+### Added
+
+- **Focused regression tests for the medium-band correction.**
+  The tests assert that `inspect_operation` reports both raw graph LUT
+  mutations as HIGH risk for timeline/group targets, MEDIUM for item targets,
+  and that the safe-mode gate follows the same distinction before handler
+  execution. The existing all-actions drift guard still verifies that
+  `inspect_operation` and the gate agree for every registered destructive
+  action.
+
+- **A medium-band review matrix pins the actions left at MEDIUM.**
+  The table covers every remaining MEDIUM-rated destructive action and asserts
+  that each one is recognised, destructive, not confirmation-gated, and carries
+  the reviewed blast radius. That makes MEDIUM an explicit rating rather than a
+  quiet resting place for anything that was not promoted.
+
 ## What's New in v2.211.0 — dry_run on an action that cannot honour it now refuses instead of executing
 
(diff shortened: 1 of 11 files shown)


samuelgursky:
Thanks — the premise is right and I checked it against the handler: `source` defaults to the timeline-level graph, so a plain `set_lut` restyles every clip on the timeline, and MEDIUM/item under-stated it. Your level split and all of your tests shipped unchanged.

Landed on main in v2.212.0 with one generalisation: the same `source` governs every graph mutation, not only the two LUT actions. Before this, `reset_all_grades` on a color-group graph — which wipes the grade of every clip in the group — still reported an item-sized blast radius. The release derives the radius from `source` for all five graph actions (timeline / item / project for a color group), keeps your HIGH-vs-MEDIUM split for the LUT writes, and leaves the other levels as they were. One extra test pins the radius of every graph action across all five source values.

Release: https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.212.0 — credited to you in the changelog and release notes. Closing in favour of that landing.

PR #193, opened 2026-09-08
chore: update project dependencies

federicosada-pixel:
## Why
Keep the Python and Node dependency surfaces current while preserving compatibility with the MCP server's supported runtime paths.

## What changed
- Upgrade the MCP SDK and compatible Node libraries in both package manifests and lockfiles.
- Update `pyaaf2` and the Python MCP SDK lower bound used by the installer.
- Keep `mcp` below 2.0, `js-yaml` on 4.x, and `better-sqlite3` on the Node 20-compatible 11.x line to avoid known breaking API and runtime changes.

## Validation
- Root package smoke and pack checks pass.
- Advanced Node suite passes: 943 passed, 35 skipped.
- Focused Python dependency and installer tests pass.

The full Python suite still requires the project's supported Python environment and installed MCP dependencies; the system Python used locally lacks `anyio` and runs Python 3.9 in some installer tests.

### DIFF
--- a/install.py
+++ b/install.py
@@ -1385,7 +1385,7 @@ def install_dependencies(venv_path, project_dir):
     # downgrade it, and the fix does not depend on install ordering. Lift both
     # together when server.py is ported to the 2.x layout.
     subprocess.run(
-        [str(pip), "install", "-q", "mcp[cli]>=1.29,<2"],
+        [str(pip), "install", "-q", "mcp[cli]>=1.30,<2"],
         check=True, capture_output=True
     )
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,13 +9,13 @@
       "version": "2.212.1",
       "license": "MIT",
       "dependencies": {
-        "@modelcontextprotocol/sdk": "^1.27.1",
-        "adm-zip": "^0.5.16",
-        "fast-xml-parser": "^5.7.3",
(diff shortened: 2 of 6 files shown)


samuelgursky:
Thank you, @federicosada-pixel — merged and released as [v2.212.4](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.212.4), published to npm.

This was more valuable than a routine bump: the root `package.json` had lagged `resolve-advanced/package.json` since July, so root `npm audit` read nine advisories (five high) while the advanced package read one. Your change brings both manifests into agreement.

What I verified before landing: `npm ci` from both lockfiles on Node 20, the advanced Node suite (948 pass / 0 fail), the full Python suite with the `mcp` 1.30.0 wheel shadowing the venv, and — because zod 3 → 4 is a major the advanced server takes directly — a probe of its two single-argument `z.record()` sites on zod 4.5.4 plus a real stdio `tools/list` (same 18 tools before and after). Two additions on landing: `uuid` was removed from the root manifest rather than bumped (nothing imports it), and the venv was moved to the new 1.30 floor with pip-audit clean.

The five advisories that remain in the root are all transitive under the MCP SDK and clear only when the SDK bumps its own deps, so nothing more to do on our side. The CHANGELOG credits you and this PR.

PR #194, opened 2026-09-08
Fix stale update-cache recommendations and clarify beat detection docs

diesdaas:
## Summary
Two separately reviewable commits:

- Reclassify cached update results against the running MCP version. A cached 2.135.0 -> 2.210.0 update recommendation must not remain an update prompt after upgrading to 2.212.1. Covers persisted, in-memory, and throttled cache paths while preserving error/disabled states.
- Correct the English and Chinese README limitations: optional librosa beat detection and beat/bar/phrase cut-point planning already exist. Clarify inferred downbeats and that a cut-point plan is not a finished music edit.

## Validation
- All 10 tests in tests/test_update_check.py passed, including four new regression tests.
- Test execution substituted a preserving temporary-directory context so test artifacts were retained; production code and assertions were unchanged.
- git diff --check passed.

No Resolve project, media, or analysis workflow behavior is changed.


samuelgursky:
Thank you, @diesdaas — both commits landed as-is. Released as [v2.212.3](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.212.3), published to npm.

The cache reclassification fixes a real annoyance (a persisted "update available" outliving the upgrade it recommended), and the README rows were plainly wrong — the optional-extras table twenty lines above already listed `librosa` for beat detection. I checked every claim in the new wording against `beat_detection.py` and the SKILL entry, and the Chinese wording reads as a faithful translation.

One small addition on landing: the update-check tests gained a per-test reset of the module-wide cache, since the new tests seed it and it could otherwise leak into a later module's cached-status read. The CHANGELOG credits you and this PR.

PR #195, opened 2026-09-08
fix: preserve failed verification evidence in operation results

denoise:
Hi! We found a case where the operation envelope could report verification as passed even when readback failed or a contradiction was present. This keeps those failures visible and leaves bulk command counts unverified until there is actual verification evidence.

Starting with this small fix to see whether contributions like this would be useful here. Happy to adjust the approach.

Validation: regression tests, full offline suite, drift guards, CLI/package checks and managed-install handshake. No Resolve API calls changed.

### DIFF
--- a/src/utils/operation_result.py
+++ b/src/utils/operation_result.py
@@ -166,75 +166,57 @@ def extract_verification(raw: Any) -> Dict[str, Any]:
     if not isinstance(raw, dict):
         return unverified
 
-    # An impl that already speaks this shape wins outright.
     existing = raw.get("verification")
-    if isinstance(existing, dict) and existing:
-        merged = dict(existing)
-        merged.setdefault(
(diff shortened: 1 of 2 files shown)


samuelgursky:
Thank you, @denoise — this is exactly the kind of contribution that's useful here. Merged as-is and released as [v2.212.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.212.2), published to npm.

Verified before landing: full offline suite in a scratch worktree, the drift guards, and the advanced Node suite. On landing I added two more regression tests from the bulk-tool side (command counts cannot mask a failed readback; readback evidence is what passes a bulk result) so a future edit can't quietly restore the old behaviour. The CHANGELOG credits you and this PR.

If you have more findings of this shape, they're welcome.

PR #197, opened 2026-09-08
fix(api_truth): correct three entries falsified by Resolve 21.1

billcarroll:
Resolve 21.1 Studio closes gaps this registry still reports to Blackmagic as missing, and moves the shipped scripting docs out from under a path `tests/live_resolve21_validation.py` reads.

## Evidence

Every claim below was measured by attribute probe on a live **Studio 21.1.0.14** — `PyFunctionCall` vs `None`, the same discriminator used elsewhere in `api_truth.py`. **None of the methods was invoked**, because calling them mutates a real project, so each corrected entry claims existence and signature only and states that limit inline along with what would falsify it. Signatures come from the `DaVinciResolveScript.pyi` that 21.1 ships.

## api_truth corrections

- **Native multicam clip creation** — withdrawn. `MediaPool.CreateMulticamClip`, `TimelineItem.FlattenMulticam`, `TimelineItem.PerformMulticamSmartSwitch` and `Timeline.AutoAlignClips` all exist on 21.1. The pre-21.1 statement is kept as the historical record and `submit: "missing"` is dropped, so it stops being reported as a gap.
- **Transition create / copy / clone** — half withdrawn. `TimelineItem.AddTransition(transitionOptions)` exists on 21.1, taking `type`, `category`, `position`, `alignment` and an optional `duration`. Readback is still missing — no accessor for an existing transition's type, alignment or duration, and no clone verb — so the entry stays submittable on that narrower ground rather than being withdrawn wholesale.
- **`GetClipProperty('Transcription')` truncation** — still true on 21.1, and now has a real route around it. `MediaPoolItem.GetTranscription()` returns `{language, segments[{start, end, text, speaker, words[...]}]}` with timecode strings. Measured against an already-transcribed interview clip: 1550 segments, per-word start/end, populated `speaker`, `'(...)'` as Resolve's own silence marker. It reads the **source** clip, which the entry records, so timeline positions still need mapping through `GetStart()` / `GetSourceStartFrame()`.

`docs/reference/api-limitations.md` regenerated via `scripts/gen_api_limitations.py` (42 → 41 missing capabilities).

## Dead doc path

21.1 deletes `Developer/Scripting/README.txt` in favour of `README.md` plus a typed `DaVinciResolveScript.pyi` and a `CHANGELOG.md`. `readme_date()` opened the `.txt` only — and 21.1 wraps the line as `*Last Updated: 31 Aug 2026*`, so it would have missed the date even against the new file. It now tries both names and strips the markdown markers. `AGENTS.md`'s refresh instruction and the version-gate evidence comment in `src/utils/resolve_versions.py` named the same dead file.

## Deliberately not done here

`docs/reference/resolve_scripting_api.txt` is still the 24 Jul 2026 `README.txt`. Copying 21.1's `README.md` over it renumbers every line, and `scripts/audit_api_parity.py`, `tests/test_resolve_bridge.py` and several dozen `resolve_scripting_api.txt line N` comments in `src/` read it. That refresh wants its own change; `AGENTS.md` now says so.

`Timeline.DeleteTrack` is a related open question and is **not** touched here. It resolved to `None` and crashed Resolve on 21.0.4; on 21.1 it resolves to a callable. That falsifies "it does not exist" but says nothing about the crash, and retesting it needs a scratch project.

## Testing

`python -m unittest discover -s tests -p 'test_*.py'` → 3331 tests, 5 errors, 19 skipped. All five errors are pre-existing on `main` (`test_offline_fallback` DRT authoring, `KeyError: 'success'`) and were reproduced on a clean `origin/main` worktree before this change. No new failures.


samuelgursky:
Thank you, @billcarroll — merged and released as [v2.213.1](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.213.1), the same day 21.1 shipped.

One change on landing, about provenance rather than substance: there is no 21.1 build on this machine, so I could not reproduce the probes. The three entries now say so explicitly — *reported by you in #197 from an attribute probe on Studio 21.1.0.14, not reproduced here* — following the ledger's existing measured / reported distinction, and keep your "existence and signature only, none invoked" caveat and what would falsify each. `docs/reference/api-limitations.md` was regenerated from the reworded ledger.

The `readme_date()` fallback and the `AGENTS.md` note about the moved README landed as-is; agreed that the `resolve_scripting_api.txt` refresh wants its own change because of the line anchors. If you have the 21.1 `DaVinciResolveScript.pyi` to hand, a PR that adds it under `docs/reference/` (without touching the `.txt`) would be welcome — it is the first typed surface Blackmagic has shipped, and the version-gate ledger could cite it directly.

PR #199, opened 2026-09-08
media_pool_item: read the full transcript on Resolve 21.1+

billcarroll:
## What

`media_pool_item get_transcription` read the `Transcription` clip property. That property is a preview: it stops at an ellipsis once the transcript is longer than it exposes, and there was no way to get the rest. The action shipped a `truncated` flag instead of the text.

Resolve 21.1 added `MediaPoolItem.GetTranscription(useNestedClipTranscription)`, which returns `{language, segments[{start, end, text, speaker, words[]}]}` and does not truncate. This prefers that method when it is present and keeps the property as the 21.0.x fallback. `source` in the result says which route ran, so a caller can tell the difference without probing the version.

## Two things the docstring now states, because they bite at the call site

- **The segment timecodes are SOURCE timecodes, not timeline positions.** Anything placing them on a timeline has to map through the timeline item's own start.
- **`words` is several times the bulk of the segment text**, so it is stripped unless `include_words` is passed. Measured on a live Studio 21.1.0.14: 1550 segments on one clip, each carrying its own word list.

## Evidence

`MediaPoolItem.GetTranscription` is declared in 21.1's shipped `Developer/Scripting/DaVinciResolveScript.pyi` and listed under 21.1 "Added" in the shipped `CHANGELOG.md` ("Media Transcription - MediaPoolItem.GetTranscription"). Its return was read on a live Studio 21.1.0.14 and matched the declared shape, including a populated `speaker` field and `(...)` as Resolve's own silence marker.

The `api_truth` entry for the truncated property is unchanged and still correct — the property is still a preview on 21.1. This adds the route around it.

## Compatibility

No new action, so the action list is unchanged. The pre-existing keys (`text`, `truncated`, `status`, `has_transcription`) keep their meaning on both routes; `segments`, `language` and `source` are additions.

## Testing

`tests/test_transcript_read.py` covers both routes plus every fallback: method absent, method present but returning no segments, and method raising. Full offline suite run: 3344 tests, 19 skipped, and the same 5 pre-existing `test_offline_fallback` DRT errors that are present on an unmodified `main`.

### DIFF
--- a/docs/SKILL.md
+++ b/docs/SKILL.md
@@ -909,8 +909,12 @@ Key actions: `get_name`, `get_metadata(key?)`, `set_metadata(key, value)`,
 `set_name(name)`, `link_full_resolution_media(path)`,
 `replace_clip_preserve_sub_clip(path)`, `monitor_growing_file`,
 `transcribe_audio(use_speaker_detection?)`, `clear_transcription`,
-`get_transcription` (read back `{text, truncated, status, has_transcription}`;
-`truncated` flags when Resolve's preview cut the text off),
+`get_transcription(include_words?, use_nested_clip_transcription?)` (read back
+`{text, segments, language, source, truncated, status, has_transcription}`; on
+Resolve 21.1+ it uses `MediaPoolItem.GetTranscription`, so `segments` carries
+`{start, end, text, speaker}` in SOURCE timecode and nothing is truncated, and
+on 21.0.x it falls back to the `Transcription` clip property, where `truncated`
+flags a cut-off preview — `source` says which route ran),
 `perform_audio_classification`,
 `analyze_for_intellisearch(identify_faces?, is_better_mode?)`, `analyze_for_slate(marker_color?)`,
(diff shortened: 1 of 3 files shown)


samuelgursky:
Thank you, @billcarroll — merged and released as [v2.213.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.213.2).

Landed as-is. One addition: `MediaPoolItem.GetTranscription` is now in the version ledger (`src/utils/resolve_versions.py`) as a *reported* 21.1 surface citing this PR, so `resolve_control check_version_support` answers for it and `get_version` lists it as unavailable on older builds. The two tests that define "the newest build the ledger knows" moved from 21.0.4.5 to 21.1.0.14 with it.

As with #197, the route itself could not be exercised here (no 21.1 install), so the ledger entry and the CHANGELOG say reported rather than measured. Your fallback tests cover the shapes that matter on this side.

PR #200, opened 2026-09-08
timeline_frame: put a user's mark range back after a frame capture

billcarroll:
## What

`timeline_frame capture` at quality `frame` renders a single frame, which means setting the project render range to `MarkIn == MarkOut == <frame>`. The cleanup then reset the range to the whole timeline, with this reasoning in the code:

> Best-effort, not a restore: without GetRenderSettings there is nothing to restore FROM, so put the mark range back to the whole timeline rather than leaving it pinned to the captured frame.

That is true of the render settings as a whole — `GetRenderSettings` does not exist, and it is still absent in 21.1's shipped `DaVinciResolveScript.pyi`. It is not true of the mark range specifically.

`Timeline.GetMarkInOut()` returns `{video: {in, out}, audio: {in, out}}`, and the stub documents those as *"Record frame relative to timeline start"* — the same space `RenderSettings.MarkIn`/`MarkOut` take, and the same space the existing fallback already writes `GetStartFrame()`/`GetEndFrame()` into. This repo already exposes `get_mark_in_out` / `set_mark_in_out` / `clear_mark_in_out` on the `timeline` tool, so the read costs nothing new.

So the range is read before the capture and put back after.

## Why it matters

Without this, a capture silently discards a range the user had set, and the next `prepare_render_job` inherits the whole timeline instead of their selection. The failure is quiet in both directions: the job reports success, with the correct preset, target dir and custom name, and only its range is wrong.

## What is deliberately unchanged

- **The fallback.** With no marks set there is still nothing to restore, so the range goes back to the whole timeline rather than staying pinned to the captured frame — exactly the old behaviour.
- **The scope of the claim.** This does not pretend to restore the render settings. The comment still says so, and still names the missing `GetRenderSettings`.
- **A half-set range** (an in point and no out) is treated as no range, because restoring it as a range would mean inventing the out point.

## Testing

`tests/test_playhead_frame_capture.py` gains the restore case, the half-set case, and an unreadable-`GetMarkInOut` case; the existing whole-timeline test is kept as the no-marks fallback. Full offline suite run: 3344 tests, 19 skipped, and the same 5 pre-existing `test_offline_fallback` DRT errors that are present on an unmodified `main`.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -14927,6 +14927,19 @@ def _playhead_frame_render(proj, tl, p: Dict[str, Any]):
         original_tc = tl.GetCurrentTimecode()
     except Exception:
         pass
+    # The capture pins the render range to the captured frame, and there is no
+    # GetRenderSettings to read the surrounding settings back from (still absent
+    # in 21.1). The mark range is the exception: Timeline.GetMarkInOut reports it
+    # in the same record-frame space SetRenderSettings takes, so a range the user
+    # set can be captured here and put back below instead of being flattened to
+    # the whole timeline.
+    original_marks = None
+    try:
+        marks = (tl.GetMarkInOut() or {}).get("video") or {}
+        if "in" in marks and "out" in marks:
+            original_marks = marks
+    except Exception:
+        pass
 
     job = None
     try:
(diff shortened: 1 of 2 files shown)


samuelgursky:
Thank you, @billcarroll — merged and released as [v2.213.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.213.2), with one adaptation on landing that is worth knowing about.

The premise that `GetMarkInOut` and `SetRenderSettings.MarkIn/MarkOut` share a frame space did not hold when measured. On Studio 19.1.3.7, on a timeline starting at 86400:

| `SetRenderSettings` | rendered |
|---|---|
| `MarkIn=MarkOut=86420` | timeline frame 20, one frame |
| `MarkIn=MarkOut=20` | timeline frame **0**, one frame, `True` returned, no error |

So `SetRenderSettings` takes absolute record frames and silently clamps anything below the start. `GetMarkInOut`, per Blackmagic's own README example (`{'in': 0, 'out': 134}`) and the 21.1 stub you quoted, reports marks *relative* to the timeline start. Handed back verbatim, a UI-set range would have been restored as a clamped range with every readback agreeing — the quiet failure in the other direction. `SetMarkInOut` itself stores whatever it is given (10 reads back 10, 86410 reads back 86410), so a script-written range can be in either space.

The landed version offsets a mark below the timeline start by `GetStartFrame()` and keeps one at or above it as absolute; your absolute, half-set and unreadable tests are unchanged and a relative-range test sits beside them. The clamp is now an `api_truth` bug entry with the measurement. The restore itself is the right idea and is what shipped.

PR #201, opened 2026-09-08
docs(timeline_frame): capture docstring still says the mark range is not restored

billcarroll:
`_playhead_frame_render`'s docstring lists the mark range with TargetDir and CustomName as things that are "NOT readable on builds without GetRenderSettings, so this resets those to sane values rather than truly restoring them."

Since #200 and v2.213.2 that is no longer true of the mark range: it is read with `Timeline.GetMarkInOut`, offset into `SetRenderSettings`' absolute frame space, and put back. The docstring is the first thing a caller reads to decide whether `quality="frame"` is safe against their render setup, so it pointing the wrong way is worth the six lines.

Rewritten as the three cases that actually happen on the way out — format/codec genuinely restored, mark range restored when one was set (whole timeline as the fallback), TargetDir and CustomName reset because nothing can read them back.

Docstring only. No behaviour change; `ast.parse` clean.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -14866,12 +14866,18 @@ def _playhead_frame_render(proj, tl, p: Dict[str, Any]):
     A single-frame render honours the grade, Fusion and titles, is frame-exact,
     runs in well under a second, and needs no GUI panel or foreground window.
 
-    The cost is that render settings are project-level state. Format and codec
-    are readable and are restored; the rest (TargetDir, CustomName, mark range)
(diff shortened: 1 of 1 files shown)


samuelgursky:
Thank you, @billcarroll — merged as-is and released as [v2.213.3](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.213.3). Agreed on the reasoning: that docstring is the thing a caller reads to decide whether `quality="frame"` is safe against their render setup, so it pointing the wrong way was worth its own change.

PR #202, opened 2026-09-08
feat(security): log mutating operations

Rohitkanithi:
## Summary

This adds a compact structured operation log for recognised mutating MCP tool calls.

The existing execution trace system is useful for understanding a multi-step editing run, and the existing security audit records destructive gate decisions. This change adds a narrower chronological trail for project-changing operations: what operation was attempted, what risk level the classifier assigned, whether it was a dry run, when it happened, how it ended, and a short summary derived from the result envelope when one is available.

The goal is traceability without pretending to solve rollback or simulation yet. The log records facts the server already has at lifecycle time and avoids inventing semantic changes when a handler did not report them.

## What changed

- Added `src/utils/operation_log.py` as a small append-only JSONL writer for mutating operation records.
- Added an `OperationLogHook` to the execution lifecycle so records are written after the operation envelope has normalized status, duration, verification, and changes.
- Added lifecycle error-path logging so mutating operations that raise an unhandled exception still write a failed record with the exception type and message.
- Kept read-only operations out of the operation log so the file stays focused on project-changing activity.
- Added setup defaults for enabling/disabling and relocating the log:
  - `destructive.operation_log`
  - `destructive.operation_log_path`
- Added `RESOLVE_MCP_OPERATION_LOG_FILE` as an environment override for deployments that need to route logs outside the repository.
- Updated test isolation so default operation-log writes during the suite go to a temporary file rather than polluting a real operator log.
- Bumped version surfaces to `2.213.0` and added a changelog entry.

## Record shape

A successful mutating operation writes a record in this shape:

```json
{
  "operation_id": "op_123",
  "tool": "timeline",
  "action": "duplicate_clips",
  "operation": "timeline.duplicate_clips",
  "risk_level": "medium",
  "risk_established": true,
  "blast_radius": "timeline",
  "dry_run": false,
  "timestamp": "2026-09-08T10:30:00Z",
  "summary": "timeline.duplicate_clips success; items_added=1",
  "status": "success",
  "duration_ms": 5,
  "changes": {
    "items_added": 1
  }
}
```

If a mutating operation is blocked before mutation, it is logged as `blocked`. If a mutating handler raises before returning a normal result, it is logged as `failed` with an `exception` object. Read-only actions are skipped.

## Why separate from the existing logs

Execution traces answer the larger question of how a run unfolded across one or more tool calls. The security audit records destructive security decisions such as safe-mode blocks and dry-run refusals. This operation log is intentionally simpler: one compact entry per recognised mutating operation attempt.

Keeping it separate avoids overloading the execution trace format and avoids turning the security audit into a general activity log.

## Validation

- `python3 -m py_compile src/utils/operation_log.py src/utils/execution_lifecycle.py tests/test_operation_log.py`
- `.venv/bin/python -m unittest tests.test_operation_log tests.test_execution_lifecycle tests.test_destructive_hook`
  - `Ran 73 tests in 0.889s` / `OK`
- `.venv/bin/python -m unittest discover -s tests`
  - `Ran 3329 tests in 95.875s` / `OK (skipped=36)`
- `git diff --check`
- `npm run smoke`
  - reports `DaVinci Resolve MCP 2.213.0` and `2.213.0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,40 @@
 
 Release history for the DaVinci Resolve MCP Server. The latest release is summarized in the root README; older entries live here to keep the README focused.
 
+## What's New in v2.215.0 — mutating operations write a structured operation log
+
+### Added
+
+- **Recognised mutating operations now write compact JSONL operation records.**
+  Each record captures the operation id, tool, action, `tool.action` name,
+  risk level, whether the risk was established by the classifier, blast
+  radius, dry-run flag, timestamp, final status, duration when available, and
+  a short human-readable summary. When a result already reports semantic
+  changes through the operation envelope, those changes are copied into the log
+  instead of guessed.
+
+- **The operation log is separate from both execution traces and the security audit log.**
+  Execution traces explain multi-step work, and the security audit records
+  destructive gate decisions. The new log is the simpler chronological trail
+  for mutating tool calls: what was attempted, whether it ran, and what the
+  result said. Read-only operations are deliberately skipped so the file stays
+  focused on project-changing activity. If a mutating handler raises an
+  unhandled exception before it can return a normal result, the lifecycle error
+  path still writes a failed record with the exception type and message.
+
+- **Setup defaults can enable, disable, and relocate the operation log.**
+  `destructive.operation_log` defaults to `true`, and
+  `destructive.operation_log_path` defaults to `logs/operation-log.jsonl`.
+  The path also honors `RESOLVE_MCP_OPERATION_LOG_FILE` for deployments that
+  need to route records outside the repository. Test isolation redirects the
+  default path so synthetic suite activity never pollutes a real operator log.
+
+### Validation
+
+- Added coverage for record shape, lifecycle writes for mutating operations,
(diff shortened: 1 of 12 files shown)


samuelgursky:
Thank you, @Rohitkanithi — merged and released as [v2.215.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.215.0), published to npm.

Everything landed as authored: the JSONL writer, the lifecycle hook (including the `on_error` path), the setup keys and the env override, and the offline-guard redirect — that redirect is what keeps the suite from writing a real `logs/operation-log.jsonl`, which matters because the suite had been polluting `server.log` the same way until v2.214.4. The one thing dropped, as with your earlier PRs, is the bundled version bump and CHANGELOG entry; the release notes credit you and this PR. Reviewed against the merged main: 3378 tests, your seven included, green.

One observation for next time rather than a change: `_dry_run_requested` reads `bool(params["dry_run"])`, which is what the destructive hook already does, so it is consistent — but both treat the string `"false"` as true. If you want a small follow-up, coercing both through the hook's `_coerce_bool` would close that.

PR #204, opened 2026-09-09
fix(granular): recognize lowercase Resolve 21.1 item types

legionsound:
Resolve Studio 21.1.0.14 returns lowercase `video` from `TimelineItem.GetType()`, as documented in its shipped typed API. Granular timeline-item tools compare against `Video`/`Audio`, rejecting ordinary video clips before calling SetProperty and omitting type-specific fields from resource reads.

This change normalizes string type values for internal checks, retains title-case compatibility, and treats missing/non-string type values as unknown. It also checks that the optional GetMediaType accessor is callable: on the probed 21.1 video clip it is not, so the existing resource read otherwise fails even after entering the correct video branch. Returned public `type` values remain unchanged.

Validation:

- Seven regression tests exercise the actual granular handlers. Before the fix they produced nine subtest failures; after the fix they pass. Cases include transform/crop/composite/stabilization/keyframe-mode dispatch, lowercase audio reads, title-case compatibility, optional accessor handling, and refusing writes to non-video/unknown types.
- Read-only live validation on macOS Studio 21.1.0.14: the property reader returned transform, crop and composite sections with no error. The runnable `tests/live_granular_item_types.py` records this without publishing clip/project names or changing project state.
- Focused existing granular tests and static/drift checks pass.
- Full suite: `PYTHONUTF8=1 python -m unittest discover -s tests -t . -p 'test_*.py'` ran 3,388 tests, OK with 27 skipped. In the fresh checkout I installed the locked advanced-server dependencies first. The initial unconfigured run hit missing jszip, ASCII subprocess decoding and test-package/log initialization failures; the configured run above passed without changes outside this PR.

Live mutation validation is now complete for the representative transform path on Studio 21.1.0.14. In a disposable project with a synthetic 640x360 test-pattern clip, the actual granular set_timeline_item_transform handler accepted lowercase video and set ZoomX to 0.5. Native readback returned ZoomX=ZoomY=0.5 with ganged zoom. Resolve's exported frame changed from full-frame pattern to the expected centered half-width/half-height image with black borders; both frames were visually inspected. Other changed guards have stub contracts, not separate live mutation claims.

No tool additions, version bump, or production timeline edits. Existing property names, units and keyframe API behavior are outside this change.

### DIFF
--- a/src/granular/timeline_item.py
+++ b/src/granular/timeline_item.py
@@ -5,6 +5,25 @@ from src.utils.clip_colors import clip_color_refusal
 
 resolve = ResolveProxy()
 
+
+def _item_type(item, method="GetType"):
+    """Normalize type values without assuming an optional API is callable.
+
+    Resolve 21.1 documents lowercase types. Keep title-case compatibility,
+    and treat missing/non-string results as unknown rather than as a clip.
+    """
+    getter = getattr(item, method, None)
+    if not callable(getter):
+        return ""
+    value = getter()
+    return value.lower() if isinstance(value, str) else ""
+
+
+def _has_audio_type(item):
+    return (_item_type(item) == "audio"
+            or _item_type(item, "GetMediaType") == "audio")
+
+
 @mcp.resource("resolve://timeline-item/{timeline_item_id}")
 def get_timeline_item_properties(timeline_item_id: str) -> Dict[str, Any]:
     """Get properties of a specific timeline item by ID.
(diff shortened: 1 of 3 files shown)


samuelgursky:
Thank you, @legionsound — merged as-is and released as [v2.215.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.215.2). The regression tests driving the real handlers and the rendered before/after on synthetic media are exactly the evidence this repo asks for. I checked the compound server for the same class and it has no title-case `GetType()` comparison, so the fix is correctly scoped to the granular layer.

One adjustment on landing: `tests/live_granular_item_types.py` picked its clip with `item.GetMediaPoolItem() is not None`. On Studio 19.1.3.7 that attribute resolves to `None` on some items of a conform timeline (Resolve fabricates missing methods as `None` rather than raising), so the script died before the read under test; it now checks the accessor is callable first. Your 21.1 measurement is recorded as yours in the CHANGELOG — this machine runs 19.1.3.7 and cannot reproduce the lowercase behaviour.

PR #206, opened 2026-09-09
feat: expose twelve Resolve 21.1 read-only controls in both modes

legionsound:
Resolve 21.1 exposes additional discovery data that currently requires custom scripting. This adds twelve read-only actions in compound mode and twelve equivalent granular tools: Studio identification, keyboard and project-settings preset discovery, audio render formats/codecs, normalization modes, clip speed/fades, and timeline/clip output blanking.

Native values are preserved, including false, empty inherited blanking and fractional frame durations. Missing native methods return explicit errors. Audio codec queries validate the format argument; granular clip readers validate their locators. Tool counts, generated agent rules, version floors and coverage documentation are updated together.

Validation on macOS Resolve Studio 21.1.0.14: all twelve readers returned matching domain payloads through both interfaces. The included live script omits private project, clip and preset names. Offline contract tests cover missing methods, argument forwarding, false/empty payloads and invalid locators. Full suite: 3,386 tests passed, 27 skipped. API parity, version gates, documentation counts and generated rules checks pass.

This contribution only adds readers. Setters and other 21.1 families will be separate changes. GetProjectLastModifiedTime is deferred because live calls returned None for existing projects.


samuelgursky:
Thank you, @legionsound — merged as-is and released as [v2.216.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.216.0), published to npm.

The half of this I can measure here is the version gate: on this machine's Studio 19.1.3.7, all twelve compound actions refuse with an explicit "requires DaVinci Resolve 21.1+" error and no exception, which is exactly what `_requires_method` is for. The positive payloads are recorded in the CHANGELOG as your 21.1.0.14 measurement through both interfaces. The regenerated agent rules, the 353 → 365 counts across every surface, the ledger floors and the coverage doc all passed their drift guards on the merged tree, which is the part most contributions get wrong on first landing — you didn't.

Deferring `GetProjectLastModifiedTime` on a `None` readback rather than shipping it was the right call; if you find the condition, an `api_truth` entry with the measurement is where it belongs.

Issue #207, opened 2026-09-09
Track Resolve 21.1 native API and official MCP coverage contributions


legionsound:
## Purpose

Track Resolve 21.1 scripting and official MCP coverage in manageable contributions. I can test on macOS Resolve Studio 21.1.0.14 using synthetic media and disposable projects. This builds on the typed API snapshot requested in #197.

The official MCP and community server can access the same underlying Resolve API. A missing dedicated action is different from an impossible workflow: generic scripting and existing alternatives may already cover it. Proposed native replacements need behavior tests, not just method detection.

## Inventory baseline

The shipped typed API contains 410 class-qualified methods and 46 option dictionaries with 513 fields. Compared with the existing README parser, 57 methods are newly represented and eight old names are absent from the stub. Those figures are documentation differences, not 57 proven new features or eight removed runtime methods. Static source scanning found 50 candidate names without executable references at upstream 91d03a5; dynamic calls and equivalent workflows require manual review. #205 provides the reproducible inventory and preserves the legacy reference.

## Work by release-note family

| Family | Current finding / next contribution |
|---|---|
| Clip type compatibility | #204 fixes lowercase types rejected by granular guards. Live transform write and exported before/after images verified on synthetic media. |
| Typed API and option inventory | #205 adds the shipped stub, changelog, provenance and diagnostic audit. Review option fields as well as methods. |
| Keyboard presets | Enumeration/current-preset readers contributed in #206. Load/import/export/delete remain; preserve global preferences during tests. |
| Project settings presets | Enumeration contributed in #206. Set/save/update/import/export/delete require isolated validation. |
| Render presets | Review UpdateRenderPreset and quick-export enablement, preserving existing presets. |
| Audio render formats/codecs | Both readers contributed in #206 and live compared across compound/granular interfaces. |
| Transcription | Full GetTranscription already landed in compound mode (#199); check granular parity rather than duplicate it. |
| Multicam and alignment | CreateMulticamClip, FlattenMulticam, PerformMulticamSmartSwitch and AutoAlignClips need dedicated coverage and synthetic multi-angle tests. |
| Native audio properties and enabled states | Generic property reads see native fields; review specialized audio helpers against new AudioVolume/AudioPan names and ranges. |
| Speed and fades | Readers contributed in #206. Native setters accepted 50% speed and 12-frame fades with expected readback and restoration on synthetic media. Playback/render validation still outstanding. |
[...]

PR #208, opened 2026-09-09
feat: add native Resolve 21.1 speed and fade setters

legionsound:
Resolve 21.1 has native SetSpeed and SetFades methods, while dedicated community tools currently expose their readers only. This adds compound timeline_item set_speed/set_fades actions and equivalent granular tools, retaining the existing retime interface.

Both accept an options dictionary. Unknown keys, non-finite speed values, non-boolean flags and negative/fractional fade durations are refused before writes. Valid partial options are forwarded unchanged, including zero and false; native failures remain success:false. Both methods have the 21.1 version floor. Granular count changes from 365 to 367, with generated rules, documentation and the read/write report updated.

Contributor-validated on macOS Studio 21.1.0.14 using a disposable project and synthetic 24 fps media, through both interfaces:

- 50% speed readback matched. The Resolve-exported frame at timeline 2s matched the untreated 1s frame pixel-for-pixel and differed from untreated 2s.
- 24-frame fades read back correctly. The first and final exported frames were black; the interior frame matched the untreated image exactly.
- Original speed and fades restored after tests.

The included live fixture script exports these comparison frames. This is sampled video-frame evidence, not full-movie or audio validation. Reverse/freeze, ripple behavior, keyframe stretching, linked audio pitch and other clip types remain unverified. Offline contracts exercise both interfaces, missing methods, failure results, malformed inputs and argument forwarding. Full suite: 3,408 tests passed, 27 skipped. Documentation counts, version floors, UTF-8 discipline and the generated read/write report passed their guards.

Part of #207. Builds on the now-merged #206 readers.


samuelgursky:
Thank you, @legionsound — merged and released as [v2.217.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.217.0), published to npm. The validation layer, the pass-through of partial dictionaries, the honest `success: false` on a native `False`, and the exported-frame evidence with its limits stated are all as they should be. On this machine's 19.1.3.7 both compound actions refuse with the 21.1 floor before any write.

One addition on landing that is worth knowing for the setters still to come: **a new mutating action has to be registered in two tables**, or every gate silently skips it. As contributed, the risk classifier answered `recognised: false, destructive: false` for both actions, which means safe mode, the dry-run refusal, the security audit log and the new operation log would all have treated a clip-speed rewrite — including one with `RippleTimeline: true`, which moves every later clip — as a read. They are now in `src/utils/destructive_hook.py`'s `timeline_item` set beside `set_retime`, and in `_MEDIUM_RISK_ACTIONS` in `src/utils/execution_lifecycle.py` (MEDIUM by that table's own definition: existing content altered, not deleted). A test pins both, and `tests/test_destructive_hook.py`'s "no registered destructive action is unrated" guard is what fails if only one of the two is done. For the preset, blanking and multicam setters on your list: those two entries plus the existing floor gate is the whole shape.

PR #209, opened 2026-09-09
feat: add native Resolve 21.1 transition creation

legionsound:
Resolve 21.1 provides native TimelineItem.AddTransition, but the community server has no dedicated wrapper for it. This adds compound timeline_item add_transition and granular add_timeline_item_transition without replacing the existing offline transition workflow.

The options dictionary requires type, category, position and alignment; duration is optional. Invalid options are refused before writes. Native None/False returns success:false. Successful calls report the returned transition's actual identity and frame span, rather than echoing the requested duration. Documentation warns that inserting a transition changes track item indexes.

The action is registered in both destructive_hook and the MEDIUM-risk execution_lifecycle table, with a 21.1 method floor and destructive granular annotations. Tests pin registration and compound dry-run refusal before native writes. Tool counts (367 to 368), generated rules, documentation and the read/write report are updated.

Contributor-validated on macOS Studio 21.1.0.14 using synthetic red/blue media in a disposable project, through both community interfaces:

- 24-frame centered Cross Dissolve with source handles: actual span 59–83 around cut 71; adjacent clip spans unchanged.
- Both ProRes renders completed with 142 frames. Their complete decoded RGB movies are identical. Boundary samples progress from red through purple to blue; red decreases and blue increases through the blend.
- Zero outgoing/incoming handles: both native calls returned failure, with no transition.

The included live fixture script creates the scratch timelines and renders for repetition. The api_truth update records this as contributor evidence, preserving remaining parameter-readback/clone limitations. Other alignments, automatic duration, audio, Fusion/OFX effects and repeated insertion remain unverified. This is evidence for the tested Cross Dissolve, not every effect accepted by the API.

Full suite: 3,416 tests passed, 27 skipped. Focused transition, write-gate and version-floor checks pass (52 tests); updated API-ledger checks pass (20 tests).

Part of #207. Follows the two-table write-registration guidance from #208.


samuelgursky:
Merged and released as [v2.218.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.218.0), unchanged — no adaptation was needed on this one.

I verified the parts that fail silently rather than loudly, since a missed write registration is invisible until something bypasses a gate:

- **Both write tables, probed live rather than read off the diff.** `classify_operation_risk("timeline_item", "add_transition")` returns MEDIUM / `destructive=True` / `recognised=True`, and `destructive_hook.is_destructive` agrees. This is the check I ask for on every new mutating action, and it is the second PR in a row where you got it right without being asked twice.
- **Full suite here: 3,436 passed, 1 skipped**, on Studio 19.1.3.7 — where the whole family correctly refuses with the 21.1 floor. That refusal is the half I can measure; the rendered evidence stays credited to your 21.1.0.14 measurement in both the ledger and the release notes.
- Agent-rule drift, doc tool counts, api-parity and read/write symmetry all clean with the 367 → 368 bump.

Two notes on the craft, neither of which held anything up:

**The tests are the best part.** `test_actual_span_returned_not_requested_duration` pins the thing that would otherwise rot into a comfortable lie — a wrapper echoing the duration it was handed rather than what Resolve built. And `test_malformed_options_never_call_native` asserting `native.calls == []` proves the refusal happens *before* the write, which is the only version of that claim worth anything.

**One genuinely open question, for whenever you are next in front of 21.1.** `test_optional_duration_is_forwarded_without_inventing_default` deliberately forwards an explicit `"duration": None` verbatim. I agree with the instinct — inventing a default would be the server lying about what it asked for — but omitting the key and passing an explicit null are different payloads at the API boundary, and only the 24-frame case was measured. If `AddTransition` treats a null duration as "use the project default" it is fine as written; if it refuses, then callers passing an explicit null get a failure where they expected automatic duration. Worth a one-line probe next time you have a scratch timeline up. Either answer is useful, and if it refuses I would rather drop the key than paper over it.

The offline `.drp` route stays the render-proven path below 21.1, so this reads as the addition it is. Thanks — and #207 is updated.

legionsound:
Checked the null-duration boundary through the official Resolve MCP, contributor-validated on macOS Studio 21.1.0.14. I used separate fresh timelines with the same synthetic red/blue clips and source handles as the original fixture.

- Duration key omitted: transition created, GetDuration() = 8, span 67–75 around cut 71.
- Explicit duration=None: transition created, GetDuration() = 8, span 67–75 around cut 71.

So explicit null and omission behaved identically for this fixture. No key-dropping workaround is needed on the measured build. This is native creation/readback evidence only; I did not render these automatic-duration cases, and eight frames is the observed result here rather than a universal default claim. Scratch project saved and closed afterward.

samuelgursky:
That settles it — recorded in [v2.218.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.218.2).

The shipped behaviour was right, and now it is right for a measured reason rather than a principled guess: omission and explicit `duration: null` are the same request on 21.1.0.14, both landing an 8-frame transition at 67–75 around the cut at 71. `resolve_control api_truth "AddTransition"` and `docs/reference/resolve211-native-transitions.md` both carry it, with the rule it implies for anyone who touches this later — **do not strip an explicit null to route around a refusal, because there is no refusal**. That is the kind of thing that gets "fixed" by a well-meaning patch two years from now unless the ledger says why it is deliberate.

Two things I kept out of the claim, both from your own framing: the 8 frames is recorded as what that build chose for that fixture rather than a documented default, and these two cases are marked creation-and-readback only, not rendered — unlike the 24-frame fixture behind v2.218.0. Separating "I measured this" from "this is the rule" is the habit that makes your reports usable verbatim, and I have adopted your evidence-tier wording in the ledger entries because of it.

That closes item 3 on the #207 list. The multicam set is now the front of the queue, and it is the other one where a rendered comparison — not a readback — is the bar.

PR #211, opened 2026-09-09
feat: add native multicam creation and flattening

legionsound:
Adds native Resolve 21.1 multicam creation and flattening in both interfaces: media_pool create_multicam_clip / granular create_multicam_clip, and timeline_item flatten_multicam / granular flatten_timeline_item_multicam. The existing stacked-timeline workflow remains available.

Creation resolves every clip ID before writing, rejects missing/duplicate IDs and malformed options, supports all eleven documented MulticamOptions fields, and accepts live Resolve constant names or integral native enum values. Omitted fields remain omitted. It returns the IDs/names actually created, with success:false for an empty native result. Flattening preserves native failure and supports both documented grade constants. Both actions have method/version gates, entries in both write-risk tables, destructive granular annotations, and dry-run refusal tests.

Contributor-validated on macOS Studio 21.1.0.14 with synthetic media in a disposable project: both interfaces created native Multicam items, rendered, flattened with COPY_GRADE, then rendered again. All four complete decoded RGB movies were identical, each 144 frames. Flattened media type became Video and the clip span stayed unchanged. A decoded frame was visually inspected. This proves preservation of the native-selected angle in an ungraded fixture, not angle ordering, alternate-angle selection, grade transfer, audio routing or synchronization.

Full suite passed: 3,423 tests, 27 skipped. Focused wrapper/write/version checks passed (52 tests); updated ledger and wrapper checks passed (27 tests). The reproducible live script, evidence limits, tool-count update (368 to 370), generated rules and api_truth contributor update are included.

Part of #207. Smart Switch and AutoAlignClips remain follow-up work, not implicitly covered here. Native-only timecode alignment has position-readback evidence; the silent-card Smart Switch fixture returned false and is not a meaningful positive speaking-camera test.


samuelgursky:
Merged and released as [v2.219.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.219.0), unchanged. This was the family I said mattered most on #207 — the one where a rendered comparison rather than a readback was the bar — and it cleared it.

Verified here before merging:

- **Both write tables, probed rather than read.** `media_pool.create_multicam_clip` and `timeline_item.flatten_multicam` each classify MEDIUM / destructive / recognised, and `destructive_hook.is_destructive` agrees for both. Third PR running where that arrives correct without a round trip.
- **Full suite green**, and the drift guards clean with the 368 → 370 bump.
- **The return shape is right by documentation as well as by your measurement.** I checked the shipped 21.1 stub: `CreateMulticamClip(clips, multicamOptions) -> list[MediaPoolItem]`. I had flagged to myself that iterating a native result would break if it returned a bare item; the stub and your live run agree it does not.

Two things I want to name because they are the habits that make these easy to land:

**The docstring says "all-or-nothing input resolution (not transactionality)."** That parenthesis is the whole difference between a useful guarantee and an overclaim. Resolving every ID before writing is real and worth having; it is not a transaction, because the native call offers none, and writing that down stops someone later from relying on a rollback that was never there.

**"Omitted fields remain omitted."** Same instinct as the null-duration question on #209 — not inventing a default so the server never misreports what it asked Resolve for. That is now twice you have reached for it unprompted.

The scoping in your evidence section carried into the release notes verbatim: this proves the natively-selected angle survives flattening in an **ungraded** fixture, and angle ordering, alternate-angle selection, grade transfer, audio routing and synchronisation remain unverified. Smart Switch and `AutoAlignClips` are named as not implicitly covered.

#212 is next — it is still marked draft and stacked on this branch, so refresh it against `main` whenever suits and I will pick it up. #207 is updated.

PR #212, opened 2026-09-09
feat: add native timecode and waveform alignment

legionsound:
Adds native Resolve 21.1 AutoAlignClips through compound timeline auto_align_clips and granular auto_align_timeline_clips. Both accept explicit timeline item IDs and the documented SyncUsing/UseTrack options, with named or integral native constants. All IDs resolve before mutation, order is preserved, malformed inputs are refused, and native false stays false. Both risk tables, version floor and destructive annotations are included.

Selection semantics matter: on the measured build, waveform alignment refused video-only selection; audio-only selection aligned audio but left linked video at its old position. Selecting all video/audio items aligned the full camera fixture. The wrapper does not silently expand selection, and its tool documentation explicitly tells callers to include both sides of linked pairs.

Contributor-validated on macOS Studio 21.1.0.14 with synthetic speech/color-card media in a disposable project:

- Timecode: source timecodes differ by one second; starts0/48 moved to0/24.
- Waveform/MIX: identical speech starts0/24 moved to0/0, including video and audio when all were explicitly selected.
- For each mode, both wrappers' complete decoded rendered video and PCM audio matched independently positioned manual reference timelines exactly. Representative rendered frames were visually inspected.

The live fixture, evidence limits, generated rules/read-write report and tool counts (370 to371) are included. Other microphones, drift, variable frame rates, other track selections and long recordings are unverified. Smart Switch remains separate.

Full suite on v2.221.0 integration: 3,445 tests passed, 27 skipped; 58 focused wrapper/write/version checks passed. Part of #207. Refreshed onto current v2.221.0 main (374 to 375 tools); only alignment changes remain in this PR. Post-rebase wrapper, safety, version and documentation checks: 58 passed.


samuelgursky:
**#211 has landed** — merged and released as [v2.219.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.219.0), so the parent this branch is stacked on is now in `main`. Refresh against `main` and mark it ready whenever suits; I have left it as your draft rather than touching it, since you said you would rebase once multicam landed.

Note that `main` has moved a little further than just #211 — v2.218.1 (a Windows 11 process-detection fix, unrelated to 21.1) and v2.218.2 (an `api_truth` note recording your null-duration measurement) are also in. Nothing there should touch your alignment diff, but the tool count is now **370**, so the 370 → 371 bump in this PR will land cleanly as written.

Reading ahead so the review is quick when you flip it to ready — the part I think is the strongest thing in the description, and which I would want kept exactly as it is:

> the wrapper does not silently expand selection, and its tool documentation explicitly tells callers to include both sides of linked pairs

That is the right call. Waveform alignment refusing video-only selection, and audio-only selection moving the audio while leaving linked video behind, is precisely the kind of asymmetry a wrapper is tempted to paper over by quietly adding the linked items — which would then silently move clips the caller never named. Refusing and documenting beats guessing, and a caller who gets an honest refusal fixes their selection in seconds.

One thing I will look for, purely so you can pre-empt it: the same two-table write registration as #208/#209/#211, since `auto_align_clips` moves existing items. And your description already separates the timecode result (position readback) from the waveform result (rendered video *and* PCM audio matched against independently positioned manual references) — please keep those tiers distinct in the ledger entry, because they are not the same strength of claim.

No rush, and no need to bundle anything else into it.

samuelgursky:
Merged and released as [v2.222.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.222.0).

**The selection semantics are the right call and I want the reasoning on record.** Waveform alignment refusing a video-only selection, and an audio-only selection moving audio while leaving linked video behind, is exactly the asymmetry a wrapper is tempted to paper over by quietly adding the linked items. That 'help' would move clips the caller never named — silently, on a timeline they are mid-edit on. An honest refusal costs the caller ten seconds and costs them nothing they cannot see.

**You also kept the two modes at their true evidence tiers**, which I carried into the release notes rather than flattening: timecode is a position result, waveform/MIX was verified against rendered output with complete decoded video *and* PCM matching independently positioned manual references. Those are not the same strength of claim and the ledger should never imply they are.

Verified here: `timeline.auto_align_clips` classifies MEDIUM / destructive / recognised in both write tables; full suite 3,465 passed. Adapted on merge to tool count **375** — generated files regenerated, your code untouched.

legionsound:
Refreshed onto current v2.221.0 main with the count at 374 → 375 and generated rules regenerated. Both write tables remain registered. The integrated full suite passed 3,445 tests (27 skipped); 58 focused integration checks passed.

One evidence clarification: the early native timecode probe was position-readback only, but the final included live fixture now renders **both** timecode and waveform cases through both interfaces against separately positioned manual references. Full decoded video and PCM audio matched for each mode. Selection is still explicit; no linked items are silently added.

I have also switched the contribution queue to one ready-for-review change at a time. #215 is back in draft; later implementations will stay local/draft and be refreshed against main immediately before review, to avoid asking you to keep adapting the counts.

PR #213, opened 2026-09-09
feat: add native timeline and clip output blanking

legionsound:
Adds native Resolve 21.1 timeline/clip output blanking and the explicit clip inheritance switch in both interfaces. Top/Bottom/Left/Right are native pixel coordinates, not independent margin widths. Whole-valued native floats can round-trip without conversion; malformed values and unknown keys are refused. No undocumented geometry limits are invented.

Clip override deliberately does not silently disable timeline inheritance. Contributor-validated on macOS Studio 21.1.0.14: a clip write while inherited returned false without changing inheritance, then succeeded after explicit disable. Native false is preserved. All three actions have method floors, entries in both write-risk tables, destructive granular annotations and dry-run refusal tests.

Actual compound/granular wrappers exported matching synthetic red-clip PNGs with exact lit-pixel bounds: full0/0/640/360; timeline64/36/576/324; clip128/72/512/288; restored inheritance64/36/576/324. The clip override was visually inspected. Native-only partial Top80.0 updated one coordinate while preserving the others in readback. These are sampled-frame/readback results, not whole-movie or out-of-range behavior claims.

Includes focused contracts, a repeatable disposable-project live script, evidence limits, generated rules/read-write report and tool counts 370 to 373. Full suite passed:3,437 tests, 27 skipped after integration onto v2.219.0; 57 focused checks passed. Part of #207. Existing read-only blanking tools remain unchanged.


samuelgursky:
Merged and released as [v2.220.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.220.0). Closing manually because the merge was adapted rather than a clean fast-forward, so GitHub did not link it automatically — the commit is `be39e68` on `main`.

**What I adapted, and why nothing in your work changed.** The branch was cut before #211, so its tool-count bump (368 → 371) and every generated agent-rule file collided with the 370 that multicam had just landed. I resolved the counts to **373** and regenerated the agent-rule files rather than hand-merging them — regeneration is the only resolution that cannot silently disagree with its source. Both documentation pointers were kept. No behaviour was touched.

**The design decision I want to single out:** a clip override does *not* silently disable timeline inheritance. You measured that a clip write while inheriting returns false, and succeeds only after an explicit disable — and you preserved that false instead of flipping the switch for the caller. A wrapper that 'helpfully' disabled inheritance would leave a clip permanently detached from its timeline's blanking, which is a change nobody asked for and nobody would notice until a delivery looked wrong. Refusing and documenting is right.

Also correct: naming Top/Bottom/Left/Right as **native pixel coordinates, not independent margin widths**. The field names invite exactly the wrong mental model, and the tool documentation now heads it off.

The evidence tiering carried into the release notes as you wrote it — exact lit-pixel bounds across full/timeline/clip/restored, plus the partial `Top: 80.0` write preserving the other coordinates, recorded as sampled-frame and readback results rather than whole-movie claims.

One process note, meant helpfully: #211, #213 and #214 each arrived branched from a point before the previous one landed, so each needed the same count adaptation. If you branch from `main` right before opening the next one, they will merge untouched. The current tool count is **374**.

PR #214, opened 2026-09-09
feat: add native audio level normalization

legionsound:
Adds native Resolve 21.1 NormalizeAudioLevel in compound timeline normalize_audio_level and granular normalize_timeline_audio_level. Explicit audio timeline item IDs resolve before writes; duplicate/missing IDs and malformed options are refused. All four NormalizeAudioOptions fields are supported, including named or integral native setLevelMode constants. Native false stays false, and native defaults remain native defaults.

The action has a 21.1 callable-method floor, registration in both write/risk tables, destructive granular annotations and compound dry-run refusal tests. Source audio files are untouched.

Contributor-validated on macOS Studio 21.1.0.14 using synthetic stereo tones in a disposable project. Independent FFmpeg measurements of actual wrapper-exported WAVs:

- Relative peak normalization to -6 dBFS preserved the source 12 dB difference: segments measured -6.0/-18.0 dBFS.
- Independent peak normalization measured -6.0/-6.0 dBFS.
- EBU R128 target -23 LKFS measured -22.9 LUFS, within 0.1 LU of target.
- Complete decoded PCM was identical between both interfaces for every case.

Includes reproducible fixture script, evidence limits, generated rules/read-write report and tool counts 370 to 371. Full suite after rebasing onto v2.219.0: 3,437 tests passed, 27 skipped; 57 focused integration checks passed. Other normalization modes, difficult true-peak limiting, multichannel routing, long programs and different source formats are not claimed verified. Part of #207.


samuelgursky:
Merged and released as [v2.221.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.221.0).

**This is the strongest evidence in the 21.1 series so far**, and it is worth naming why: you measured the *exported audio* with independent FFmpeg analysis rather than reading back what you had just written. −6.0 / −18.0 dBFS preserving the source 12 dB difference under relative peak, −6.0 / −6.0 under independent peak, and −22.9 LUFS against a −23 LKFS target, is a claim about what Resolve actually did to the samples. A readback would have told us only that the option stuck. That distinction is the whole reason this project keeps evidence tiers, and audio is the surface where it matters most — a wrong normalization mode produces a file that plays fine and fails QC.

Verified here: `timeline.normalize_audio_level` classifies MEDIUM / destructive / recognised and is in the destructive registry; full suite 3,461 passed.

Adapted on merge the same way as #213 — the branch was rebased onto v2.219.0, so its 370 → 371 count collided with the 373 that output blanking landed. Resolved to **374**, confirmed by the agent-rule generator, generated files regenerated rather than hand-merged. Your code is untouched.

Two things I kept exactly as you scoped them: that source audio files are not modified (this writes timeline clip levels, not media on disk), and that other normalization modes, difficult true-peak limiting, multichannel routing, long programs and other source formats are **not** claimed.

Same process note as on #213: each of #211, #213 and #214 was branched before its predecessor landed, so each needed the same count adaptation. Branching from `main` immediately before opening the next one avoids it entirely — the count is now **374**.

PR #215, opened 2026-09-09
feat: expose native Resolve DCTL validation

legionsound:
Adds compound dctl validate_native and granular validate_dctl_native, calling Resolve 21.1 ValidateDCTL. The existing dctl validate action remains the static offline checker. Source layout is passed unchanged; native diagnostics are returned verbatim. None means valid, strings mean invalid, and unexpected native result types are errors rather than false success.

The native action is version-gated and explicitly classified read-only; it does not install, encrypt, apply or render a DCTL. Contributor-validated on macOS Studio 21.1.0.14 against the official MCP and both actual community interfaces: multiline identity succeeds, its one-line version reproduces the known missing-return diagnostic, and invalid source returns the missing-entry-function diagnostic. Both wrappers match native results exactly.

Includes three live fixtures, focused contracts for CRLF/Unicode preservation and errors, static-route regression coverage, updated api_truth guidance and generated documentation/counts (370 to 371). Full suite: 3,438 tests passed, 27 skipped. Validation success is not shader render evidence. Part of #207; encryption remains separate work.


samuelgursky:
Merged and released as [v2.223.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.223.0).

This one closes a loop that started in a comment. When you narrowed the `ValidateDCTL` layout sensitivity back on #207, I wrote three rules into the ledger for any future wrapper — pass the native diagnostic through verbatim, never reflow the user's source, keep the multi-line identity fixture as the control. This PR implements all three. A wrapper that tidied the source until validation passed would hide precisely the behaviour that ledger entry exists to warn about, and your CRLF/Unicode contracts prove the source reaching Resolve is byte-for-byte what the caller supplied.

**The detail I'd single out: an unexpected native result type is an error, not a false success.** `None` means valid, so the lazy implementation treats anything non-string as fine — and then a changed API silently reports every shader as valid. Refusing to interpret what you don't recognise is the harder and correct choice.

Correctly annotated read-only, and verified as LOW / non-destructive here — it does not install, encrypt, apply or render. The notes also keep your line that validation is not a rendered shader test.

Full suite 3,470 passed. Adapted on merge to tool count **376**.

PR #216, opened 2026-09-09
Add native Resolve 21.1 DCTL encryption with safe output handling

legionsound:
Adds native Resolve 21.1 DCTL encryption through dctl encrypt_native and granular encrypt_dctl_native. The caller provides an existing .dctl input, a new .dctle output path and optional expiry. Resolve writes to isolated staging; only a verified non-empty output is published. Existing destinations, including dangling symlinks and files appearing during encryption, are never replaced. Cross-volume fallback uses exclusive creation and removes its own partial output on copy failure.

Native filename and expiry boundaries are handled explicitly: Resolve appends .dctle itself, so a fixed staging stem prevents doubled user suffixes. Empty expiry is normalized to null because the isolated native probe returned false for empty string and true for null/omission. Other expiry strings pass to Resolve. The wrapper reports the actual final path, size and hash, preserves source bytes, and does not install or apply the shader.

The action has its 21.1 method floor, a destructive-action hook, LOW write classification (new files only, no overwrite), and dry-run refusal coverage. Staging uses the repository safe-directory helper. Native false or missing output is not reported as success.

Contributor-validated on macOS Studio 21.1.0.14 through both actual interfaces using synthetic identity code: each exported a non-empty file, returned its correct size/hash, preserved source bytes, used owner-only output permissions and refused repeat export without changing the destination. This is encryption/export evidence, not encrypted shader render acceptance or a claim about cipher security. File sizes are observations, not format guarantees.

Validation: full integrated suite passed 3,458 tests (27 skipped), including export/race/fallback/write/version coverage and generated-document checks. Both actual interfaces were rerun successfully after integration on Studio 21.1.0.14. Based on current v2.223.0 main, with count 376 → 377 and generated rules refreshed. Part of #207.


samuelgursky:
Merged and released as [v2.224.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.224.0) — **the first in this series to merge with no adaptation at all.** Branching from current `main` did exactly what I hoped; nothing needed retargeting.

Verified here:

- `dctl.encrypt_native` classifies **LOW** / destructive / recognised and is in the destructive registry. LOW is the right call, not a downgrade: the action can only ever create a new file, so rating it alongside operations that rewrite existing work would make the rating mean less.
- Full suite 3,478 passed; drift guards, api-parity and read/write symmetry clean at 377.

**The output handling is the best part, and two details in it are ones people usually get wrong:**

Using `os.path.lexists` rather than a plain existence test means a **dangling symlink** counts as an occupied destination. A plain `exists()` reports False for a symlink pointing nowhere, and you would then happily write through it — which is both a silent overwrite and a way out of the directory you thought you were in. And `O_EXCL` closes the window between checking and writing, so a file appearing *during* encryption cannot be clobbered either. Together with `0o600` and removing your own partial output on a failed cross-volume copy, that is a genuinely careful export path.

The two native-boundary fixes are the kind of thing only measurement finds: the fixed staging stem so Resolve's own `.dctle` append cannot double a user-supplied suffix, and normalizing an empty expiry to null because the isolated probe showed `""` returning false while null and omission return true. Both are recorded as what you measured rather than as how the API is documented.

Scoping kept verbatim: this is encryption and export evidence, not a claim that an encrypted shader is accepted by a render, and nothing about cipher strength. File sizes recorded as observations, not format guarantees.

Note for whatever you open next — the tool count is now **377** and the version is **2.224.0**.

PR #217, opened 2026-09-09
Trap-aware execution: push verified Resolve API facts at the callsite

Grimthereapper:
## What this changes

`src/utils/api_truth.py` records behaviours of the Resolve scripting API that were measured against a live build rather than read off a signature — calls that return `True` having done nothing, settings keys silently rejected, methods that are not there at all. It is the most valuable thing in this repo and the hardest to reconstruct.

It had two problems:

1. **It was pull-only.** `lookup_api_truth()` answered `resolve_control(action="api_truth")`, so an agent had to know to ask. Exactly one call site pushed proactively — `_setting_limitation()`, for settings keys.
2. **It was thinnest where work actually gets lost.** `ApplyGradeFromStill`, `ExportLUT` and `ArchiveProject` had no entries at all; `CopyGrades` had one.

This closes both.

## Measured, not recalled

Probed live on **DaVinci Resolve Studio 21.1.0.14**, against a disposable project with ffmpeg-synthesised media:

| Symbol | Measured behaviour |
|---|---|
| `TimelineItem.CopyGrades` | **Replaces** the target grade wholesale and creates **no version**. Proven by baking each state to a 33-point LUT: afterwards the target's LUT is byte-identical to the source's (md5 `50c4514e`) while `GetVersionNameList` is unchanged. Returns `True` throughout, so on hand-graded clips this is unrecoverable loss reported as success. |
| `TimelineItem.ApplyGradeFromStill` | **Does not exist** on `TimelineItem` or `Graph` in 21.1, and is absent from the typed stubs. `Graph.ApplyGradeFromDRX` is the real symbol. |
| `TimelineItem.ExportLUT` | Color-page gated: measured `False` from media, edit, fusion, fairlight and deliver; `True` only from color. It fails cleanly — no file is written on the failing pages. |
| `Timeline.DuplicateTimeline` | Silently moves the current-timeline pointer to the duplicate. `timeline_versioning` already compensates; the API fact itself existed only in a code comment. |

The existing `ProjectManager.DeleteProject` entry was **reconfirmed** rather than duplicated — the duplicate-symbol guard caught the collision, and the entry already documented more than this probe did.

## The push mechanism

- **`ACTION_SYMBOLS`** maps a compound `(tool, action)` to the Resolve symbols it actually calls, declared explicitly. Exact symbol equality only. The reason is `_setting_limitation`'s own: *"attaching an unrelated explanation to a failure is worse than attaching none — it reads as a diagnosis."*
- **`traps_for(tool, action)`** resolves the facts; **`trap_notice()`** is a compact three-field push form. The full entry stays one lookup away — response weight is a real cost on long sessions.
- **`destructive_op`** grows a thin outer wrapper: work-destroying traps refuse until `acknowledge_trap: true`, advisory traps ride along as `known_limitation`. `_inner` has a dozen return paths; one outer attach point covers them all and cannot drift as those paths change.

Dry runs are exempt — a preview destroys nothing, and preempting `_inner` swallowed its `DRY_RUN_UNAVAILABLE` refusal, which is the more useful answer. The existing suite caught that; a regression test keeps it honest.

That exemption only sees an **explicit** dry-run request. `bulk_match_to_hero` defaults `dry_run` to `True` inside its handler (`p.get("dry_run", True)`), so a bare first call previews and mutates nothing while `_explicit_dry_run_requested({})` is `False` — it was being refused for a destruction it was never going to perform. The exemption above fixes that instance; a guard covers the principle, failing if any still-refusing action defaults `dry_run` to `True`.

### Which actions refuse

`destroys_prior_work` is a property of the **symbol**, but four actions call `TimelineItem.CopyGrades`, and two of them already make the caller confirm:

| Action | Existing gate | Behaviour |
|---|---|---|
| `copy_grades` | none | **refuses** until `acknowledge_trap` |
| `apply_look_to_items` | none | **refuses** until `acknowledge_trap` |
| `safe_copy_grade` | issues a `confirm_token` whose preview names the exact risk | advisory only |
| `bulk_match_to_hero` | dry-run-by-default, then `confirm_token` | advisory only |

`TRAP_REFUSAL_EXEMPT_ACTIONS` stands the guard down for the latter two rather than teaching it to detect the other mechanism — two mechanisms enforcing one rule drift apart, and the confirm-token flow is older and more specific. They keep the advisory `known_limitation`; the fact is worth having, the second refusal is not.

Without that, a caller would pay two acknowledgements found serially — add `acknowledge_trap`, retry, then discover a `confirm_token` is also needed — and it landed hardest on `safe_copy_grade`, whose name promises it is the careful route. Making that the most irritating call pushes people toward the raw `copy_grades` the guard exists to protect them from.

One detail worth recording for future readers: the confirmation for both exempt actions lives **inside the handler**, not in `_TOKEN_GATED_DESTRUCTIVE_ACTIONS`, so `_PENDING_CONFIRM_CHECK` returns `False` for them and the wrapper's short-circuit never fired. The user-visible flow is the two-step; the mechanism is just elsewhere. That is why the guard ties the exemption to the handler source rather than to that registry.

`RESOLVE_MCP_DISABLE_TRAP_GUARD=1` turns the whole thing off.

## Keeping the facts alive

A fact nobody can re-measure decays into folklore the moment Blackmagic ships a build — and one of these now refuses calls, so a stale entry would block legitimate work rather than merely mislead. `color_grade_live_probe` re-derives three of them and records `drifted` when Resolve stops agreeing.

Every setup return in that probe is checked. `SetCDL` and `ExportLUT` both report refusal as a bare `False`, so a probe that graded nothing and exported nothing would compare two empty LUTs and conclude "no replacement" with total confidence — and that gets written down as a fact. The two `ResetAllGrades` calls are recorded but not gated on: the stubs type them `-> bool`, but "nothing to reset" plausibly returns `False` on a clean graph, and gating would abort every run.

## Guards added

- every mapped symbol is a real `API_TRUTH` entry
- every mapped action is a real handler (same AST technique as `test_destructive_registry_drift`)
- every `destroys_prior_work` entry is reachable from some action, or the refusal it powers can never fire
- every `destroys_prior_work` entry is named by a live probe, or it becomes a superstition
- an exempt action's handler still mentions `confirm_token`, so deleting that gate fails loudly instead of silently becoming no confirmation at all
- no still-refusing action defaults `dry_run` to `True`

The handler lookup behind the last two falls back to the inline branch body: it returned `None` for `copy_grades`, which dispatches inline rather than to a helper, and the guard was silently skipping the one action that matters most.

## Reviewer notes

- **Behaviour change:** a call that previously returned `{"success": true}` from `copy_grades` may now refuse pending acknowledgement. Documented in README and `docs/SKILL.md` with the opt-out env var.
- **`VERIFIED_ON` was deliberately left alone.** The module constant claims 21.0.2 while entries beneath it cite 21.0.4.5 and 21.1.0.14, so it is already inaccurate. This adds a per-entry `verified_on` key instead; globally bumping it would assert 113 re-measurements that did not happen. Making the constant a queryable floor is a clean follow-up.
- **Deliberately out of scope:** structural changes to the 32k-line `src/server.py`, deleting any tool, and exposing traps over the `knowledge` tool. The last is a good follow-on, but shipping it before the database is trustworthy only exports gaps faster.

## Testing

Full offline suite: **3458 passed, 32 skipped, 1094 subtests**. Live probing was done on a disposable project that is created and deleted by the probe; no real project or source media is touched.


samuelgursky:
Thank you for this — it is an unusually well-argued PR, and the central diagnosis is right. `api_truth` being pull-only has bothered me for a while: a fact an agent has to know to ask for is a fact that does not fire when it matters. Pushing at the callsite is the correct shape.

**The best thing here is not the guard — it is the drift re-verification.** Adding `_verify_copygrades_replaces_wholesale`, `_verify_exportlut_page_gate` and `_verify_duplicatetimeline_moves_pointer` to the live probe, each recording `drifted` when reality stops matching what the ledger claims, is exactly right. Your comment says it better than I would: *a fact nobody can re-measure decays into folklore the moment Blackmagic ships a build.* That mechanism is worth more over five years than the entries it currently checks. The `CopyGrades` proof — baking each state to a 33-point LUT and comparing digests — is real evidence rather than an inference from a return value, and reconfirming the existing `DeleteProject` entry instead of duplicating it is the right instinct.

I have run the suite (3,477 passed here) and probed the guard directly. Three things need resolving before I merge, and two of them are the same root cause.

### 1. Two of the four blocked actions already implement the acknowledgement you are asking for

The guard refuses on `destroys_prior_work`, which currently covers `copy_grades`, `safe_copy_grade`, `bulk_match_to_hero` and `apply_look_to_items`.

`safe_copy_grade` already has a confirm-token gate. It issues a preview naming the precise risk — *"Replaces the entire node graph on every successfully resolved target item"* — then requires the token back. That **is** "make the caller say out loud that they know what it does"; it predates this PR.

Because the guard sits in the outer wrapper, it refuses *before* `_inner` ever issues that preview. So the sequence for a caller is: get refused, add `acknowledge_trap=true`, retry, and only then discover they also need a confirm token. Two acknowledgements for one operation, found serially. `safe_copy_grade` is also the action whose name promises it is the careful path — making it the most annoying one to use pushes people toward the raw `copy_grades` you are trying to protect them from.

### 2. `bulk_match_to_hero` gets refused on a call that mutates nothing

`bulk_match_to_hero` is dry-run-by-default — its contract is *"first call (no confirm_token, dry_run=true by default) returns {proposals, blocked, confirm_token}"*. But that default lives inside the handler, not in `DRY_RUN_DEFAULT_TRUE_ACTIONS`, and your exemption only fires on an **explicit** dry run. I checked:

```
_explicit_dry_run_requested({})            -> False
_explicit_dry_run_requested({'dry_run':1}) -> True
```

So a first call with no params — today a safe proposal preview that writes nothing — now returns a refusal telling the caller to acknowledge a trap they were never going to spring. Your reasoning for exempting dry runs is right; it just needs to catch handler-level defaults too.

**For both of these, either drop `destroys_prior_work` from the actions that already gate, or teach the guard to recognise an existing confirmation path.** I lean toward the former: two mechanisms enforcing one rule will drift apart, and the confirm-token flow is the older and more specific of the two. `copy_grades` and `apply_look_to_items` are the ones with no existing gate, and those are exactly where a refusal earns its keep — I have no objection to those two refusing.

### 3. The design document should not be committed

`docs/superpowers/specs/2026-09-09-trap-aware-execution-design.md` is an in-flight design doc ("Status: Approved design, pending implementation plan"). `.gitignore:106` carries the policy: *"gameplans and in-flight design docs are never committed; they live in local/."* The filename slips past the `*gameplan*.md` glob, but it is the same class of file — it goes stale the moment the implementation diverges from it, and then it is a confident, wrong description of the code sitting next to the code. Please drop it from the PR. The reasoning in it is good; it just belongs in the PR description, which is where it will be read anyway.

One smaller note: the README and SKILL documentation of `RESOLVE_MCP_DISABLE_TRAP_GUARD` is the right call, and I would keep the env kill switch regardless of how the above lands. A behaviour change that turns a working call into a refusal needs an escape hatch for whoever is mid-pipeline when they upgrade.

Nothing here is a disagreement with the design — sort those three and I will merge it and cut a release. The refusal-over-warning judgement for genuinely unrecoverable operations is one I agree with.

Grimthereapper:
All three sorted in 995d9db. I verified each claim against the source before changing anything, and one of them needs a correction to the framing.

### 1 & 2 — the guard stands down where an action already confirms

`TRAP_REFUSAL_EXEMPT_ACTIONS` takes your preferred option, for your reason: two mechanisms enforcing one rule drift apart, and the confirm-token flow is older and more specific. The exempt actions keep the advisory `known_limitation` — the fact is worth having, the second refusal is not.

| Action | Own gate | Behaviour now |
|---|---|---|
| `copy_grades` | none | refuses; handler never runs |
| `apply_look_to_items` | none | refuses; handler never runs |
| `safe_copy_grade` | confirm token | runs, advisory only |
| `bulk_match_to_hero` | dry-run default + token | runs, advisory only |

Confirmed at the source: `_safe_copy_grade:17-25` issues the token with that preview, and `_bulk_match_to_hero:101` is `dry_run = bool(p.get("dry_run", True))` — so a bare first call really was being refused for a mutation it was never going to perform.

**The correction:** neither action is in `_TOKEN_GATED_DESTRUCTIVE_ACTIONS`. Their confirmation lives *inside the handler*, so `_PENDING_CONFIRM_CHECK` returns `False` for both and the wrapper's pending-confirm short-circuit never fired. The user-visible two-step is exactly as you describe; the mechanism is just elsewhere. That changed how I wrote the drift guard — tying the exemption to that registry would have passed vacuously, so it asserts against the **handler source** instead: an exempt action whose handler stops mentioning `confirm_token` fails the suite, rather than silently becoming no confirmation at all.

Your underlying point in (2) is covered beyond the one instance: a second guard fails if any *still-refusing* action defaults `dry_run` to `True`, since `_explicit_dry_run_requested` will never see a handler-level default.

Writing that guard turned up a bug in itself worth mentioning: the handler lookup returned `None` for `copy_grades`, which dispatches inline rather than to a helper — so the check was silently skipping the single most important action. It now falls back to the branch body and **fails** rather than skips when it cannot find a handler.

### 3 — design doc dropped

Removed from the tree, and `docs/superpowers/specs/` added to that same `.gitignore` policy block so the next one cannot slip the `*gameplan*.md` glob either. Its reasoning is now in the PR description. Agreed on the diagnosis — a stale design doc next to the code is a confident, wrong description of it.

The env kill switch stays, as you asked.

### One caveat I would rather state than leave for you to find

Both exempt handlers gate behind `_confirm_token_required()`. If an operator turns confirm tokens **off**, those two actions now have no refusal from either mechanism — only the advisory push. That follows from choosing the older mechanism as authoritative, and it is arguably consistent (disabling confirm tokens already opts out of gating on considerably more destructive actions than these). But it is a real hole rather than a theoretical one, and the fix is small if you would prefer the trap guard re-arm when confirm tokens are disabled. Say the word and I will add it.

Suite: 3458 passed, 32 skipped, 1094 subtests. CI green.

One note on history: the commit that added the design doc is still in the branch, with a later commit removing it — so the file never reaches `main` under either merge strategy, but the history is not pristine. Happy to squash it away if you would rather; I left the branch alone by default rather than force-push something under review.

samuelgursky:
Merged and released as [v4.0.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.0.0) — a **major**, because the refusal is a breaking change and deserves to be labelled as one rather than slipped in as a minor. Landed unchanged.

This is the best-reasoned PR this repo has received. Three things in particular.

**The exemption argument is the part I'd have gotten wrong.** My instinct on reading the summary was that `destroys_prior_work` is a property of the symbol, so every action calling it should refuse — uniform, easy to reason about. Your argument is better and it turns on something I'd have missed: two acknowledgements discovered *serially* is a much worse experience than one, and the cost lands hardest on `safe_copy_grade`, whose name is a promise that it's the careful route. Making the careful route the most irritating to call pushes people toward raw `copy_grades`, which inverts the whole point of the guard. "Two mechanisms enforcing one rule drift apart, so the older and more specific one wins" is the right general principle and I want it on the record here.

The detail about the confirmation living inside the handler rather than in `_TOKEN_GATED_DESTRUCTIVE_ACTIONS` — so `_PENDING_CONFIRM_CHECK` returns `False` and the wrapper's short-circuit never fired — is exactly the kind of thing that would have made a future reader "simplify" the exemption into a registry lookup and silently break it. Tying the guard to the handler source instead is right.

**Wrapping `_inner` rather than threading the attach through a dozen return paths.** One outer attach point that cannot drift as those paths change is the correct structural call in a function that size. And catching that the dry-run exemption initially swallowed `DRY_RUN_UNAVAILABLE` — where the refusal is the *more* useful answer because it tells the caller the action cannot be previewed at all — is a genuinely subtle interaction to have found and then written a regression test for.

**"A fact nobody can re-measure decays into folklore."** This is the thing that made me comfortable shipping a refusal at all. A guard powered by a measurement that has quietly gone stale doesn't merely mislead — it blocks legitimate work, which is a worse failure than the one it prevents. That `color_grade_live_probe` exists, that it records `drifted` rather than asserting, and that a test fails if any `destroys_prior_work` entry has no probe naming it, is what makes the refusal defensible rather than a guess with teeth. Checking every setup return inside the probe — so a run that graded nothing and exported nothing can't compare two empty LUTs and confidently write down "no replacement" — closes the loop on that.

Leaving `VERIFIED_ON` alone was also right, and for the reason you gave. Bumping it would have asserted 113 re-measurements that did not happen.

**One thing I changed in how this is presented, and I want to be explicit about it.** The release notes state that your four entries were measured by you on Studio 21.1.0.14 and were **not** re-measured here — this machine has Studio 19.1.3 and free 21.0.3, so I have no 21.1 to check them against. They're recorded as a contributor's measurement with a per-entry `verified_on`, not as a maintainer reconfirmation. That's not doubt about your work; it's the same rule I hold myself to, because a "confirmed on X" claim nobody actually made is how a ledger like this rots. One of the four has independent support: your `ExportLUT` page-gating result matches this repo's own separately recorded finding that grade calls return `False` off the Color page on 19.1.3.7 — arrived at from a different direction, which is the best kind of corroboration.

If you're still on a 21.1 box, running `color_grade_live_probe` once more after this release and pasting the output would let me upgrade that wording from "reported" to "reconfirmed".

**On the follow-ups you named as out of scope** — exposing traps over the `knowledge` tool, and making `VERIFIED_ON` a queryable floor. Both are good, and your reasoning for deferring the first ("shipping it before the database is trustworthy only exports gaps faster") is why I'd take the second one first. Open it whenever you like.

PR #218, opened 2026-09-09
feat(markers): add dry-run preview

Rohitkanithi:
## Summary

This PR adds a native dry-run path for `timeline_markers add` and fixes dry-run boolean parsing in the safety/logging path.

The marker add action is a good first mutation to make previewable because it is bounded, low-risk, and already has centralized payload normalization. With this change, callers can ask what marker would be added without creating a marker, writing a version archive, or producing a misleading simulated success.

### Native marker-add dry run

`timeline_markers add` now accepts `dry_run=true` or `dryRun=true`.

When that flag is present, the tool:

- resolves the marker frame using the same path as a real add
- preserves the existing current-playhead default when frame/timecode is omitted
- normalizes marker color through the existing marker color validator
- applies the same defaults for name, note, duration, and custom data
- returns a structured `would_change` preview
- returns `executed=false`
- does not call Resolve's `AddMarker`

Example response shape:

```json
{
  "success": true,
  "dry_run": true,
  "executed": false,
  "would_change": {
    "operation": "timeline_markers.add",
    "frame": 12,
    "color": "Green",
    "name": "Review",
    "note": "Check audio",
    "duration": 4,
    "custom_data": "marker-12"
  }
}
```

### Dry-run boolean parsing

This also fixes the dry-run parsing issue where string values such as `"false"` were treated as truthy because they were passed through `bool(...)`.

The destructive hook and operation log now share a small boolean coercion helper, so common string values are handled consistently:

- true values: `"true"`, `"1"`, `"yes"`, `"on"`
- false values: `"false"`, `"0"`, `"no"`, `"off"`

That means a caller passing `dry_run="false"` now gets the normal mutation path instead of accidentally requesting a dry run.

### Version-on-mutate behavior

`timeline_markers add` is now registered as a native dry-run action.

For an explicit dry-run preview, the destructive hook treats the payload as plan-only. That avoids creating timeline archives and versioning rows for a request that does not mutate Resolve state. A normal marker add remains a mutating operation and keeps the existing safety/versioning behavior.

Unsupported dry-run requests still refuse before handler execution. For example, a marker action without a native preview path continues to return `DRY_RUN_UNAVAILABLE` instead of pretending the operation was simulated.

### Operation log behavior

The operation log now records dry-run values using the same boolean coercion rules as the destructive hook.

Successful dry-run records summarize as previews, which makes the JSONL log easier to scan:

```text
timeline_markers.add dry-run preview
```

## Why this shape

The implementation keeps the preview honest. It does not synthesize a generic response before validation, and it does not report success for an operation that the real handler might reject.

Instead, the dry-run path sits after marker payload resolution and before the actual `AddMarker` call. That gives callers the same resolved frame/color/name/note/duration/custom data that would be passed to Resolve, while guaranteeing no marker is added.

The shared boolean helper keeps the safety layer and operation log from drifting. Both places now agree on whether the caller actually requested a dry run.

## Tests

Added coverage for:

- `timeline_markers add` with `dry_run=true` returning a preview without adding a marker
- `timeline_markers add` with `dry_run="false"` performing the real marker add
- destructive hook coercion for false-like dry-run strings
- native dry-run payloads skipping archive/version lookup and reaching the handler
- unsupported destructive dry-run requests still refusing before handler execution
- operation log coercion for true-like and false-like dry-run strings
- successful dry-run operation summaries being labelled as previews

Validation run locally:

```text
.venv/bin/python -m unittest tests.test_marker_params tests.test_destructive_hook tests.test_operation_log
Ran 79 tests
OK

.venv/bin/python -m py_compile src/utils/bool_params.py src/utils/destructive_hook.py src/utils/operation_log.py src/server.py tests/test_marker_params.py tests/test_destructive_hook.py tests/test_operation_log.py
OK

git diff --check
OK

npm run smoke
OK
```

I also ran the full unittest suite outside the restricted sandbox. It ran 3,456 tests and had one order-dependent existing failure in `tests.test_log_isolation`; that test passes by itself. No failures were observed in the new focused coverage.

## Version

Bumped release surfaces from `2.223.0` to `2.224.0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,41 @@
 
 Release history for the DaVinci Resolve MCP Server. The latest release is summarized in the root README; older entries live here to keep the README focused.
 
+## What's New in v2.224.0 — marker add dry-run preview
+
+Contributed by @rohitkanithi.
+
+### Added
+
+- **`timeline_markers add` now has a native dry-run path.** When callers pass
+  `dry_run=true` or `dryRun=true`, the tool validates the same marker payload
+  it would use for the real edit, resolves defaults such as the current
+  playhead frame, normalizes marker color/duration/custom data, and returns a
+  `would_change` preview without calling Resolve's `AddMarker`.
+- The preview reports `success=true`, `dry_run=true`, and `executed=false`, so
+  automated clients can distinguish an honest no-mutation preview from a
+  completed marker edit.
+
+### Fixed
+
+- Dry-run parsing now treats common string booleans consistently across the
+  destructive-action hook and operation log. Values such as `"false"`, `"0"`,
+  `"no"`, and `"off"` no longer behave like truthy dry-run requests just
+  because they are non-empty strings.
+- Native dry-run previews are treated as plan-only payloads by the destructive
+  hook, so `timeline_markers add` previews do not create timeline archives or
+  versioning rows before returning the preview.
+
+### Validation
+
+- Added marker tests proving `timeline_markers add` with `dry_run=true` returns
+  the resolved preview without adding a marker, while `dry_run="false"` still
+  performs the real add.
+- Added destructive-hook tests for native dry-run plan-only behavior and string
+  boolean coercion.
+- Added operation-log tests so dry-run string values are recorded correctly and
+  successful dry-run attempts summarize as previews.
+
 ## What's New in v2.223.0 — native Resolve 21.1 DCTL validation
 
 Contributed by @legionsound (#215), live-validated on Studio 21.1.0.14.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 English | [简体中文](README.zh-CN.md)
 
-[![Version](https://img.shields.io/badge/version-2.223.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
+[![Version](https://img.shields.io/badge/version-2.224.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
 [![npm](https://img.shields.io/npm/v/davinci-resolve-mcp.svg?label=npm&color=CB3837)](https://www.npmjs.com/package/davinci-resolve-mcp)
 [![API Coverage](https://img.shields.io/badge/API%20Coverage-100%25-brightgreen.svg)](docs/reference/api-coverage.md)
 [![Tools](https://img.shields.io/badge/MCP%20Tools-36%20(376%20full)-blue.svg)](#server-modes)
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -2,7 +2,7 @@
 
 [English](README.md) | 简体中文
 
-[![Version](https://img.shields.io/badge/version-2.223.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
+[![Version](https://img.shields.io/badge/version-2.224.0-blue.svg)](https://github.com/samuelgursky/davinci-resolve-mcp/releases)
 [![npm](https://img.shields.io/npm/v/davinci-resolve-mcp.svg?label=npm&color=CB3837)](https://www.npmjs.com/package/davinci-resolve-mcp)
 [![API Coverage](https://img.shields.io/badge/API%20Coverage-100%25-brightgreen.svg)](docs/reference/api-coverage.md)
 [![Tools](https://img.shields.io/badge/MCP%20Tools-36%20(376%20full)-blue.svg)](#服务器模式)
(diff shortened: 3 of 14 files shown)


samuelgursky:
The `bool("false") is True` bug is real and worth fixing on its own — thank you for catching it. The marker dry-run is a sensible first previewable mutation for exactly the reason you give: bounded, low risk, and the payload normalization already exists.

I verified the fix against `timeline.ripple_insert`, the one action currently in `DRY_RUN_DEFAULT_TRUE_ACTIONS`:

| `dry_run` | main | this PR |
|---|---|---|
| absent | plan-only | plan-only ✓ |
| `"false"` / `"0"` / `"no"` | plan-only **(the bug)** | mutates ✓ |
| `False` | mutates | mutates ✓ |
| `"maybe"` | plan-only | **mutates** ⚠ |

The first three rows are the fix and they are correct. You also preserved the absent-key default (`dry_run is None -> True`), which is the part I checked most carefully, since losing it would have been a silent safety regression.

### The one thing I want changed

`coerce_bool` returns its `default` for an unrecognised string, and `explicit_bool_param` calls it with the default of `False`. So `dry_run="maybe"` — or any typo, or a value from a client that stringifies oddly — now means "not a dry run", and `ripple_insert` shifts every downstream clip. On `main` that same garbage kept you in plan-only.

The asymmetry is what makes this worth a change rather than a shrug: guessing wrong toward *preview* costs a retry, guessing wrong toward *mutate* costs a timeline. And `DRY_RUN_DEFAULT_TRUE_ACTIONS` exists precisely because those actions should require an opt-in to mutate — an unparseable value silently opting you in inverts that.

The surgical fix is one line: have `explicit_bool_param` return `None` for a string it does not recognise, rather than the default.

```python
def explicit_bool_param(params, *keys):
    for key in keys:
        if key in params:
            value = params[key]
            if isinstance(value, str) and value.strip().lower() not in _TRUE_STRINGS | _FALSE_STRINGS:
                return None          # unreadable is not the same as "false"
            return coerce_bool(value)
    return None
```

That keeps every row in the table above, because `dry_run is None` still falls through to the default-True path for those actions, and the `is True` check for native dry-run actions is unaffected. **"I could not read this" and "you said false" are different answers, and only one of them should be allowed to start a mutation.** Erroring on the unrecognised value instead would also be fine by me, and is arguably better — the caller learns they have a bug. Your call which.

Keeping `coerce_bool(value, default)` as-is for preferences is right; that is a genuinely different context, where falling back to a default is the sane behaviour.

### Two smaller notes

**Version bump.** This PR bumps to 2.224.0, but that version is now taken — #216 landed as [v2.224.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.224.0) while this was open. Please drop the version/CHANGELOG/lockfile changes from the PR entirely and let the release bump happen when it merges; that also removes the conflict you would otherwise hit on `package-lock.json`. Current `main` is 2.224.0 with a tool count of 377.

**The `would_change` preview shape is good** — returning `executed: false` alongside `dry_run: true` is worth keeping. `success: true` on a preview reads slightly oddly to me, since nothing succeeded, but it is consistent with how the other dry-run paths in this repo answer, so I would rather have the consistency than a one-off.

Fix the coercion edge and drop the version bump and I will merge this.

samuelgursky:
Merged and released in [v4.1.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.1.0), adapted onto v4.0.0.

**The boolean fix turned out to be worth more than the feature it came with.**

You found it as a consistency problem between the destructive hook and the operation log. A few hours before this merged, #217 landed as v4.0.0 — a guard that refuses `copy_grades` because `TimelineItem.CopyGrades` replaces a hand grade with no version and returns `True` while doing it. That guard exempts an explicit dry run, correctly, since a preview destroys nothing. It decided "explicit dry run" using `_explicit_dry_run_requested` — the exact function your PR rewrites.

So on v4.0.0, a call carrying `dry_run="false"` read as a dry run and **skipped the destructive-work refusal**. Nothing actually got through: it fell into `lacks_native_dry_run`, which had the identical flaw and refused it with `DRY_RUN_UNAVAILABLE`. But the guard was being answered by a second bug rather than by its own logic, and that is not a state to leave a guard in. Your PR closes it. Two PRs written independently, neither aware of the other, and the second hardens the first.

**On the feature itself.** Putting the preview after payload resolution and before `AddMarker` is the whole design, and your write-up says why better than I would: it keeps the preview honest. It does not synthesize a response before validation, so a payload the real handler would reject is rejected in the preview too, instead of previewing a success that could never happen. That is the difference between a dry run and a plausible-looking lie, and it is the reason `DRY_RUN_UNAVAILABLE` exists in this repo at all — a simulated success is worse than an admission that the action cannot be previewed.

Choosing `timeline_markers add` as the first native preview was also a good read. Bounded, low-risk, already centrally normalized, and now it is the reference shape for the next one.

**Two adaptations:**

1. **The version bump and CHANGELOG were dropped** in favour of main's surfaces. The contributor guidance asks for PRs without them precisely because two PRs landing the same day collide — and that is exactly what happened here: your `2.224.0` met a tree that is now `4.1.0`, and six of the seven merge conflicts were nothing but version surfaces.
2. **The two real conflicts were additive**, not competing: `destructive_hook.py` needed both the `api_truth` import from #217 and your `bool_params` import. Kept both.

Full offline suite after the merge: 3573 → 3579, accounting exactly for your six tests, no new failures.

The `test_log_isolation` order-dependence you hit is real and known here — it is why `RESOLVE_MCP_LOG_FILE` exists and the tests package redirects the suite's log. If it bit you under `unittest discover`, it is worth a separate issue with the runner you used.

Thank you — the shared helper is the kind of change that looks small in the diff and removes a whole class of disagreement between two subsystems.

Rohitkanithi:
Apologies @samuelgursky, I missed your earlier review comment and didn’t make the requested follow-up changes myself.
Thank you for pushing the fix during merge and for explaining the dry-run coercion edge case so clearly. The distinction between “unreadable” and “false” makes sense, especially where the safer default is preview-only.
I also understand the version/changelog guidance now and will leave those out of future PRs so they don’t conflict with the release process.
Really appreciate the merge and the detailed notes.

PR #221, opened 2026-09-10
fix(offline_fallback): give the authoring bridge's import() a file URL

Dev-next-gen:
I ran the offline test suite on Windows (Python 3.12, Node 25) and every `TestAuthoring` case in `tests/test_offline_fallback.py` errored with the same message:

```
Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

It comes from `scripts/author_interchange.mjs`, which hands `path.join(here, '..', 'resolve-advanced', 'server', 'author-interchange.mjs')` straight to `import()`. On Windows that is `C:\...`, and Node reads the drive letter as a URL scheme. So `timeline(action="author_offline")` fails for every target on Windows before anything is written. Because the import runs before target validation, the bridge's own error for an unknown target never shows up either (`TestBridgeScript.test_unknown_target_is_an_error` fails there too). The launcher hit exactly this in 06d5bd6 and got `pathToFileURL()`; the bridge was added later and repeated it. CI only runs on ubuntu-latest, so it stayed green.

The fix is the same as 06d5bd6: wrap the path in `pathToFileURL(...).href`. Nothing changes on macOS or Linux.

Since no Linux run can see this, I added a small static test that fails if a dynamic `import()` in `scripts/*.mjs` or `bin/*.mjs` is given anything other than a string literal or `pathToFileURL(...)`. On main it reports `scripts/author_interchange.mjs:45`, and it passes with the change.

What I ran:

- Windows, on main: the otio/edl authoring tests, `TestBridgeScript` and the new test give 2 failures and 4 errors. With the change, all 8 pass.
- Linux (Node 22): the same authoring and bridge tests pass on main and with the change.

One thing I could not check locally: the `.drt` cases need `jszip` from the `resolve-advanced` npm install, which I didn't have on that Windows machine, so there they now stop at `Cannot find module 'jszip'` instead. They get past the import, which is the only part this touches, and CI installs the dependency.

AI tools used

### DIFF
--- a/scripts/author_interchange.mjs
+++ b/scripts/author_interchange.mjs
@@ -17,7 +17,7 @@
 
 import fs from 'node:fs/promises';
 import path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 
 const here = path.dirname(fileURLToPath(import.meta.url));
 
@@ -42,8 +42,10 @@ async function main() {
   }
   if (!outputPath) throw new Error('outputPath is required');
 
+  // A file URL, not a path: on Windows the ESM loader reads `C:\...` as a URL with
+  // scheme `c:` and refuses it (ERR_UNSUPPORTED_ESM_URL_SCHEME). Same fix as 06d5bd6.
   const module = await import(
-    path.join(here, '..', 'resolve-advanced', 'server', 'author-interchange.mjs')
+    pathToFileURL(path.join(here, '..', 'resolve-advanced', 'server', 'author-interchange.mjs')).href
   );
   const authored = await module.authorInterchange(events, target, opts);
 
--- a/tests/test_offline_fallback.py
+++ b/tests/test_offline_fallback.py
@@ -247,6 +247,30 @@ class TestBridgeScript(unittest.TestCase):
         self.assertIn("aaf", result["error"])
 
 
+class TestDynamicImportsTakeFileUrls(unittest.TestCase):
+    """A path handed to import() works on macOS and Linux and fails on Windows.
+
+    Node reads `C:\\...` as a URL whose scheme is `c:` and throws
+    ERR_UNSUPPORTED_ESM_URL_SCHEME, so every offline authoring call errored on Windows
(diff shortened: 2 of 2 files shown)


samuelgursky:
Merged and released as [v2.224.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v2.224.2), unchanged. Thank you — this is a precise report for a bug nobody on this project could have seen, and I approved the CI run for your first PR so it merged with a green check on record.

What I verified here before merging, since I have no Windows machine to reproduce the failure itself:

- **The guard earns its place.** I ran your new test against the *unfixed* bridge first: it fails naming exactly `scripts/author_interchange.mjs:45`, and passes once the fix is applied. A regression test that has never been seen to fail is a hope, not a guard, so that was the check I cared about most.
- **Your precedent is exactly right.** 06d5bd6 (2026-07-15) applied the identical `pathToFileURL(...).href` fix to the launcher. Same bug, shipped twice, invisible to a Linux-only CI both times — which is the whole argument for checking it statically.
- **The fix is complete, not a first instance.** A repo-wide scan for non-literal dynamic imports — including `resolve-advanced/server/`, which your guard does not scan — found only the two launcher imports already fixed in 06d5bd6 and the one you fixed. Nothing else carries it.
- Full suite with your change: all green.

One optional follow-up, not a blocker: the guard globs `scripts/*.mjs` and `bin/*.mjs`, which today covers every non-literal import site in the repo. A future bare-path `import()` added under `resolve-advanced/server/` would slip past it. Widening the glob would be cheap if you feel like a second PR, but there is nothing there to catch right now.

The release notes credit the Windows measurement to you — two failures and four errors on main, all eight passing with the change on Python 3.12 / Node 25 — and state plainly that this project did not verify it on Windows hardware. They also carry your note that the `.drt` cases need `jszip` from the `resolve-advanced` install and now stop there rather than at the import.

Your note that the import running *before* target validation hid the bridge's own unknown-target error is the detail I would not have spotted. It meant the Windows failure did not even look like the bug it was.

PR #223, opened 2026-09-11
Add gated LUT file discovery, install and removal

legionsound:
Adds LUT file discovery, installation and removal — the gap against Blackmagic's own `list_luts` / `generate_lut` / `delete_lut` — gated the same way v3.0.0 gates the plugin-folder writes it sits beside. Branched from `main` at v3.0.1 (`15680d5`) immediately before opening, as you asked.

`graph set_lut` could put a LUT on a node, but nothing answered the question it raises: which LUTs exist? No listing, no install, no removal — while the same needs for DCTL shaders, in the same directory tree, were fully served.

## What it adds

Compound `lut` — `path`, `list`, `read`, `install`, `remove`, `attenuate`, `capabilities` — and seven granular twins in `graph.py`. Counts **36/377 → 37/384**, docs and agent rules regenerated. No version bump or CHANGELOG.

- **Reads roam, writes do not.** `list`/`read` walk the whole master LUT root; `install`/`remove`/`attenuate` touch only `MCP/`. Stock and vendor LUTs are never modified or removed.
- **Master root, not the user LUT dir**, because `Graph.SetLUT` resolves only against the master root (PR #90's measurement, and now re-proved below). Every listing reports `set_lut_path` in the exact form `set_lut` takes.
- **`generate_from_code: false`, with the reason**, kept exactly as you asked.

## Gates, per your review bar

| Action | Registry | Risk | Non-timeline | Native dry run |
| --- | --- | --- | --- | --- |
| `install` | yes | MEDIUM | yes | yes |
| `attenuate` | yes | MEDIUM | yes | yes |
| `remove` | yes | HIGH | yes | yes |
| `path` / `list` / `read` / `capabilities` | — | read | — | — |

- `@_destructive_op("lut")` on the tool; `lut` in `NON_TIMELINE_WRITE_TOOLS`, so writes are audited but never archive a timeline or reach the version provider.
- `attenuate` is MEDIUM rather than LOW even though it refuses an existing destination, because it writes into the folder Resolve loads LUTs from — the same reasoning as the plugin installs, rather than `encrypt_native`'s user-chosen output.
- All three writes honour `dry_run` themselves and write nothing, so they're in `NATIVE_DRY_RUN_ACTIONS`; the static scan in `test_destructive_hook` agrees.
- `tests/test_lut_write_gates.py` follows `test_plugin_write_gates.py`: recognised/destructive/rated for each write, reads not registered, archiving skipped with `not_a_timeline_mutation` and still audited, safe mode blocks `remove` but not installs, and dry runs reach the handler without writing. The ratchet passes with nothing added to the backlog.

## The claim the tool rests on — measured

You asked for `install` → `set_lut` → `get_lut` on 21.1, plus a rendered frame. Studio 21.1.0.14, disposable project, solid-red clip, constant-green 3D LUT so application is unmistakable. Every row is a real TIFF render, decoded:

| Path | set_lut | Readback | Centre pixel |
| --- | --- | --- | --- |
| Baseline, no LUT | — | — | `[255, 0, 0]` |
| Compound `lut install` → `refresh_luts` → `graph set_lut` → `get_lut` | True, no fallback | `MCP/mcp_apply_probe_compound.cube` | `[0, 255, 0]` |
| Granular `install_lut_file` → `refresh_lut_list` → `graph_set_lut` → `graph_get_lut` | True, no fallback | `MCP/mcp_apply_probe_granular.cube` | `[0, 255, 0]` |

"No fallback" is the important part: compound `set_lut` retries through relocation for LUTs outside the master root, and a success that way would have hidden whether the installed path resolved on its own. It didn't fire in either run. Harness: `tests/live_lut_apply.py`, which refuses to run outside the disposable project.

Native facts from the same session: `SetLUT` works immediately after `RefreshLUTList` (no delay needed), `GetLUT` reads back the exact path, and `SetLUT(1, "")` clears the node.

Two **pre-existing** behaviours I hit and left alone, since they're outside this PR:
- Compound `graph` defaults to `source="timeline"`. The timeline graph had 0 nodes, so my first run's compound `set_lut` returned False until I passed `source="item"`. Documented, arguably worth a clearer default someday.
- Compound `get_lut` answers `{lut}`; granular `graph_get_lut` answers `{lut_path}`.

## Containment, kept as tests

`../escaped.cube` refused with nothing created outside the root; removing a stock LUT refused with the file intact; re-install without `overwrite` refused. Plus the earlier real-tree discovery run: 249 LUTs listed identically by both interfaces, with Resolve closed.

## Not tested

Application of LUT formats other than 3D `.cube`, `layer_index` > 1 and colour-group graphs, `attenuate` against a real vendor LUT, Windows/Linux master roots, and a read-only master root.

`search_api` follows once this lands, branched fresh from whatever `main` is then.


samuelgursky:
Merged and released as [v3.2.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v3.2.0). Thank you — this is the review bar met point for point, and the rendered proof is exactly the evidence I asked for.

**Verified here before merging**, with the PR merged onto current `main` rather than tested on its own branch:

- **The write-enforcement ratchet passes with nothing added to its backlog.** `lut` is the first new tool to land under it. Every write is rated, registered and enforced: `install` and `attenuate` MEDIUM, `remove` HIGH, all three native-dry-run, decorated, and non-timeline. The native-dry-run drift scan agrees with your `NATIVE_DRY_RUN_ACTIONS` entries exactly.
- Every gate and drift guard is clean, and the full suite passes: **3,549 passed, 0 failed**.

**One adaptation on merge.** The granular `remove_lut_file` was labelled `EXTERNAL_WRITE_TOOL`; it's now `EXTERNAL_DESTRUCTIVE_TOOL`. The granular server has no gate decorator, so for it the MCP annotation *is* the signal a client reads — `destructiveHint=false` on a file delete undersold it. A test pins the label. Your compound-side gating was already right; this was only the granular half.

**The parts I'd single out:**

- **"No fallback" in the apply proof.** Compound `set_lut` retries through relocation for LUTs outside the master root, so a success there could have hidden whether the installed path resolved on its own. Checking that it *didn't* fire is what makes the red-to-green pixel mean what it says.
- **"Reads roam, writes do not."** It's the right line for a folder shared with stock and vendor LUTs, and the containment tests hold it there: `../escaped.cube` is refused, a stock LUT can't be removed, and re-installing without `overwrite` is refused.
- **Keeping `generate_from_code: false` with the reason**, so the absence reads as a decision rather than a gap.

**The two pre-existing quirks you flagged are now on our follow-up list**, not yours:

- compound `graph` defaults to `source="timeline"`, which returns False on a node-less timeline graph;
- compound `get_lut` answers `{lut}` while the granular `graph_get_lut` answers `{lut_path}`.

**`search_api` is next, whenever it suits.** Branch it fresh from `main` right before opening it. `main` is now **v3.2.0**, at **37 compound / 384 granular**.

PR #226, opened 2026-09-12
fix(lut): copy a LUT file byte-for-byte on install

Dev-next-gen:
`install_lut(name, source_path=...)` is documented as "a LUT file on disk to copy in", and `LUT_EXTENSIONS` / `lut(action="capabilities")` advertise `.3dl`, `.cube`, `.dat`, `.lut` and `.olut`. But the copy went through a text round-trip:

```python
with open(source_path, "r", encoding="utf-8", errors="strict") as handle:
    payload = handle.read()
...
with open(absolute, "w", encoding="utf-8") as handle:
    handle.write(payload)
```

`.dat` and `.olut` are binary, so pointing `source_path` at one raises `UnicodeDecodeError` before anything is written, and it surfaces to the caller as a `LUT_ERROR`. The same happens to a perfectly ordinary `.cube` that a vendor wrote in latin-1 — the accent in a `TITLE` line is enough:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe8 in position 11: invalid continuation byte
```

I only noticed because the extension list promises two binary formats that the install path can never accept. The text-source branch (`source=`) is not affected and I left it alone; it is only the file-copy branch that needs to be byte-exact.

The fix reads the source as bytes and writes it back as bytes. Two tests cover it: a binary `.olut` copied byte for byte, and a latin-1 `.cube` that comes out identical. Both fail on `main` with the decode error above and pass with the change.

Checks I ran locally on Windows with Python 3.14:

- `python -m unittest tests.test_lut_file_controls` — 30 tests, OK
- the drift guards the CI runs (`test_static_undefined_names`, `test_duplicate_definitions`, `test_action_list_drift`, `test_panel_docs_drift`, `test_doc_tool_counts`, `test_agent_rules_drift`, `test_import`) — OK
- full discovery over `tests/` before and after: identical results either side (35 failures / 62 errors both times, all of them pre-existing here because this box has no numpy, pillow or ffmpeg), plus the two new tests passing

One note on overlap: I have another small PR open, #225, that touches the `overwritten` flag a few lines below in the same function. The two hunks do not overlap and either order merges cleanly, but they are independent and can be taken separately.

AI tools used

### DIFF
--- a/src/utils/lut_files.py
+++ b/src/utils/lut_files.py
@@ -161,15 +161,24 @@ def install_lut(name: str, *, source: Optional[str] = None,
     if source_path is not None:
         if not os.path.isfile(source_path):
             raise LutPathError(f"source_path not found: {source_path}")
-        with open(source_path, "r", encoding="utf-8", errors="strict") as handle:
+        # A copy has to be byte-exact. Two of the extensions this server
+        # advertises -- .dat and .olut -- are binary, and a .cube carries its
+        # TITLE in whatever encoding the vendor wrote it in, so decoding the
+        # source as UTF-8 turned "install the LUT I already have on disk" into
+        # a UnicodeDecodeError surfacing as LUT_ERROR.
+        with open(source_path, "rb") as handle:
             payload = handle.read()
     else:
         payload = source
     if not payload.strip():
         raise LutPathError("refusing to install an empty LUT")
     os.makedirs(os.path.dirname(absolute), exist_ok=True)
-    with open(absolute, "w", encoding="utf-8") as handle:
-        handle.write(payload)
+    if isinstance(payload, bytes):
+        with open(absolute, "wb") as handle:
+            handle.write(payload)
+    else:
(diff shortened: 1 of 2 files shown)


samuelgursky:

The way you found this is the part worth recording: you didn't hit the bug, you noticed that `LUT_EXTENSIONS` and `lut capabilities` advertise two formats the install path can never accept. A promise the code contradicts is a real defect even before anyone trips on it, and that class is much harder to spot than a traceback.

The latin-1 `.cube` is the case that would actually have burned someone. Binary `.olut` at least fails obviously and immediately; a vendor `.cube` with an accent in its `TITLE` looks like every other text LUT until the install refuses it, and the `LUT_ERROR` wrapping a `UnicodeDecodeError` gives no hint that the encoding is the problem.

Leaving the `source=` branch alone was correct — that one is genuinely text, arrives as a `str`, and has nothing to round-trip.

PR #227, opened 2026-09-12
fix(lut): decide LUT writability by path segment, not string prefix

Dev-next-gen:
While reading the LUT file controls from #223 I noticed that `list_luts` decides the `writable` flag with a string prefix test:

```python
is_writable = os.path.realpath(current).startswith(writable_root)
```

`writable_root` is `<master>/MCP`, so any directory whose realpath merely begins with those characters clears the test. A sibling of the writable subdir under the master root does: `MCP_old/` left behind by a hand backup, or a vendor pack that unpacks as `MCPresets/`. Every LUT inside is reported as `writable: true`.

Nothing destructive follows from it, which is why it is easy to miss, but the listing then contradicts the only tool that consumes the flag. `remove_lut` resolves names inside `MCP/`, so feeding back the `set_lut_path` the listing just handed out fails:

```
MCP/MCP_old/Backup.cube is not present in MCP/. This server only removes LUTs
it can install; stock and vendor LUTs are left alone.
```

That is the wrong way round for a flag whose docstring is "whether this server may remove it", and it tells a caller that vendor LUTs are this server's to delete when the module's stated rule is that they are never touched here.

The fix uses `commonpath`, which is the comparison this repo already uses for confinement: `_is_relative_to` in `utils/media_analysis.py` and in `utils/media_analysis_jobs.py`, and `resolve_writable` two functions above in this same file. The other prefix tests in the tree (`media_analysis.py:7768`, `resolve_bridge_ops.py:125`) append `os.sep` for the same reason; `list_luts` was the one place that did not.

No signature, no key and no value type changes. The flag only flips for directories that share the prefix without being under `MCP/`, and for those `remove` already refused, so nothing that worked before stops working. The documented case in `docs/reference/lut-file-controls.md` — an installed `MCP/` LUT appearing in the listing marked writable — still holds, and `_is_within` returns True for the writable root itself.

Proof, with `tests/test_lut_file_controls.py`:

- `test_a_sibling_folder_sharing_the_prefix_is_not_writable` fails on `main` with `AssertionError: True is not false` for `MCP_old/Backup.cube`, and passes with the change.
- `test_writability_matches_what_remove_will_accept` takes the listing at its word and calls `remove_lut` on anything marked writable. On `main` it errors with the `LutPathError` above; with the change the flag is False, so the contradiction cannot arise.
- `tests/test_lut_file_controls.py`, `tests/test_lut_paths.py` and `tests/test_lut_write_gates.py` are green: 47 tests, no failures.

This is my third small fix on this file; it does not overlap the hunks in #225 or #226.

AI tools used

### DIFF
--- a/src/utils/lut_files.py
+++ b/src/utils/lut_files.py
@@ -50,6 +50,20 @@ def writable_dir() -> str:
     return os.path.join(master_lut_dir(), WRITABLE_SUBDIR)
 
 
+def _is_within(path: str, parent: str) -> bool:
+    """Whether ``path`` is ``parent`` itself or sits under it, compared by segment.
+
+    ``commonpath`` is the comparison the rest of the repo uses for confinement
+    checks (`media_analysis._is_relative_to`), and it is the only one that is
+    correct here: a string prefix test also accepts a sibling whose name merely
+    starts with the same characters.
+    """
+    try:
+        return os.path.commonpath([path, parent]) == parent
+    except (ValueError, OSError):
+        return False
+
+
 def normalize_relative(name: str, *, default_ext: Optional[str] = None) -> str:
     """Validate a caller-supplied LUT name and return it as a POSIX relative path.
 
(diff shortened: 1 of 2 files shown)


samuelgursky:

You framed this exactly right: nothing destructive follows from it, and that is precisely why it survived review. The bug is that the listing contradicts the only tool that consumes the flag — `list` hands out a `set_lut_path` marked writable and `remove` then refuses that same path. For a flag documented as "whether this server may remove it", telling a caller that vendor LUTs are this server's to delete is the wrong failure direction, even when the deletion never happens.

`test_writability_matches_what_remove_will_accept` is the test I'd have asked for. It doesn't assert a value, it asserts the two halves agree — so it stays true if the writable root ever moves, and it fails on any future divergence rather than on today's specific paths.

And you were right that the repo already had the answer: `_is_relative_to`, `resolve_writable` two functions above, and the `+ os.sep` prefix tests. `list_luts` was the one place that did it by raw prefix. That's the useful generalisation here — when a confinement check exists in four places and one of them is spelled differently, the odd one out is the bug.

Three good PRs on one file in one afternoon, each independently reasoned and each landing clean. Thank you.

PR #228, opened 2026-09-12
fix(media-analysis): release the timeline-brain DB before deleting an analysis root

Dev-next-gen:
I was looking at what happens when a project's analysis root is thrown away, and `cleanup_artifacts(frames_only=false)` turned out to report success whether or not it actually removed anything.

The root contains `_soul/timeline_brain.sqlite`, and `timeline_brain_db` keeps that connection in a process-wide cache for the life of the server. Nothing lets go of it before the `shutil.rmtree`, and the rmtree runs with `ignore_errors=True`, so the failure is swallowed.

The two ends behave differently, and both are wrong:

- On Windows the open handle makes the DB file undeletable. rmtree stops there, the root stays on disk with `_soul/` and the brain-edit history still in it, and the tool returns `{"success": true, "removed": [root]}`. The user is told the analysis root is gone when it is not.
- On POSIX the root does get removed, but the stale connection stays in `_CONNECTIONS`. The next `connect()` on that project hands it straight back, so the following write goes to a file that no longer has a directory entry. `timeline_brain.sqlite` never reappears and the edit is lost — the dashboard, which opens the path fresh, sees nothing.

What I changed:

- `timeline_brain_db.close(project_root)` drops and closes one cached connection. `close_all()` was the only way to do this and it is too broad here — it would also drop other projects' connections. `reset_for_test` now reuses the new helper instead of duplicating the same pop-and-close.
- `cleanup_artifacts` calls it before the rmtree when `frames_only=False`, and returns `success: False` if the root is still on disk afterwards, rather than reporting a removal that did not happen.

The `frames_only=True` path, which is the one documented in the media-analysis guide, is untouched.

Proof, one test in `tests/test_media_analysis.py`. It logs a brain edit, runs the full cleanup, then checks the root is gone and that a subsequent write lands on disk. It fails on both platforms before the change, each for its own half of the problem:

- Windows: `AssertionError: True is not false : analysis root still on disk after cleanup: ...\project-analysis`
- Linux: `AssertionError: False is not true : write went nowhere: /tmp/.../project-analysis/_soul/timeline_brain.sqlite`

and passes on both after it.

Full offline suite, `python -m unittest discover -s tests -t . -p "test_*.py"`, before and after on each platform:

- Linux: 3529 -> 3530 tests, same 1 failure / 5 errors on both sides (missing optional deps in my environment), and the failing-test list is byte-identical.
- Windows: 3529 -> 3530 tests, same 35 failures / 62 errors on both sides, list also identical.

AI tools used

### DIFF
--- a/src/utils/media_analysis.py
+++ b/src/utils/media_analysis.py
@@ -7057,7 +7057,21 @@ def cleanup_artifacts(project_root: str, *, frames_only: bool = True) -> Dict[st
                     shutil.rmtree(full, ignore_errors=True)
                     removed.append(full)
     else:
+        # The whole root goes, including `_soul/timeline_brain.sqlite`. That DB
+        # is kept open in a process-wide cache, so let go of it first: on
+        # Windows the open handle makes the file undeletable, and rmtree's
+        # ignore_errors=True would hide that and report a root we never removed.
+        from src.utils import timeline_brain_db as _brain_db
+
+        _brain_db.close(root)
         shutil.rmtree(root, ignore_errors=True)
+        if os.path.isdir(root):
+            return {
+                "success": False,
+                "error": f"Could not fully remove the project analysis root: {root}",
+                "removed": removed,
+                "frames_only": frames_only,
+            }
         removed.append(root)
     return {"success": True, "removed": removed, "frames_only": frames_only}
 
--- a/src/utils/timeline_brain_db.py
+++ b/src/utils/timeline_brain_db.py
@@ -250,6 +250,25 @@ def close_all() -> None:
         _CONNECTIONS.clear()
 
 
+def close(project_root: str) -> None:
+    """Drop and close the cached connection for `project_root`, if any.
+
+    Callers that are about to delete or move a project's analysis root need
+    this. On Windows the open handle makes `_soul/timeline_brain.sqlite`
+    undeletable; on POSIX the cache would otherwise hand the next writer a
+    connection to a file that no longer has a directory entry, so the write
+    lands nowhere.
+    """
+    path = db_path_for_project(project_root)
+    with _CONNECTION_LOCK:
+        conn = _CONNECTIONS.pop(path, None)
+    if conn is not None:
+        try:
+            conn.close()
+        except sqlite3.Error:
+            pass
+
+
 @contextmanager
 def transaction(project_root: str) -> Iterator[sqlite3.Connection]:
     """Context manager wrapping a write transaction.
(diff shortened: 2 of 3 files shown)


samuelgursky:
Merged and released in [v3.2.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v3.2.2), with two extensions.

Your analysis was right on both platforms and the fix is the right one. `close(project_root)` rather than `close_all()` is the distinction that matters — a cleanup on one project has no business dropping another's connection — and returning `success: false` when the root survives is the part that turns a silent Windows failure into a reportable one. `ignore_errors=True` over a handle nothing released is a good example of two defensive choices combining into a lie.

**Two things I added on top.**

**1. The same bug at a second site.** A `session_only` run without `keep_artifacts` ingests every report into the brain DB under its output root (`_ingest_report_into_db(output_root, ...)`) and then rmtree's that root with the connection still cached — the identical shape, at `media_analysis.py` in the `if session_only:` branch. It's arguably worse there than in `cleanup_artifacts`, because each session-only run gets a *fresh* temp root, so the cache accumulates one dead connection per run for the life of the process. Both sites now go through one `_drop_brain_db_and_rmtree` helper, and `artifacts_cleaned_up` reports the observed outcome rather than the attempt — the same correction @Dev-next-gen made to `overwritten` in #225, which is a nice coincidence.

**2. Your fix did not work on macOS, and finding out why was the valuable part.** Your regression test fails here *after* the change:

```
AssertionError: False is not true : write went nowhere:
/var/folders/.../project-analysis/_soul/timeline_brain.sqlite
```

The connection cache keys on the caller's spelling of the path. `cleanup_artifacts` normalizes `root` through `realpath` before using it, while the test (and real callers) pass the unresolved path — and on macOS every temp root under `/var/folders` is a symlink to `/private/var/folders`. So `close()` computed a key that was never in the cache, popped nothing, and left exactly the connection it was called to release. Same class as your POSIX half: a silent no-op that looks like it worked.

Two spellings of one root also opened two SQLite connections to one file, which is its own problem. The cache now keys on the resolved DB path, so `connect()` and `close()` cannot disagree, and two tests in `tests/test_timeline_brain_db.py` pin it — both fail without the change.

Worth flagging for your own testing: you validated on Linux and Windows, which is more than most PRs get, and the gap was still the third platform. macOS symlinked temp roots catch a whole family of path-identity bugs that neither of those does.

Thank you — this was the most consequential of your four, and the two-platform write-up is what made the third one findable.

PR #205, opened 2026-09-09
docs: add Resolve 21.1 typed API snapshot and diagnostic inventory

legionsound:
Adds the unmodified Resolve Studio 21.1.0.14 typed API snapshot requested in the review of #197, plus the installed scripting changelog and provenance. The legacy README snapshot and its line anchors remain unchanged.

The new diagnostic inventories all 410 class-qualified methods and 46 TypedDicts (513 option fields), recording signatures and executable source references by compound/granular/helper layer. Comments cannot count as implementation. Matching method names are explicitly marked receiver-unresolved, not covered; the report does not infer types, public action exposure, parameter forwarding or live behavior. At the inspected main it reports 50 candidate gaps. It complements the existing parity guard without claiming a new passing coverage gate.

Validation: six focused regression tests pass; combined parser/static/duplicate-definition/agent-rule checks ran 13 tests, OK with one skipped. Legacy parity audit passes, vendor SHA-256 verified, and git diff --check is clean. No Resolve runtime behavior changes or live project mutations.

PR #229, opened 2026-09-12
Make the shipped typed API searchable from the server

legionsound:
The second of the two you asked for, branched from `main` at v3.2.0 (`c8fbe18`) immediately before opening.

`resolve_control api_truth` answers what is broken. Nothing answered what exists. #205 shipped the vendor `DaVinciResolveScript.pyi`, and `scripts/audit_typed_api.py` can inventory it, but only from a shell, so an agent talking to this server had no way to ask what the native API contains. Your MCP's `search_scripting_api` is the equivalent.

## What it adds

- Compound `resolve_control`: `search_api(pattern, kind?, limit?)`, `describe_api(symbol)`, `api_surface()`.
- Granular twins: `search_resolve_api`, `describe_resolve_api`, `get_resolve_api_surface`, annotated `READ_ONLY_TOOL`. After your note on #223 about the granular annotation being the signal a client reads, I checked this one first.
- All read-only, and none need a Resolve connection. They read the stub that already ships in `docs/reference/`.

Granular count 384 → 387; compound unchanged at 37. Docs and agent rules regenerated. No version bump or CHANGELOG.

## The part that earns its place

Every result carries `referenced_in_this_server` plus the files that reference the method, so a lookup doubles as a parity check: does the native API have it, and do we wrap it? The flag counts **executable syntax only** — attribute access, calls, `getattr(obj, "Name")` — never docstrings or comments. It's built the same way `audit_typed_api.py` builds its source references, for the same reason.

Ambiguous bare names like `GetName` list their candidates instead of guessing. An invalid regex is refused rather than raised. Results are capped, with an explicit `truncated` flag rather than a silent cut.

## The self-consistency assertion, kept as you asked

The parser independently reports **410 methods, 46 TypedDicts, 513 fields**, matching the inventory in `resolve-211-typed-api.md`. `tests/test_typed_api_search.py` asserts that equality, so a future stub refresh that changes the surface fails the suite instead of drifting quietly. The same file checks both interfaces agree on surface, search and describe, and that neither needs a connection.

## Gates

The ratchet passes with nothing added to its backlog: `search_api`, `describe_api` and `api_surface` aren't write-style names, and nothing here writes. Full suite: 3,549 passed, 22 skipped. The 22 skips are environment-specific to my machine; the 21 new tests account for the difference from your v3.2.0 run.

## Scope

It describes the stub checked into this repository, currently 21.1.0.14, not whatever build happens to be installed, and it reports a missing stub rather than guessing. Refreshing the stub stays the documented process in `resolve-211-typed-api.md`.

After this lands I'll wait for you to pick the next of the held branches. Each one gets rebuilt fresh on `main` and registered under the ratchet the way `lut` was.


samuelgursky:
Merged and released in [v3.3.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v3.3.0) — untouched, which is what branching from current `main` immediately before opening buys. Second of the two, exactly as agreed in #207.

`referenced_in_this_server` is the right call and it is the reason this is more than a port of `search_scripting_api`. Restricting it to executable syntax is what makes it trustworthy: a method named in a docstring is the *specific* false positive that would make the flag worse than useless here, given how much prose in this repo discusses API methods it does not wrap. Reusing `audit_typed_api.py`'s technique rather than inventing a second one means the two can't diverge into different answers to the same question.

Verified it answers correctly on a real lookup before release:

```
search_api("SetLUT") -> Graph.SetLUT(nodeIndex: int, lutPath: str) -> bool
  referenced_in_this_server: true
  source_files: ["src/granular/graph.py", "src/server.py"]
```

The 410 / 46 / 513 assertion is the part I most wanted kept, and I want to be explicit about why, since it is easy to read as a brittle magic-number test. It is not asserting that those are the right numbers — it is asserting that two independent derivations of the same surface agree. A stub refresh that legitimately changes the surface *should* fail it, and the fix is to update both sides deliberately. That's the opposite of brittle; it's the only thing standing between a refresh and silent drift.

Checking the granular annotation first, after the #223 note, was the right instinct — that annotation is what a client actually reads to decide whether a tool is safe to call unprompted, and three read-only tools mislabelled would have been a real problem.

**On the held branches.** You asked me to pick the next ones. Two things first:

1. `main` is now **v3.3.0**, at **37 compound / 387 granular**. Everything before this in your queue quotes older counts.
2. Between your last comment and now, four PRs from @Dev-next-gen landed against the `lut` tool you contributed in #223 — three correctness fixes in [v3.2.1](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v3.2.1) and one in v3.2.2. None of them is a criticism of the contribution; they are the kind of thing that only surfaces once code is in the tree and someone else reads it cold. Worth a look, because two of them (`overwritten` reporting a permission rather than an observation, and a confinement check spelled as a string prefix) are patterns worth not repeating in the remaining branches.

I'll pick the next pair in #207 rather than here, so the queue stays in one place. Same rules: branch from current `main` immediately before opening, no version bump, no CHANGELOG.

PR #230, opened 2026-09-13
Add serial source review to the control panel

tpellet:
Reviewing a bin currently requires opening each clip separately. Add an uncropped frame preview with previous/next navigation, verified include/exclude saves, independent star ratings, and selection filtering. Reuse the existing correction store so decisions survive reloads without changing notes.

Previews use analyzed frames; no source-media conversion or Resolve edits.

Validation: offline Python suite, source-media guards, persistence/navigation regressions, static/drift checks, CLI/package smoke checks, and independent review.

### DIFF
--- a/docs/guides/control-panel.md
+++ b/docs/guides/control-panel.md
@@ -101,6 +101,19 @@ detection has run (`detect_entities` + a one-frame-per-cluster confirmation
 in chat), a `Recurring across this bin` card lists the labeled people,
 places, and objects with their shot counts.
 
+Use **Enlarge / review** on a card to inspect its complete analyzed frame without
+cropping. Previous/Next (or arrow keys) follow the visible bin order. **Include**
+(`I`) and **Exclude** (`X`) save the selection and advance after readback;
+**Unreviewed** (`U`) clears the decision without advancing. Ratings save independently
+and do not clear exclusions. The selection filter can show all, non-excluded,
+or excluded clips; it applies to the bin browser, not search results.
+
+Selections are stored as `user.selection` corrections alongside existing notes
+and ratings. They express review preferences, not automatic timeline edits.
+Previews use existing analyzed frames, so their resolution depends on the analysis;
+this does not generate proxies or stream original media. Existing clip detail
+controls remain available for notes and metadata edits.
+
 ### Clip detail
 
 ![Clip detail with shot strip](../images/control-panel/03-clip-detail.png)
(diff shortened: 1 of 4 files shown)


samuelgursky:
Merged and released in [v3.4.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v3.4.0), unchanged. Welcome, and thank you — this is an unusually well-judged first contribution.

Three things I want to call out, because they are the reasons this went in without adaptation.

**The verified save.** Re-reading the clip after each write and refusing to advance unless the value read back matches is the right design for *this* flow specifically, and I don't think that's an accident on your part. In a form the user is looking at, an optimistic save that quietly fails is recoverable — they see the stale value eventually. In a serial review the user is already looking at the next frame, so a silently dropped decision is invisible until the pass is over and the counts are wrong. Building the refusal in from the start is the difference between a review tool and a review tool you can trust a bin to.

**.** Using identity rather than  to reject a JSON  as a rating of 1 is the kind of thing that reads like a mistake to a reviewer skimming, and is exactly right. That you also wrote the test asserting it is what makes it survive the next person who "fixes" it to .

** → .** This one is a judgement call reaching outside the new feature into the existing grid, and it's correct: a source review exists to judge framing, and a cropped thumbnail hides the edges of frame that the decision turns on. I called it out in the release notes as a deliberate change rather than letting people discover their thumbnails look different.

Also worth noting: shelling out to a Node harness so the browser-side behaviour is actually executed rather than asserted about is the right shape for panel tests here, and matches what the repo already does elsewhere.

One thing for next time, purely procedural: leave the version bump and CHANGELOG out of the PR — you did, which is what I'd have asked for. I handle those at release, so PRs that include them collide with each other whenever two land the same day.

I ran your branch in a worktree before merging (CI doesn't run automatically on a first-time contributor's PR), then merged locally and ran the full offline suite on the merge result: 3555 → 3557, no new failures.

If you want another one, the review bin is somewhere I'd welcome more attention.

samuelgursky:
Correcting my comment above — the two middle paragraphs lost their inline code spans on the way in, which gutted them. Restating both properly:

**`type(value) is not int`.** Using identity rather than `isinstance` to reject a JSON `true` as a rating of 1 is the kind of thing that reads like a mistake to a reviewer skimming — Python's `bool` is a subclass of `int`, so `isinstance(True, int)` is `True` and a rating of `true` would have been stored as 1. Your check is exactly right, and the fact that you also wrote the test asserting it is what will make it survive the next person who "fixes" it to `isinstance`.

**`object-fit: cover` → `contain`.** This is a judgement call that reaches outside the new feature into the existing review grid, and it's the correct one: a source review exists to judge framing, and a cropped thumbnail hides the edges of frame the decision turns on. I called it out in the release notes as a deliberate change rather than letting people discover their thumbnails look different.

PR #231, opened 2026-09-13
feat(security): confirm raw copy grade operations

Rohitkanithi:
## Summary

This PR adds an explicit confirmation-token step before the raw
`timeline_item_color copy_grades` action can call Resolve's
`TimelineItem.CopyGrades`.

`CopyGrades` is not a small metadata update. It replaces the target timeline
item's grade graph with the source item's grade. If a user or agent calls it by
mistake, the target's existing hand grade can be lost. The repo already knows
this API is dangerous and already shows a known-trap warning for the raw
`copy_grades` action. This PR keeps that warning, then adds the existing
one-time confirmation-token flow before the destructive call is allowed to run.

## Previous behavior

Before this change, the raw `copy_grades` flow was:

1. Caller invokes `timeline_item_color copy_grades`.
2. The destructive trap guard blocks the first call and asks for
   `acknowledge_trap=true`.
3. Caller retries with `acknowledge_trap=true`.
4. The raw action immediately resolves targets and calls `item.CopyGrades(...)`.

That meant the trap acknowledgement was the only final barrier before replacing
target grades.

## New behavior

After this change, the raw `copy_grades` flow is:

1. Caller invokes `timeline_item_color copy_grades`.
2. The existing trap guard still blocks the first call and asks for
   `acknowledge_trap=true`.
3. Caller retries with `acknowledge_trap=true`.
4. The action resolves the requested target IDs and returns
   `confirmation_required` with a one-time `confirm_token`.
5. Caller retries with the same params plus that `confirm_token`.
6. Only then does the action call `item.CopyGrades(...)`.

So the trap warning remains first, and the confirmation token becomes the final
explicit commit step before the destructive grade replacement.

## Confirmation preview

The confirmation response includes a preview shaped around the actual resolved
targets:

```json
{
  "operation": "timeline_item_color.copy_grades",
  "warning": "Replaces the entire node graph on every successfully resolved target item.",
  "target_count": 1,
  "target_ids": ["target-1"],
  "missing": []
}
```

This gives the caller a last chance to check what will be affected before
committing the operation.

## Token safety

This uses the repo's existing confirmation-token helpers.

That means:

- tokens are short-lived
- tokens are single-use
- tokens are bound to the action name
- tokens are bound to a fingerprint of the params
- if the caller changes `target_ids` after receiving a token, the token no
  longer matches and the action is refused

The PR does not introduce a second token system.

## Why this is scoped to raw `copy_grades`

The safer grade-copy paths already have their own guardrails:

- `safe_copy_grade` supports dry-run target inspection and confirmation
- `bulk_match_to_hero` plans first and confirms before execution

The gap was the raw `timeline_item_color copy_grades` action. It is intentionally
still available for advanced callers, but because it directly reaches
`TimelineItem.CopyGrades`, it should require the same final confirmation step
before execution.

## What changed in code

- `timeline_item_color copy_grades` now reuses `_timeline_items_for_grade_copy`
  to resolve targets and missing IDs.
- If confirmation is required and no token is present, it returns
  `_issue_confirm_token(...)` instead of calling `CopyGrades`.
- If a token is present, it is checked with `_consume_confirm_token(...)`.
- Only a valid token allows `item.CopyGrades(targets)` to run.
- The raw action help text now documents `confirm_token?` and says the action
  replaces target grades.

## Tests

Added `tests/test_copy_grades_confirmation.py` covering:

- raw `copy_grades` without a token returns `confirmation_required`
- `CopyGrades` is not called before confirmation
- a valid token allows the mutation
- changing params after token issuance is rejected by the existing fingerprint
  check
- the existing trap acknowledgement is still required before any confirmation
  token is issued

Validation run locally:

```text
venv/bin/python -m unittest tests.test_bulk_match_to_hero tests.test_copy_grades_confirmation tests.test_confirm_token tests.test_action_symbol_registry tests.test_execution_lifecycle
Ran 69 tests
OK

venv/bin/python -m py_compile src/server.py tests/test_copy_grades_confirmation.py
OK

git diff --check
OK

npm run smoke
OK
```


samuelgursky:
Merged and released as [v4.2.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.2.0). Thank you — this is the right gap and the right gate.

I ran it rather than reading it. Merged onto `main` in a worktree: **3,612 passed** with nine failures identical to `main` at the same commit (the known `source_media_guard` / `offline_fallback` worktree path artifact — they pass in the primary checkout). All drift guards green, `test_write_enforcement_ratchet` included. In the primary checkout after the bump: **3,620 passed, 1 skipped, 0 failed.**

Three things I checked specifically, because each could have turned this into a regression:

- **No lockout when gating is off.** Falling through to `_consume_confirm_token` with no token is safe — it returns `None` both when `require_confirm_token` is disabled and when no token is present, so `destructive.require_confirm_token: false` still bypasses cleanly instead of hard-refusing an action that can no longer be completed.
- **No `KeyError` on a missing `target_ids`.** You kept the bare `p["target_ids"]` subscript, which is correct here: the parameter is validated upstream and returns a `MISSING_TARGET_IDS` envelope before the handler runs.
- **The fingerprint actually binds what matters.** Your fourth test is the one that earns the design — a token issued for one target set is refused when `target_ids` grows, which is the case where a confirmation gate would otherwise be theatre.

One thing worth calling out that your summary undersells: routing target resolution through `_timeline_items_for_grade_copy` closes the *other* half of the docstring's warning. "No target existence check" was literally true before — IDs that resolved to nothing were silently dropped and the caller got `success: true` for a copy that reached fewer clips than asked. They now surface as `missing` in the preview. That is a silent-lie fix riding along with the gate.

### Two follow-ups, neither blocking

**1. The zero-target case still confirms and then acts.** If every ID is missing, or there is no current timeline, the action mints a token, and the confirmed call runs `CopyGrades([])`. The safe sibling refuses outright when `missing` is non-empty. Refusing early on `target_count == 0` would spend one fewer round trip to accomplish nothing.

**2. The larger half of this class is still open, and it is the granular twin.** `ti_copy_grades` in `src/granular/timeline_item.py` reaches `TimelineItem.CopyGrades` with **no trap guard and no confirm token** — it appears in neither `DESTRUCTIVE_ACTIONS_BY_TOOL` nor `execution_lifecycle`, so every gate reads it as a read. It is also strictly more dangerous than the compound action you just fixed, because it takes bare 0-based indices instead of unique IDs: an off-by-one silently replaces the wrong clip's grade graph, and `CopyGrades` leaves no recovery version to undo it with. The `--full` granular server is a shipped 377-tool surface, not a dev-only path.

If you want that one, it is yours — same shape as this PR, plus the registry and risk-table entries the compound action already has. Branch from current `main` (now v4.2.0), and no version bump or CHANGELOG in the PR; I cut the release.

Rohitkanithi:
Quick clarification on the ti_copy_grades follow-up: I checked current main, and it looks like this may already be handled there now. I can see ti_copy_grades has the trap acknowledgement, confirmation token flow, destructive hint, high-risk rating, and tests.
Are you pointing to something more specific that is still missing in that feature, or am I missing a part of the issue? I want to make sure I’m looking at the same gap before opening another PR.

PR #233, opened 2026-09-15
Stop default project archives crashing Resolve 21.1, and gate them

legionsound:
Archive coverage, the first of the pair you confirmed on #207. `ArchiveProject` had no `api_truth` entry and both archive actions sat in the ratchet's unrated backlog. The measurement found something worse than a gap: **on 21.1.0.14 a default call crashes Resolve.**

## What `ArchiveProject` actually does on 21.1.0.14

Measured on a disposable local project with one synthetic clip, archived by name, one isolated call per row. Presence was checked with `dir()` membership against a control, per your `hasattr` note.

| Case | Returns | Effect |
| --- | --- | --- |
| Open project, all flags off | `False` | nothing written |
| Closed project, all flags off | `False` | nothing written |
| Closed project, render cache only | `False` | nothing written |
| Closed project, **source media** | `None` | empty directory at target, **Resolve crashes** |
| Closed project, **proxy media** | `None` | empty directory at target, **Resolve crashes** |
| Unrelated file at target, flags off | `False` | file byte-identical |
| Populated directory at target, flags off | `False` | untouched |
| **Source media onto an existing file** | `None` | **Resolve crashes**, file byte-identical |

- Every `False` is instant, and Resolve logs nothing about it.
- A crashing call comes back through the bridge as `None`, not a bool, and every later handle is dead.
- **Media-flag calls crashed 4 of 4; flags-off calls crashed 0 of 5.** All four crashes (one on an earlier attempt inside a Blackmagic Cloud library) are SIGSEGV stamped the same second as the call, with identical top stack frames in Fusion script-symbol teardown on the UI thread.
- A fifth crash in the same session, during `DeleteProject` then `LoadProject` shortly after a relaunch, has a **different** stack past the generic crash-handler frames. So the archive signature isn't general instability, and I've kept that crash out of the claim. Crash uptimes were 1:45, 14:16, 1:31 and 0:45.

**Collision, the case you most wanted:** the destination is never the casualty. A file or folder already at the target survived every case byte for byte, including the crash. Resolve can't create its bundle folder over an existing file and crashes anyway, so the crash is in starting the archive job, not in writing the destination. What's lost is unsaved work in the open project.

19.1.3.7 agrees where it was measured: the repo's own mode matrix (2026-08-02) and the project-lifecycle kernel recorded `False` for both `.dra` and folder paths with flags off. **No scriptable call on either build has produced an archive.** The coverage row saying "API accepts; archiving is slow" is corrected.

## What changed

- **`src/utils/archive_guard.py`** (shared):
  - every flag defaults off;
  - only real booleans are accepted, since `bool("false")` is `True` and two flags crash;
  - `src_media` and `proxy_media` are refused unless `acknowledge_trap=true`;
  - the result reports the native return as observed: `True`, `False` ("wrote nothing"), or `None` ("likely crashed; check Resolve is running").
- **Compound `archive`**: used to default `src_media=True, render_cache=True`. It now goes through the guard. That's a behaviour change, and it's the point.
- **`safe_project_archive`**: keeps `allow_media_archive` as the size guard and now also needs `acknowledge_trap` for the crashing flags. One guards size, the other the crash.
- **Granular `archive_project`**: defaults off, `acknowledge_trap`, `DESTRUCTIVE_TOOL` annotation (per your #223 note) and `@granular_destructive_op()`. It previously fell through to plain `WRITE_TOOL`, because `archive_` isn't in the verb table, and had no gate.
- **Both write tables**: both compound actions are in the `destructive_hook` registry and rated MEDIUM in the risk sets, and they're off the ratchet backlog. `safe_project_archive` honours `dry_run` natively and is in `NATIVE_DRY_RUN_ACTIONS`; the scan agrees. Raw `archive` doesn't read `dry_run`, so a dry run there is refused as before.
- **`api_truth`**: a new `ProjectManager.ArchiveProject` entry (`verified_on: 21.1.0.14`), plus `ACTION_SYMBOLS` for both actions, so the fact rides on every result through your existing `known_limitation` push.

No new tools, so counts are unchanged at 37/387. No version bump or CHANGELOG. Branched from `main` at v4.5.2 (`ce3362a`) and re-checked unmoved right before opening. Full suite 3,687 passed, 22 skipped; the ratchet passes with two entries removed from its backlog and nothing added.

## A design call I'd rather you make

The entry deliberately does **not** set `destroys_prior_work`. That flag is symbol-level, so every mapped action would refuse until acknowledged, including the flags-off call, which is a harmless no-op. I put the refusal at parameter level in the wrappers instead. If you'd rather it be symbol-level, it's a one-line change plus the granular confirmation tier, and I'll do it.

MEDIUM is also a judgment call. The action as wrapped either no-ops or, only when acknowledged, may crash, and it never overwrites. I'm happy to take HIGH if you see it differently.

## Evidence tiers

- **Live on 21.1.0.14:** every row in the table, both collision cases, the crash signature, and the four-versus-zero split.
- **Reported, not reconfirmed here:** the 19.1.3.7 `False` results, from the existing mode matrix and kernel doc.
- **Unit-tested only:** every wrapper behaviour, including defaults, boolean refusal, crash-flag refusal, acknowledged pass-through, outcome reporting, registration, rating and native-dry-run listing. **I did not run the wrappers live against Resolve.** Their flags-off path would return `False` (0 of 5 crashed natively), and I stopped live runs after Resolve crashed on relaunch.

## Not tested

- `RestoreProject` round-trip. Nothing produced an archive to restore.
- Render cache on a project that has one.
- Headless mode on 21.1.
- Builds after 21.1.0.14, and Windows or Linux.

Evidence: `docs/reference/project-archive.md` and `docs/reference/evidence/project-archive-21.1.json`.


samuelgursky:
Merged as-is and released as [v4.6.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.6.0). Thank you — this is exactly the input the `api_truth` refusal mechanism was built for, and the measurement is the best part: one isolated call per row, the collision cases, and keeping the `DeleteProject`/`LoadProject` crash out of the claim because its stack differed.

**What I checked before merging.** `main` merged into your branch in a scratch worktree: 3691 tests, only the 4 known worktree path artifacts. Your 19.1.3.7 half matches this repo's own evidence — the 2026-08-02 mode matrix and the project-lifecycle kernel doc both recorded `False` with every flag off. Then a live run here on Studio 19.1.3.7 through the new wrapper: the crash-flag refusal, the non-boolean refusal (`"false"` → `INVALID_ARCHIVE_FLAG`), and the flags-off call all behaved as the tests say — native `False`, nothing written, `known_limitation` attached to the result. I did **not** send a media flag to 19.1.3.7, so whether the crash reproduces there stays unmeasured, and the release notes say so. Your 21.1.0.14 rows land as reported, which your `reported` tag already does.

**Your two design calls, decided.**
1. **Parameter-level refusal, not symbol-level `destroys_prior_work`.** You had it right. A flags-off call is a harmless no-op, and a gate that refuses harmless work teaches people to switch the gate off. The v4.5.2 notes on `_safe_mode_allows` say the same thing about over-rating.
2. **MEDIUM stays.** The wrapped action never overwrites and the crashing path is behind an acknowledgement. HIGH would rate the acknowledgement, not the action.

**What changed on landing:** nothing in your code. I added the `acknowledge_trap` parameter to the `safe_project_archive` line in `docs/SKILL.md` with a pointer to your `project-archive.md`, and folded the release into the same minor bump as some unreleased offline relayout work that was already on `main`.

The read-completion branch is next whenever you rebuild it from `main` at v4.6.2. Tool counts are unchanged at 37/387.

PR #234, opened 2026-09-15
fix(security): refuse empty copy grade targets

Rohitkanithi:
## Summary

This PR handles the remaining zero-target edge case in raw
`timeline_item_color copy_grades`.

After #231, raw `copy_grades` correctly requires the known-trap acknowledgement
and then a confirmation token before it can call Resolve's
`TimelineItem.CopyGrades`. One edge case was still awkward: if every requested
`target_ids` value failed to resolve, the tool could still issue a confirmation
token and then, after confirmation, call `CopyGrades([])`.

That spends a confirmation round trip to do nothing, and it makes the caller
confirm an operation that has no valid target. This PR refuses that case before
issuing a token.

## Previous behavior

For a call like:

```json
{
  "target_ids": ["missing-1"],
  "acknowledge_trap": true
}
```

the tool resolved:

- `targets = []`
- `missing = ["missing-1"]`

but still continued into the confirmation-token flow.

## New behavior

If no target items resolve, `copy_grades` now returns a structured error:

```json
{
  "error": {
    "code": "NO_COPY_GRADE_TARGETS",
    "category": "invalid_input",
    "message": "copy_grades resolved no target items; nothing would be copied.",
    "state": {
      "target_ids": ["missing-1"],
      "missing": ["missing-1"]
    }
  }
}
```

No `confirm_token` is issued, and `CopyGrades` is not called.

## Why this shape

The confirmation-token path is for destructive work that is ready to commit.
When zero targets resolve, there is no destructive work to confirm. Returning a
caller-fixable error is clearer and avoids asking the user to approve a no-op.

The refusal happens after target resolution, so mixed target sets still behave
as before: resolved targets can be previewed, and missing IDs are still reported
in the confirmation preview. Only the all-missing / zero-target case is refused
early.

## Tests

Added coverage to `tests/test_copy_grades_confirmation.py` for:

- all requested target IDs missing
- no confirmation token being issued
- no `CopyGrades` call happening
- the structured `NO_COPY_GRADE_TARGETS` error including the missing IDs

Validation run locally:

```text
venv/bin/python -m unittest tests.test_copy_grades_confirmation tests.test_granular_copy_grades_confirmation tests.test_granular_destructive_op tests.test_write_enforcement_ratchet
Ran 66 tests
OK

venv/bin/python -m py_compile src/server.py tests/test_copy_grades_confirmation.py
OK

git diff --check
OK

npm run smoke
OK
```

## Notes

No version, changelog, README badge, package, or lockfile changes are included.


samuelgursky:
Merged as-is and released as [v4.6.1](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.6.1). Thank you — and thank you for leaving the version bump out this time; that made it a two-file merge.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3691, only the known worktree artifacts), and I confirmed the class is confined to the compound server — the granular `ti_copy_grades` already refuses an empty or out-of-range index list before it reaches the token flow, so there was nothing to generalise. No Resolve call path changed, since the refusal returns before the native call, so no live run was needed.

Nothing changed on landing beyond the changelog entry.

PR #235, opened 2026-09-15
fix(granular): redirect macOS /tmp in the granular _resolve_safe_dir too

Dev-next-gen:
While reading `_resolve_safe_dir` I noticed there are two copies of it, and they disagree on macOS. The one in `src/server.py` redirects `/tmp`, `/tmp/...`, `/private/tmp` and `/private/tmp/...` to `~/Documents/resolve-stills`, with a comment saying Resolve's still/gallery export fails silently into both (live-verified 2026-07-03). The copy in `src/granular/common.py`, used by the granular (`--full`) server, still only matches `/var/` and `/private/var/`.

Two granular tools reach it with a `/tmp` path on macOS. `save_project`'s export fallback stages in `_resolve_safe_dir(tempfile.gettempdir())`, and `gettempdir()` is `/tmp` when `TMPDIR` is unset. `encrypt_dctl` passes `target.parent.resolve()`, so an output under `/tmp` arrives as `/private/tmp/...`. The compound server redirects both cases; the granular server hands Resolve the temp directory as is.

The change copies the compound server's Darwin condition into the granular helper and updates its docstring. Linux and Windows are untouched. The new `tests/test_granular_safe_dir.py` patches `platform.system` to `Darwin` and checks that both copies agree on a set of paths. On `main`, four of those paths fail (`/tmp`, `/tmp/out`, `/private/tmp`, `/private/tmp/out`); with the patch both tests pass, along with `test_resolve211_encryption`, `test_gallery_still_export_cleanup`, the granular suites and `test_write_enforcement_ratchet` (154 tests). I don't have a Mac with Resolve here, so I haven't reproduced the silent export failure myself. The fix relies on the behaviour `src/server.py` already records.


samuelgursky:
Merged as-is and released as [v4.6.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.6.2). Thank you — the premise checked out exactly against `main`: the compound `_resolve_safe_dir` carries the `/tmp` and `/private/tmp` condition from the 2026-07-03 live run, and the granular copy did not, with both callers you named (`save_project`'s export fallback and `encrypt_dctl`'s resolved parent) reaching it.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3691, only the known worktree artifacts). You said you have no Mac to reproduce the silent export failure on; that is fine here — the fix rests on behaviour this repo already recorded from a live run, and the release notes say so rather than claiming a new measurement.

Nothing changed on landing beyond the changelog entry. Your equivalence test is a good shape for the two-copies problem in general, and it is the kind of guard that will catch the next drift between the servers.

PR #236, opened 2026-09-15
fix(timeline): read ripple="false" as false in delete_clips and lift_range

Dev-next-gen:
After #218 fixed `dry_run="false"` being read as true, I looked for other flags still tested with a bare `bool()`, and `ripple` on timeline deletes is the one where it matters most. `timeline.delete_clips` does `ripple = bool(p.get("ripple", False))` and hands that to `Timeline.DeleteClips`, and `_timeline_lift_range_impl` (and the delete step of `move_clips`) does the same. A client that sends `ripple="false"` therefore gets a ripple delete: the gap closes and everything downstream shifts, which is exactly what the caller declined.

The guards around it read the flag the same way, so they don't catch it. `_action_will_gate_pending_confirm` and `destructive_hook.is_strict_required` both treat `"false"` as a ripple delete, and so does the blast-radius branch in `execution_lifecycle.RiskClassificationHook`. With confirm tokens on, the user is asked to confirm a ripple delete they never requested, and once they confirm it runs as a ripple.

The change sends all six readings through `coerce_bool` from `src/utils/bool_params.py`, the helper #218 added. Real booleans behave exactly as before, and so do `"true"`/`"false"`/`"1"`/`"0"`. The one difference is that a string `coerce_bool` doesn't recognise now falls back to non-ripple instead of ripple, and for a destructive flag that seems like the safer default.

The new `tests/test_delete_clips_ripple_string.py` drives the `timeline` dispatcher and `_timeline_lift_range_impl` with a fake timeline that records the `ripple` argument, and checks the gate, the strict rule and the classifier. On `main`, three of its four tests fail (`DeleteClips` receives `True` for `"false"`, and the gate still fires). With the patch all four pass. I also ran the whole suite with `unittest` on Windows before and after. The failing set is identical in both runs (98 failures and errors, which come from Windows sqlite file locks and the missing numpy on this machine, not from this change), and the only difference is the four new tests, which pass.


samuelgursky:
Merged as-is and released as [v4.6.4](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.6.4). Thank you — and thank you for following the #218 thread to the flag where it matters most. A ripple the caller declined, confirmed through a gate that misread it the same way, is the worst version of this class.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3699 with #237 alongside, only the known worktree path artifacts), and I confirmed `coerce_bool` was already imported in `destructive_hook` so the six call sites all resolve to the one helper. Your call to let an unrecognised string fall to non-ripple is the right default for a destructive flag, and the changelog says so. The change is parameter parsing ahead of the native call, so no live run was needed.

Nothing changed on landing beyond the changelog entry. If your pipeline keeps a list of the remaining bare `bool()` reads on destructive flags, that list is welcome as an issue.

Dev-next-gen:
Merci, and thank you for shipping it in v4.6.4.

It is a real pleasure to contribute, and to add small pieces of code to the film industry 😉

DaVinci Resolve has come up a lot on my dashboard. I was afraid of flooding you — but if you are the one asking, that changes everything 🫡

(@Dev-next-gen)

Dev-next-gen:
P.S. — @samuelgursky on your question about the list: I would rather not hand one over yet, for fear of sending you false positives. As soon as I am confident in its precision and its quality, it is yours. I could even run my pipeline as an audit focused entirely on this repository — that would not produce a stream of small pull requests, but it would give you a real overview.

(@Dev-next-gen)

PR #237, opened 2026-09-15
Complete the Resolve 21.1 transcription and timeline-item type reads

legionsound:
The read-only pick from #207, rebuilt from `main` at v4.6.3 (`919ed9c`) and re-checked unmoved right before opening.

Two 21.1 reads that one interface or the other couldn't do:

- **Granular `get_media_pool_item_transcription(clip_id, use_nested_clip_transcription=False)`**: the complete `GetTranscription` dictionary, with timed words. The compound `media_pool_item get_transcription` already prefers it on 21.1; granular had no way to reach it.
- **Compound `timeline_item get_type`** and **granular `get_timeline_item_type(track_type, track_index, item_index)`**: the native lowercase `TimelineItem.GetType`, addressed by position like the other 21.1 item readers (`get_speed`, `get_fades`).

To be upfront on one point: granular already returns `type` inside the larger item-properties bundle, looked up by ID, via an unguarded `GetType()`. The new tool is the position-addressed twin of the compound action, with the same `_requires_method(..., "21.1")` guard. It isn't the only way to read type.

All three are read-only. Granular tools carry `READ_ONLY_TOOL`; nothing is registered as a write, and the ratchet passes with nothing added. `CODE_FLOORS` gains `MediaPoolItem.GetTranscription` and `TimelineItem.GetType` at 21.1. Granular count 387 → 389; compound unchanged at 37; docs and agent rules regenerated. No version bump or CHANGELOG. Full suite 3,694 passed, 22 skipped.

## Evidence

**Live on Studio 21.1.0.14, today, against this branch**, in a local disposable project, with the synthetic camera clip from the Smart Switch fixture transcribed natively:

| Read | Native | Compound | Granular |
| --- | --- | --- | --- |
| Transcription | 1 English segment, 19 timed words | identical segments | identical dict |
| Item type | `video` | `video` | `video` |

This matches the 2026-09-09 measurement exactly. `tests/live_resolve211_read_completion.py` reproduces it; it now asserts it's on the local database and reads the fixture from the current timeline rather than a named project. One small measured fact: `Transcription Status` reads empty while processing and `Transcribed` when done.

**Unit-tested:** argument validation, missing-method refusal below 21.1, and both interfaces forwarding the native value unchanged.

## Not tested

- Nested-clip transcription (`use_nested_clip_transcription=True`).
- Non-English audio and multi-speaker segmentation.
- `GetType` on audio, subtitle, generator or Fusion items. Only a video item was measured.
- Builds other than 21.1.0.14.


samuelgursky:
Merged as-is and released as [v4.7.0](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.0). Thank you — branching from `main` at v4.6.3 right before opening worked exactly as it did for `search_api`: nothing to adapt, tool counts and agent rules already regenerated, ratchet clean.

**Checked before merging.** `main` merged into your branch in a scratch worktree: 3699 tests, only the 4 known worktree path artifacts. All the drift guards pass at 389 (`test_doc_tool_counts`, `test_agent_rules_drift`, `test_action_list_drift`). Your 21.1.0.14 table lands as reported, as before. I tried to re-measure the 19.1.3.7 floor refusal through the new `get_type` here, but the open project had no timeline or clips and I was not going to create one around a live session, so that stays as "the same `_requires_method` guard the twelve v2.216.0 readers measured refusing on 19.1.3.7", and the release notes say exactly that.

**Being upfront about `type` twice** was the right call and did not change the decision: a position-addressed twin of the compound action behind the version guard is more useful than an unguarded field in a bundle, and it keeps the two servers symmetric.

That closes both picks from #207. Whenever you rebuild the next held branch, `main` is at v4.7.0 (`0ffcf7b`), 37 compound / 389 granular.

PR #238, opened 2026-09-15
fix(confirm-token): let a token handed back as confirmToken be redeemed

Dev-next-gen:
`ConfirmTokenStore.consume` accepts the token under either spelling:

```python
token = (params or {}).get("confirm_token") or (params or {}).get("confirmToken")
```

and the compound server agrees — twenty-eight gated actions in `src/server.py` are
written as `if "confirm_token" not in p and "confirmToken" not in p and
_confirm_token_required():`, so either spelling counts as "the caller already holds a
token" and suppresses re-issuance.

`fingerprint()` does not agree. It strips only the snake_case key:

```python
if isinstance(payload["params"], dict) and "confirm_token" in payload["params"]:
    payload["params"] = {k: v for k, v in payload["params"].items() if k != "confirm_token"}
```

So for a caller that spells it `confirmToken`, the token stays in the params and the
hash of "the request" is not the same before and after issuance — which is exactly what
that block's own comment says it exists to prevent. I ran this against the store as it
stands on main:

```
tokens held after issue: 1
round 2 (echo the token back as confirmToken) -> CONFIRM_TOKEN_FINGERPRINT_MISMATCH
tokens held after the refusal: 0   <- already spent
round 3 (retry the same token)              -> CONFIRM_TOKEN_INVALID
```

The camelCase caller can never get through. Redemption is refused as a params mismatch,
and because `self.tokens.pop(token, None)` runs before the fingerprint check the token
has already been spent by the time it is refused, so the retry the remediation string
asks for reports a different and misleading reason. The remediation itself — "either
re-issue the token with current params or roll back the params change" — cannot help,
because re-issuing produces the same mismatch on the next round. That is a closed loop
on the gate standing in front of `TimelineItem.CopyGrades` and project deletion.

Both servers share this store, so both surfaces are affected.

The fix puts the two spellings in one tuple that `fingerprint()` strips and `consume()`
reads, so the pair cannot drift apart again. I also dropped the now-redundant membership
test before the dict comprehension.

I kept a test for the case that worried me while writing it: stripping the key must not
loosen the gate, so `test_camel_case_token_still_binds_to_its_params` checks that a
camelCase token presented against different `clip_id` is still refused.

Verification, on Windows with `python -m unittest`:

- the new file fails 5 of its 6 tests against main and passes all 6 with the change
- `test_confirm_token`, `test_confirm_token_preferences`, `test_copy_grades_confirmation`
  and `test_granular_copy_grades_confirmation` — 38 tests, all pass
- full suite before and after: identical set of failures (98 pre-existing on this box,
  which has no numpy, pillow or ffmpeg), so no regression

One thing I could not check: I have no Resolve Studio here, so this is the token store
exercised directly, not a live confirm-and-delete round trip.


samuelgursky:
Merged as-is and released as [v4.7.1](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.1). Thank you — this one mattered more than its diff size suggests. A gate that cannot be passed by a whole class of clients is a gate people learn to switch off.

Checked before merging: I reproduced your three rounds on `main` through the compound helpers (mismatch, then `CONFIRM_TOKEN_INVALID` on the retry, while the snake_case spelling redeemed fine), grepped the other token-reading sites in `src/server.py` and `destructive_hook.py` to confirm they already handle both keys so the class was confined to `fingerprint()`, and ran the full suite with `main` merged into your branch in a scratch worktree: 3705, only the known worktree path artifacts. The test I was gladdest to see is the one you kept — the camelCase token against different `clip_id` still refused.

Nothing changed on landing beyond the changelog entry. No Resolve was involved, as you said; the store is exercised directly and that is the right level for this fix.

PR #240, opened 2026-09-16
fix(guards): a false spelling must not grant permission on six opt-in flags

Dev-next-gen:
You asked for this one on #239: "the same bare read is still on a few opt-in flags that gate something destructive — allow_media_archive, close_current, allow_generate, allow_render, allow_switch — and the archive guard I landed this morning in v4.6.0 reads acknowledge_trap the same way. [...] If you want them, they are yours; one PR for the set is fine." (https://github.com/samuelgursky/davinci-resolve-mcp/pull/239#issuecomment-5690079764)

So this is the set, in one change. I went looking for every reading of each flag rather than the first one, because on #239 a single flag turned out to have five.

## Why these are worse than overwrite

overwrite="false" made a destructive act happen that the caller had asked to be refused. These six open in the opposite direction: they are opt-in permissions that default to off, so "false" does not perform the act — it **grants the permission to perform it**. The caller who writes it out is the caller being explicit about not wanting it, and they are the one who gets it.

With bare truthiness, a client that stringifies its JSON scalars gets, by saying no:

- `allow_media_archive="false"` — ArchiveProject runs with source media on, which on Studio 21.1.0.14 is the SIGSEGV your own `archive_guard` docstring measured. `read_flags` is strict about `src_media`/`render_cache`/`proxy_media` for exactly that reason; the opt-in that gates them was not.
- `acknowledge_trap="false"` — the crash refusal stands down in both archive paths, and in `destructive_hook._trap_acknowledged` the trap guard stands down in front of `TimelineItem.CopyGrades`, which replaces a hand-built node graph with no version to go back to.
- `close_current="false"` — the currently open project is closed and deleted. That is the call that loses work someone is in the middle of.
- `allow_generate="false"` — CreateSubtitlesFromAudio runs on the timeline.
- `allow_render="false"` — RenderWithQuickExport starts and writes to disk.
- `allow_switch="false"` — SetCurrentDatabase runs, which closes whatever project is open along with its unsaved changes.

Nine readings in all: two for `close_current`, three for `acknowledge_trap` (both archive paths plus the trap guard in `src/utils/destructive_hook.py`), one each for the rest. They now go through `coerce_bool` from `src/utils/bool_params.py` — the helper v4.1.0 introduced for `dry_run`, 84a8a49 applied to `ripple` for v4.6.4, and #239 applied to the install guards for v4.7.2. Nothing new invented.

I deliberately left the neighbouring `dry_run` readings in `_safe_quick_export` and `_safe_set_current_database` alone. `dry_run="false"` is truthy there too, but the consequence is that the call stays in preview, which fails safe — a different bug with a different risk profile, and not what you asked for. Same for the granular tools: `archive_project` and `copy_grades` annotate `acknowledge_trap: bool`, so the SDK coerces before the handler sees it. Nothing else touched.

## What changed, and why it is safe

The only behaviour that moves is how a **string** is read. A real `True`/`False` is untouched — `coerce_bool` falls through to `bool(value)` for non-strings. `"true"`, `"yes"`, `"on"`, `"1"` still grant the permission, in any case. `"false"`, `"no"`, `"off"`, `"0"` now refuse it, and a string that is neither falls back to the parameter's existing default, which for all six is off. So the only callers whose result changes are the ones that were being granted a permission they had not asked for, and the refusal they now get is the documented one, with the same `retry_with` payload pointing at the flag to set.

## Proof

`tests/test_permission_flags_string_false.py`, one test per flag, each over six spellings — "false", "False", "FALSE", "no", "off", "0". Each case asserts the refusal **and** that the effect did not happen: no ArchiveProject call and no file at the target path, no CloseProject and no DeleteProject, no CreateSubtitlesFromAudio, no RenderWithQuickExport, no SetCurrentDatabase, and the grade-destroying handler never entered. The effect assertion comes first in every case, so a regression reports the call that got through rather than a missing key.

Against `main`'s `src/`: **54 of 54 subtests fail**, and every one of them fails on the effect assertion, not the envelope — the destructive call really ran in all 54.

```
$ python -m pytest tests/test_permission_flags_string_false.py -q     # main src
54 failed, 9 passed in 1.05s

$ python -m pytest tests/test_permission_flags_string_false.py -q     # this branch
9 passed, 54 subtests passed in 0.29s
```

Full suite, `python -m pytest tests -q`:

| | passed | subtests passed | failed | skipped |
|---|---|---|---|---|
| main (0e0328a) | 3701 | 1417 | 5 | 30 |
| this branch | 3710 | 1471 | 5 | 30 |

Exactly +9 tests and +54 subtests, nothing else moved. The 5 failures are the same ones before and after, all in `tests/test_offline_fallback.py`, and they are my environment rather than the code — I did not run `npm install`, so the bundled drt parser cannot require `jszip`. I also ran the suite under `python -m unittest discover -s tests -t .` on Windows: 3710 → 3719 tests, with an identical 35 failures / 63 errors on both sides, which are the repo's POSIX-path tests on a platform CI does not cover.


samuelgursky:
Merged as-is and released as [v4.7.3](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.3). Thank you — and thank you for taking it as one PR for the set, for going after every reading of each flag rather than the first, and for putting the effect assertion ahead of the envelope check. "54 of 54 fail on the effect" is the sentence that made this a release rather than a review.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3719, only the known worktree path artifacts), the diff matches your nine readings line for line, and I grepped the tree for what was left — one bare opt-in read remains, `allow_timeline_mismatch` on `apply_trace_plan`, which gates a name mismatch rather than a destructive act, so I left it out on purpose and said so in the changelog. Your call to leave the `dry_run` neighbours alone was right for the same reason in reverse.

Nothing changed on landing beyond the changelog entry. That is four releases from your pipeline today (v4.6.2, v4.6.4, v4.7.1, v4.7.2) plus this one, each a real defect with a test that fails on the old code. Whenever you are ready to share the list, false positives and all, it is welcome as an issue.

PR #242, opened 2026-09-16
fix(control-panel): let the panel serve on the ::1 loopback it accepts

Dev-next-gen:
I tried to open the control panel on the IPv6 loopback. Both guards accept `::1`: `_CONTROL_PANEL_LOOPBACK_HOSTS` in `open_control_panel`, and the panel's own `--host` check, whose refusal message lists it as an allowed value. But the panel can never start there. `main()` builds a plain `ThreadingHTTPServer`, which is AF_INET, so binding `("::1", port)` raises `socket.gaierror` and `open_control_panel(host="::1")` comes back with "Control panel child exited (rc=1) before serving". There's a second problem behind it: the launch URL, the URL written to the pidfile and the `/api/boot` URL in `_control_panel_probe` are all built as `http://::1:<port>/`, which neither a browser nor urllib can parse.

The fix is small. `make_panel_server()` uses an AF_INET6 subclass when the host is an IPv6 literal, and the three URLs now put the host in brackets. Nothing else had to change: the Host/Origin gate already accepts `[::1]` (the existing auth tests cover that), and IPv4 and `localhost` go through the same code paths as before.

`tests/test_control_panel_ipv6_loopback.py` launches the real panel through `_open_control_panel({"host": "::1", ...})` and then uses what it gets back: a GET on `/` over `::1`, the issued URL parsed with `urlsplit`, and `_control_panel_probe("::1", ...)` with the issued token. It skips when the host has no IPv6 loopback. On a92d523 the two launch tests fail with the gaierror traceback in the log tail, and the helper test errors because the helpers don't exist yet. With the change, all three pass on Linux and on Windows. On Linux, the full `unittest discover` run gives the same result before and after, apart from these three tests: the only failures left are the 5 `test_offline_fallback` drt errors, which fail on main in that environment too.


samuelgursky:
Merged as-is and released as [v4.7.5](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.5). Thank you — a guard that names `::1` as allowed in its own refusal message and then cannot bind it is exactly the kind of contradiction that never gets noticed until someone tries, and the second problem behind it (the unbracketed URLs) would have bitten the moment the bind was fixed.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3733 with #244 alongside, only the known worktree path artifacts), and your test **ran rather than skipped** here — the real panel came up on `::1` on macOS, answered `GET /` over IPv6, handed back a URL `urlsplit` parses to `[::1]:<port>`, and passed the `/api/boot` probe with the issued token. That is the strongest kind of evidence for a fix like this and I said so in the changelog.

Nothing changed on landing beyond the changelog entry. Six releases from your pipeline in two days now, each with a test that fails on the old code.

PR #244, opened 2026-09-16
fix: set serverInfo.version to project VERSION (#243)

### LINKED ISSUE #243: serverInfo.version reports the MCP SDK version (1.30.0) instead of the project version
### What happened

  The server advertises the MCP SDK's version as its own. In the `initialize`
  response, `serverInfo.version` is the version of the installed `mcp` package
  (1.30.0), not the project version (4.7.4, `VERSION` in `src/server.py:14`).
  Every MCP client shows that number, so the version a user reports back — in a
  bug report, for instance — is the SDK's, not this project's.

  ### Steps to reproduce

  1. Start the server in any transport, e.g. `python src/server.py`.
  2. Perform the MCP `initialize` handshake (any client, or a raw POST to
     `/mcp` when using `--transport streamable-http`).
  3. Read `result.serverInfo` in the response.

  ### Expected

  "serverInfo": {"name": "DaVinciResolveMCP", "version": "4.7.4"}

  Actual

  "serverInfo": {"name": "DaVinciResolveMCP", "version": "1.30.0"}

  1.30.0 is the installed mcp SDK version; it changes when the SDK is
  upgraded, with no relation to this project's releases.

  Cause: src/server.py builds FastMCP("DaVinciResolveMCP", …) without a
  version, and FastMCP.__init__ has no version parameter at all in SDK 1.30
  (verified with inspect.signature). It constructs the low-level Server
  without one (mcp/server/fastmcp/server.py:221), so
  Server.create_initialization_options() falls back to the package version:
  server_version=self.version if self.version else pkg_version("mcp")
  (mcp/server/lowlevel/server.py:183).

  Suggested fix — set it on the low-level server right after building mcp:

  mcp = FastMCP("DaVinciResolveMCP", instructions=…)
  mcp._mcp_server.version = VERSION

  Verified locally: with that line, create_initialization_options().server_version
  returns 4.7.4 instead of 1.30.0. It does reach through a private attribute,
  so the tidier route is an upstream SDK request to expose version on FastMCP;
  until then this one line is contained and easy to revisit. The granular server
  instance looks like it needs the same treatment.

  Failing call

  Not an error — no call fails. The wrong value is in the initialize handshake:
  result.serverInfo.version = 1.30.0.

DYNOSuprovo:
Closes #243.

### Problem

The MCP `initialize` handshake was reporting the installed MCP Python SDK version (`1.30.0`) in `serverInfo.version` instead of the project version (`4.7.4`, `VERSION` in `src/server.py:14` and `src/granular/common.py:96`).

Because `FastMCP.__init__` in the 1.x MCP SDK does not accept a `version` argument, `Server.create_initialization_options()` falls back to `pkg_version("mcp")`. Any MCP client connecting to the server therefore received the SDK version as the server version.

### Solution

1. **Explicitly assign `VERSION` to `_mcp_server.version`**:
   - In `src/server.py` (compound server):
     ```python
     if hasattr(mcp, "_mcp_server"):
         mcp._mcp_server.version = VERSION
     ```
   - In `src/granular/common.py` (granular server):
     ```python
     if hasattr(mcp, "_mcp_server"):
         mcp._mcp_server.version = VERSION
     ```
2. **Fallback in test/stub mode**:
   - In `src/utils/mcp_import_stubs.py`, initialized `self._mcp_server = types.SimpleNamespace(version=None)` on the stub `FastMCP` class.
3. **Unit tests**:
   - Added `tests/test_server_info_version.py` asserting both `mcp._mcp_server.version` and `mcp._mcp_server.create_initialization_options().server_version` report `VERSION` (4.7.4).

### Verification

```bash
pytest tests/test_server_info_version.py # 2 passed
pytest tests/test_mcp_resources.py       # 10 passed
pytest tests/test_mcp_prompts.py         # 6 passed
```


samuelgursky:
Merged as-is and released as [v4.7.6](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.6). Thank you, and welcome — a first contribution that fixes both server instances, gives the test stub the same attribute so the offline suite keeps working, and pins the actual `create_initialization_options().server_version` rather than the attribute alone is exactly the shape this repo asks for.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3733 with #242 alongside, only the known worktree path artifacts), I confirmed the SDK fallback line you and @Eniot666 pointed at (`server_version=self.version if self.version else pkg_version("mcp")`), and at the new stamp the compound server's initialization options report `4.7.6`. Your CI run was parked as "action required" because GitHub holds first-time contributors' workflows; I approved it, and it went green.

Nothing changed on landing beyond the changelog entry, which credits both of you. The private-attribute reach is noted there as deliberate, so whoever upgrades the SDK past a version that exposes it knows what to replace.

PR #245, opened 2026-09-17
fix(tests): install the offline bootstrap under `unittest discover` too

Adi202001:
## The bug

`tests/__init__.py` points `RESOLVE_MCP_LOG_FILE` at a temp file and installs `offline_guard` before any test module imports `src.server`. Its docstring claims that both `python -m unittest tests.test_x` and `python -m unittest discover -s tests` import it first. Only the dotted form does.

`discover` given a *path* leaves `top_level_dir` at that path, inserts `tests/` on `sys.path`, and imports every module under its bare name — `test_clip_colors`, not `tests.test_clip_colors` — so the package `__init__` never executes.

Measured on this repo, running `python -m unittest discover -s tests`:

- the first module to `import src.server` attached the root logger's `FileHandler` to the operator's real `logs/server.log` — ~1.8 MB of suite output in one run;
- the offline guard was absent for the entire run, leaving the suite free to connect to, or launch, a real Resolve — the exact behaviour `tests/offline_guard.py` exists to stop.

`tests/test_log_isolation.py` is the tripwire written for this. It failed in the full discovery run (`a handler on the root or resolve-mcp logger writes to logs/server.log during the offline suite`) and passed when run alone.

## The fix

Discovery imports `sorted(os.listdir(start_dir))`, so a module whose name sorts ahead of every other `test_*.py` loads first. `tests/test_0000_offline_bootstrap.py` imports the `tests` package there — the existing idempotent bootstrap — and asserts the log redirect and the guard are in place. A third test asserts it still sorts first, so a future file named earlier fails loudly instead of silently restoring the bug.

`tests/__init__.py`'s docstring now states what actually holds on each runner.

## Verification

- `python -m unittest discover -s tests`: `logs/server.log` no longer grows, no handler targets it, and the `test_log_isolation` failure is gone (3650 tests; 1 failure + 11 errors → 11 errors).
- The 11 remaining errors are environment, not code: 6 need `numpy` (not in `requirements.txt`), 5 are `Cannot find module 'jszip'` from an uninstalled `node_modules/`. Both sets are unchanged by this PR.
- Release static checks pass: `tests/test_import.py`, the seven drift guards, `audit_api_parity.py`, `gen_api_limitations.py --check`, `audit_readwrite_symmetry.py --check`, `agent-rules/generate.mjs --check`, `git diff --check`.


samuelgursky:
Merged as-is and released as [v4.7.7](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.7). Thank you, and welcome — this is a first contribution that corrects a claim the repo's own docstring made and the maintainer believed. I had been running `discover -s tests -t .` all week, which is exactly why I never saw it.

Checked before merging: I reproduced your finding on `main` first — `discover -s tests` without `-t` fails the `test_log_isolation` tripwire, `tests/__init__.py` never sets `RESOLVE_MCP_LOG_FILE`, and `logs/server.log` grew during the run — then confirmed on your tree that the same invocation passes and creates no server log. `main` merged into your branch in a scratch worktree ran the full suite clean (3742 with #246 alongside, only the known worktree path artifacts). The sort-order assertion is the part I am most glad to have: without it the fix would be one well-named file away from silently undoing itself. Your CI run was parked as "action required" because GitHub holds first-time contributors' workflows; I approved it.

Nothing changed on landing beyond the changelog entry, which says plainly which runner was exposed and which was not.

PR #246, opened 2026-09-17
fix(guards): read allow_non_mcp_name="false" as false in both name guards

Dev-next-gen:
After v4.7.3 I went back over the opt-in flags with a different search than the one on #240. That one looked for `not p.get("<flag>")` at the guard; `allow_non_mcp_name` never shows up that way, because the eight places that read it pass the raw value into a helper, and the helper does the truthiness test. So it slipped past both of us.

The flag lifts the `_mcp_` name guard that `project_manager` and `extension` use to keep the MCP away from things it did not create. `_require_disposable_project_name` and `_extension_safe_name` both open with `if allow_non_mcp_name:`, so `allow_non_mcp_name="false"` is truthy and the guard stands down. From a client that stringifies its JSON scalars:

- `safe_project_delete` with `name="Client Feature Final"` and `allow_non_mcp_name="false"` calls `DeleteProject` on that project. As long as it is not the open one, nothing else is in the way. It's the same shape as `close_current` in #240: the caller spells out the refusal and the delete goes ahead anyway.
- `safe_install_extension` and `safe_remove_extension` accept a user-owned Fuse, DCTL or script name. The remove path still has the `require_marker` check behind it, so the delete there is less exposed, but the name guard itself is open.
- The same reading also covers `safe_project_create`, `_export`, `_import`, `_archive` and `_restore`.

The fix goes in the two helpers, not at the eight call sites, and reads through `coerce_bool` from `src/utils/bool_params.py`, the same helper as in #239 and #240. Both helpers are private, and the only thing that changes is how a string is read. `True`/`False` go through `bool(value)` as before, and `"true"`/`"yes"`/`"on"`/`"1"` still lift the guard. `"false"`/`"no"`/`"off"`/`"0"` now keep it. A string that is neither now falls back to the default, which is off, so the guard stays up. The only callers who see a different result are the ones who were getting the guard lifted without asking for it. They now get the refusal that already exists, with the same message. I annotated the parameter as `Any` because that is what the call sites actually pass.

`tests/test_allow_non_mcp_name_string_false.py` covers both helpers directly over six false spellings, `safe_project_delete` against a project manager stub, and `safe_install_extension` in `dry_run`. The delete case checks that `DeleteProject` was never called, not just that the response has an error. It also checks that the six true spellings still lift both guards. With `main`'s `src/server.py`, 24 of 24 false-spelling subtests fail. With this branch, all 6 tests pass.

I don't have pytest on this Windows machine, so I ran the full suite with `python -m unittest discover -s tests -t . -p "test_*.py"` (3739 tests) with and without the fix and compared the lists of failing tests. The only differences are the 24 subtests above and `test_panel_serves_on_ipv6_loopback`, which errored in the run without the fix and passed with it. That's the panel's first `/api/boot` timing, not this diff. Everything else that fails here fails the same way on both sides and comes from what this machine lacks (numpy, pillow, ffmpeg). The drift guards CI runs as unittest pass.


samuelgursky:
Merged as-is and released as [v4.7.8](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.8). Thank you — and thank you for saying plainly that it slipped past both of us. The #240 search was shaped around `not p.get(...)` at the guard, and a flag that is only ever tested inside a helper is invisible to that shape. Fixing the two helpers rather than the eight call sites was the right cut, and annotating the parameter as `Any` because that is what actually arrives is the honest signature.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3742 with #245 alongside, only the known worktree path artifacts). The delete test asserting `DeleteProject` never ran, against a stub where another project is open so `close_current` is out of the way, is the assertion that matters, and it is the one you wrote.

Nothing changed on landing beyond the changelog entry. Seven releases from your pipeline this week, each with a test that fails on the old code.

PR #247, opened 2026-09-17
fix(gallery): refuse a negative still, album or item index

Dev-next-gen:
I was looking at how the granular server resolves a timeline item and noticed that `_get_timeline_item` in `src/granular/common.py` still bounds the index with `item_index >= len(items)` only. EX5 (3a48d00) fixed exactly this in the compound `_get_item`, because Python reads `items[-1]` as the last clip, but the granular twin never got the lower bound, and 85 granular tools go through it. So `item_index=-1` on any of them acts on the last clip of the track instead of returning "No item at index -1".

The same one-sided check is in every album and still lookup of both gallery surfaces: `gallery` and `gallery_stills` in `src/server.py`, and the granular tools in `src/granular/gallery.py`. The sharpest case is deleting stills. In both `gallery_stills.delete_stills` and `delete_stills_from_album`, `still_indices=[-1]` deleted the album's last still, and an index past the end was silently dropped while the others were deleted and the call still reported success.

The fix adds the missing `< 0` bound in `_get_timeline_item` and in the granular gallery lookups, and a small `_index_in_range` check in the compound gallery tools. For the two delete paths, any index that is not a 0-based position in the album now refuses the whole call, the same way `ti_copy_grades` handles its target indices, instead of deleting a set nobody asked for. Valid indices behave exactly as before; the only inputs whose outcome changes are negative or out-of-range indices, which were never documented as meaning anything (`docs/SKILL.md` and the tool docstrings describe them as 0-based). The one visible change for a caller is that `delete_stills` with a partly invalid list now returns an error rather than a partial delete.

`tests/test_negative_index_refused.py` drives the real tool bodies against fake gallery and timeline objects, no Resolve needed. On `main` (6532370) 16 subtests fail and the three "valid indices still delete" tests pass; with the patch all 12 tests pass. I also ran the full suite on Linux at both commits: 3754 tests, `failures=16, errors=5` before and `errors=5` after, the 5 errors being the same pre-existing `drt` offline-fallback ones on both sides.


samuelgursky:
Merged as-is and released as [v4.7.9](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.9). Thank you — the granular `_get_timeline_item` gap is the one that matters most here, because 85 tools sit behind it and the compound twin had been fixed long enough ago that everyone assumed both were. And refusing the whole delete on a partly invalid list, rather than quietly deleting the valid part, is the right call for the same reason `ti_copy_grades` does it.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3754, only the known worktree path artifacts), and I grepped the tree for one-sided `>= len(` bounds after your change — the only one left is an internal loop in an audio analyzer, not an index a caller supplies, so the class is closed on the tool surface.

Nothing changed on landing beyond the changelog entry. Eight releases from your pipeline this week.

PR #248, opened 2026-09-17
fix(sync): report drop-frame sync events at the timecode they happen

Dev-next-gen:
I was reading the sync-event detector's timecode reporting and noticed the two
legs of the conversion disagree.

`_timecode_for_event` reports an event at `start_timecode + offset`, where
`start_timecode` is whatever ffprobe read off the media's own timecode track
(`format.tags.timecode`, falling back to the stream tag). On an NTSC
deliverable that tag is routinely drop-frame, spelled `HH:MM:SS;FF`.

`timecode_to_frames` in `multicam.py` honours drop-frame — `01:00:00;00` at
29.97 is frame 107892, not 108000, and `tests/test_marker_params.py` already
pins that. `_frames_to_timecode` in `sync_detection.py` had no drop-frame
arithmetic at all, so the frame numbers subtracted on the way in were never
added back. The head of a drop-frame clip came back as `00:59:56:12`:

```
>>> _timecode_for_event(0.0, 29.97, "01:00:00;00")
'00:59:56:12'          # 3.6 seconds early, and now non-drop
>>> _timecode_for_event(0.0, 59.94, "01:00:00;00")
'00:59:56:24'
```

That number is not only in the report. `_event_marker_suggestion` appends it to
the marker note (`" Timecode: {...}."`), so it gets written into the Resolve
project — and lining two cameras up off a 2-pop is exactly what someone reads
that note for.

I fixed it at the source rather than at the callsite. The inverse now lives in
`multicam.py` next to `timecode_to_frames`, as `frames_to_timecode`, so the
pair cannot drift apart again: a frames-to-timecode helper written on its own
is precisely what forgets that drop-frame dropped frame *numbers*, which is how
this happened. `sync_detection` carries the drop-frame spelling of the start
timecode through to it, and its duplicate `_nominal_timecode_rate` goes with
the old helper.

Non-drop timecode is deliberately untouched: 29.97 colon timecode legitimately
lags the wall clock, so frame 107892 stays `00:59:56:12`, and a semicolon at
23.976 still drops nothing on either leg because drop-frame is only defined at
nominal 30 and 60. Both are pinned in the new tests.

Proof, on the offline suite:

```
# before, new test file on pristine source
AssertionError: '00:59:56:12' != '01:00:00;00'
AssertionError: '00:59:56:24' != '01:00:00;00'
FAILED (failures=22, errors=2)

# after
python3 -m unittest tests.test_sync_event_timecode -v
Ran 8 tests in 0.001s
OK
```

`tests.test_multicam_setup tests.test_sync_detection tests.test_marker_params
tests.test_resolve211_multicam tests.test_dead_space_markers
tests.test_static_undefined_names tests.test_duplicate_definitions
tests.test_sync_event_timecode` — 66 tests, OK (1 skipped). The round-trip
tests walk every minute of the ten-minute drop-frame pattern and both NTSC
rates, so the arithmetic is checked at the boundaries where it is easy to get
wrong, not just at one hour mark.


samuelgursky:
Merged as-is and released as [v4.7.10](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.10). Thank you — this one is the kind of bug that survives a long time because the number it produces looks like a timecode, and the diagnosis of *why* it happened (an inverse written on its own forgets that drop-frame dropped frame numbers) is exactly why moving it next to the forward conversion is the right fix rather than patching the callsite.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3762, only the known worktree path artifacts). Then, because the arithmetic is the whole question, I brute-forced `frames_to_timecode` against `timecode_to_frames` on your tree for every frame in two hours at 29.97 (216,000) and 59.94 (432,000): zero round-trip violations, every rendered timecode unique, and frames 00/01 (or 00–03) never at the top of a non-tenth minute. The hour mark lands on 107892, which is the number this repo measured against Resolve for the nominal-base rule, so the two agree. Your point about leaving non-drop colon timecode alone is right and the changelog says so.

Nothing changed on landing beyond the changelog entry. Nine releases from your pipeline this week.

Dev-next-gen:
Thank you for the brute force on this one. Checking 648,000 round trips instead of taking the table on trust, and landing on the same 107892 this repo had already measured against Resolve, is a stronger validation than an approval could ever be. It is also the whole argument for arriving with a proof rather than with a claim.

You should know that I quote you, so you hear it from me and not by stumbling on it. I keep a public record of what this pipeline produces and how projects receive it, refusals included, and your own line is in it under your name: https://github.com/Dev-next-gen/defect-hunting-pipeline and https://github.com/Dev-next-gen/defect-hunting-pipeline/blob/main/docs/reception.md. "Nine releases from your pipeline this week" is your sentence, not mine, which is precisely why it says more than anything I could write about my own work. What you did before merging is written down next to it.

If you would rather not be quoted, or want a word changed, tell me and it comes out the same day, with no explanation needed.

Leo CAMUS

(@Dev-next-gen)

PR #249, opened 2026-09-18
fix(guards): read allow_partial_item_delete="false" as false

Dev-next-gen:
This is the same reading mistake as #237 (`ripple="false"`), #239 (`overwrite="false"`) and #246 (`allow_non_mcp_name="false"`), on the flag that decides whether a range delete may take whole clips.

`_timeline_lift_range_impl` collects every item the requested range touches, then blocks the ones the range only partially covers unless the caller opted in:

```python
allow_partial = bool(p.get("allow_partial_item_delete", p.get("allowPartialItemDelete", False)))
```

A bare `bool()` makes the string `"false"` True, so the opt-in is considered given and the block never happens. For a caller who passed `allow_partial_item_delete="false"` to protect exactly that case, a clip spanning frames 0–48 is deleted whole by a `lift_range(0, 24)` — and the response says `{"success": true, "deleted": 1}` with no `blocked` list, so nothing in it suggests the guard was bypassed. `"off"`, `"no"` and `"0"` behave the same way. `timeline.apply_cuts` reads the same flag off its own params at line 25859 with the same `bool()`, and passes the result down to every cut in the list.

What made me look here is that `ripple` is coerced through `coerce_bool` about twenty lines below the first site, inside the same function — the file already knows the right way to read this, the two flags just diverged.

The fix routes both reads through `src/utils/bool_params.coerce_bool`, which is what the three earlier fixes used:

```python
allow_partial = _coerce_bool(p.get("allow_partial_item_delete", p.get("allowPartialItemDelete")))
...
allow_partial = _coerce_bool(p.get("allow_partial_item_delete"), True)
```

Both keys are still read at the `lift_range` site. `apply_cuts` keeps its `True` default, now expressed as the helper's default argument so an absent flag still means "allowed". Only strings in `coerce_bool`'s false set change meaning; real booleans still go through `bool(value)` unchanged, and an unrecognised string falls back to the same default the call had before.

Proof is offline, no Resolve needed — `tests/test_lift_range_allow_partial_string.py` drives `_timeline_lift_range_impl` and `timeline("apply_cuts", ...)` against a fake timeline holding one 0–48 item. Before the change, 5 of its 7 tests fail (the string-false cases delete the clip); after, all 7 pass. The neighbouring suites stay green: `test_cut_executor`, `test_cut_ir`, `test_delete_clips_ripple_string`, `test_delete_clips_readback_retry`, `test_action_help`, `test_validation_guards`, `test_keyed_param_guards`, `test_install_overwrite_string_false` — 90 tests, OK.

One thing I left out on purpose. `lift_range` accepts the camelCase alias `allowPartialItemDelete`, `apply_cuts` does not, so an opt-out sent to `apply_cuts` in camelCase is dropped silently and its `True` default applies. That looked like a design call for you rather than part of a coercion fix, since the alias is not documented for either action, so I did not touch it.

I also have a short list of other flags in this family that are still read raw — `delete_after` in `gallery.grab_and_export`, the three `dry_run` reads whose default is True (a `dry_run="false"` leaves the caller in a permanent dry run, while `apply_cuts` tells them to "Re-run with dry_run=false"), and `allow_negative_record_frame` in `src/utils/multicam.py`. Happy to open that as an issue as you suggested on #240, or send them separately.


samuelgursky:
Merged as-is and released as [v4.7.11](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.7.11). Thank you — and sorry for the delay: the merged-worktree run on my side hit a full disk and produced thirty-two `disk I/O error`s that had nothing to do with your two lines. With space freed, `main` merged into your branch ran the full suite clean (3769, only the known worktree path artifacts, zero disk errors), and I confirmed every other flag read left in `src/server.py` is a pass-through to a helper that already coerces.

Your two calls were both right. The `apply_cuts` camelCase alias is a design question and I left it alone, and said so in the changelog. And yes, please send the remaining raw reads — `delete_after` on `grab_and_export`, the three `dry_run` reads whose default is `True`, and `allow_negative_record_frame` in `multicam.py` — as one PR rather than an issue; the `dry_run="false"` permanent-preview case is worth fixing on its own merits, and a PR with a failing-on-old-code test lands the same day here.

Nothing changed on landing beyond the changelog entry. Ten releases from your pipeline in five days.

PR #252, opened 2026-09-19
fix(media_pool): an unresolved clip id fails the whole clip batch

denvital:
## Summary

`media_pool` `delete_clips`, `move_clips`, `relink` and `unlink` resolved `clip_ids` with

```python
clips = [_find_clip(root, cid) for cid in p["clip_ids"]]
clips = [c for c in clips if c]
```

so an id that matched no clip was dropped and Resolve was handed the rest:

- `delete_clips` errored only when **every** id missed (`"No clips found"`). A mixed batch deleted the subset that resolved and answered `{"success": true}`. With confirm tokens on, the preview and the token covered only that subset.
- `move_clips`, `relink` and `unlink` never checked for an empty result. A batch where every id missed reached `MoveClips([], target)` / `RelinkClips([], path)` / `UnlinkClips([])` and reported whatever Resolve's bool said.
- A bare string was iterated character by character, one id per character.

## Change

A new `_clips_from_ids(root, clip_ids, *, verb)` next to `_find_clip` checks every id before anything reaches Resolve, and before a confirm token is issued:

- any unresolved id → `CLIP_NOT_FOUND` / `invalid_input`, with `error.state.unresolved_clip_ids` and `error.state.resolved_clip_ids`; nothing is deleted, moved, relinked or unlinked
- empty or non-list `clip_ids` → `INVALID_CLIP_IDS`
- missing key → still the surface-wide `MISSING_CLIP_IDS` (`p["clip_ids"]` stays an item lookup)

The tool docstring and `docs/SKILL.md` now say these four actions are all-or-nothing.

## Tests

- `tests/test_media_pool_clip_ids.py`: a fake Media Pool records every mutation call. It covers the partial batch and the all-missing batch for each action, a partial `delete_clips` under token gating (no token issued), the missing / empty / bare-string shapes, and the happy path with a nested clip. **Against the pre-fix code 5 of its 10 tests fail (17 subtests); all 10 pass after.** The 5 that pin unchanged behaviour pass on both.
- Offline suite: 3,721 run, 84 skipped, 11 errors. All 11 come from the machine the suite ran on: `numpy` and `requests` are missing from the venv, and `node_modules` is not installed, so `jszip` is unavailable. None of them touch `media_pool`.
- Drift guards, `audit_api_parity`, `gen_api_limitations --check`, `audit_readwrite_symmetry --check` and `agent-rules/generate.mjs --check` are clean. `test_static_undefined_names` was skipped because pyflakes is not installed.

`tests/live_clip_ids_check.py` ran against a real project with 144 folders and 2,600 clips. The check never changes anything. The MediaPool it hands to the actions forwards only `GetRootFolder` / `GetCurrentFolder`. Every other method is intercepted and never reaches Resolve, so nothing can be deleted, moved, relinked or unlinked, whatever the code does. Token gating is forced on and no token is passed. The destructive hook does not write to the analysis root, and its logs go to a temp directory.

- **This branch: PASS.**
  - Partial and all-missing batches return `CLIP_NOT_FOUND` for all four actions, and no token is issued for the partial delete.
  - The input-shape errors come back as expected.
  - A whole batch still resolves to the real clip and target folder.
  - After the run, every clip's folder, every folder's parent and the probe clip's File Path are unchanged.
- **Pre-fix code: FAIL with 18 findings.**
  - The check intercepted 10 MediaPool mutations the old code tried to send. The partial batches sent `MoveClips` / `RelinkClips` / `UnlinkClips` with only the real clip, and the all-missing batches sent the same three calls with an empty list.
  - The partial delete was issued a confirm token.
  - The pool was unchanged after this run too.

## Not changed (possible follow-up)

The same drop-the-misses pattern is still in `create_timeline_from_clips` (simple mode), `append_to_timeline` (legacy `clip_ids` mode), `export_metadata` and `auto_sync_audio`. When every id misses, `export_metadata` calls `ExportMetadata(path, [])`, and what Resolve does with an empty list has not been measured. The granular `delete_media_pool_clips` / `move_clips_to_folder` also act on partial batches. They are listed in the CHANGELOG entry.

## Release

Per `docs/process/release-process.md`, this bumps every version surface to **4.8.2** and regenerates `package-lock.json` with `npm install --package-lock-only`. Renumber as you see fit when landing. No tag or GitHub Release was created.


samuelgursky:
Merged as-is and released as [v4.8.2](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.2). Thank you, and welcome — this is about as complete as a first contribution gets. The version bump matched the release process to the letter, the changelog entry is in the house format with a Not changed section that names the rest of the pattern honestly, and the tripwire harness is the right instrument for a destructive path: it proves the old code *tried* to send ten mutations without letting any of them land.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3807, only the known worktree path artifacts); the parity audit, the symmetry audit, the ratchet and the release-surface drift guard all pass with your bump, so the merge commit is the release commit. Your live run on Studio 21.1.0.14 lands as your measurement in the release notes. Your CI run was parked as "action required" because GitHub holds first-time contributors' workflows; I approved it and it went green.

Nothing changed on landing. The follow-up list you wrote — `create_timeline_from_clips`, `append_to_timeline`, `export_metadata` (and what `ExportMetadata(path, [])` actually does), `auto_sync_audio`, and the two granular tools — is yours if you want it, one PR for the set is fine, and a measurement of the empty-list case would be welcome either way.

PR #254, opened 2026-09-19
fix(media_pool): folder ids resolve at any depth; unresolved ids fail the batch

denvital:
> **Stacked on #252.** This branch contains #252's commit (`2745857`, the clip-id fix) followed by one new commit, `dc22aee`. Please review and merge #252 first. Only `dc22aee` is new here. If #252 lands as-is, this PR's diff reduces to that commit. If it is squash-merged, I will rebase this branch onto `main`.

## Summary

`media_pool.delete_folders` and `media_pool.move_folders` had two defects.

1. **Nested folder ids never resolved.** Both actions scanned only `root.GetSubFolderList()`. Any id below the first level came back `{"error": {"message": "No folders found"}}`, even though the caller held the id that `GetUniqueId()` / `folder get_subfolders` had just returned. This was reproduced on Resolve Studio 21.1.0.14 against `Master/OUTDOORS/1_FOOTAGE/<bin>`: the action failed, while `mp.DeleteFolders()` on the recursively resolved folder object worked.
2. **Unresolved ids were dropped.** Only an all-empty result was an error. A mixed batch deleted or moved the subset that resolved and answered `{"success": true}`. `move_folders` did not even stop on an empty result, and answered success after `MoveFolders([], target)`.

## Change

A shared `_folders_from_ids(root, folder_ids, *, verb)` uses the existing recursive `_find_folder_by_id` walk. It applies the same all-or-nothing rule that #252 applies to the clip actions:

- any unresolved id → `FOLDER_NOT_FOUND` / `invalid_input`, with `unresolved_folder_ids` and `resolved_folder_ids` in `error.state`. The check runs before anything reaches Resolve and before a confirm token is issued.
- empty or non-list `folder_ids` → `INVALID_FOLDER_IDS`; a missing key keeps `MISSING_FOLDER_IDS`
- the Master folder, reachable now that the search is recursive, → `ROOT_FOLDER_NOT_ELIGIBLE`
- `move_folders`' "Target folder not found" now carries `FOLDER_NOT_FOUND` / `invalid_input`

The tool docstring and `docs/SKILL.md` are updated.

## Tests

- `tests/test_media_pool_folder_ids.py` covers a three-level-deep folder, the partial-batch case, Master, and the missing / empty / bare-string inputs. **Against the #252 code, 10 of its 11 tests fail; all 11 pass after.**
- Offline suite: 3,732 run, 84 skipped, 11 errors. The errors are the same environment gaps as in #252 (`numpy`/`requests` missing from the venv, no `node_modules`) and none of them touch `media_pool`. Drift guards, api-parity, api-limitations, read/write symmetry and `generate.mjs --check` are clean.

`tests/live_nested_folder_ids_check.py` ran against a real project with 144 folders and 2,600 clips. It probed a folder **seven levels below Master**. The check changes nothing and uses the same tripwire as #252's clip check. The MediaPool handed to the actions forwards only `GetRootFolder` / `GetCurrentFolder`. `DeleteFolders`, `MoveFolders` and everything else are intercepted and never reach Resolve. Token gating is forced on. The move probes aim at the folder's current parent. The destructive hook does not write to the analysis root, and its logs go to a temp directory.

- **This branch: PASS.**
  - `delete_folders` on the depth-7 id stops at `confirmation_required` and names exactly that folder.
  - `move_folders` resolves the folder and its current parent as the target.
  - Partial and all-missing batches return `FOLDER_NOT_FOUND`, Master returns `ROOT_FOLDER_NOT_ELIGIBLE`, and malformed input returns `INVALID_FOLDER_IDS` / `MISSING_FOLDER_IDS`.
  - Every folder's parent and every clip's folder read back identical after the run.
- **#252 code (without this fix): FAIL with 14 findings.**
  - The reported bug reproduces: the depth-7 id comes back `No folders found`.
  - The tripwire intercepted five `MoveFolders` calls with an empty folder list.
  - The pool was unchanged after this run too.

## Release

Bumps every version surface to **4.8.3**, on top of #252's 4.8.2, and regenerates `package-lock.json` with `npm install --package-lock-only`. Renumber as you see fit when landing. No tag or GitHub Release was created.


samuelgursky:
Merged as-is and released as [v4.8.3](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.3). Thank you — two in one morning, and the stacking worked exactly as you set it up: #252 landed with a merge commit, so this branch reduced to your one new commit and merged clean on top of it, bump and all.

Checked before merging: `main` (at v4.8.2) merged into your branch in a scratch worktree ran the full suite clean (3818, only the known worktree path artifacts); the parity audit, the symmetry audit, the ratchet and the release-surface drift guard pass with your 4.8.3 bump, so again the merge commit is the release commit. Your depth-7 live reproduction on Studio 21.1.0.14 lands as your measurement in the release notes. The Master-folder refusal is a good catch: the recursive walk made it reachable for the first time, and refusing it before Resolve sees it is the right shape. Your parked CI run is approved.

Nothing changed on landing. The follow-up list from #252 still stands whenever you want it.

PR #251, opened 2026-09-19
feat(project_manager): add read-only snapshot action

tpellet:
Add `project_manager(action="snapshot")`, a read-only action that returns the state an agent reads before it plans an edit: `project`, `timeline` (per-track items), `gaps_overlaps`, `render` and `media_pool` counts.

**Why.** In 8,407 mined agent tool turns from one real project, 2,222 were state inspection. Agents call `get_current`, `probe_timeline_structure`, `get_items`, `detect_gaps_overlaps` and the render status actions one LLM turn at a time. One call saves those turns.

**How.** It composes the existing helpers, so values match what those actions already return. No new traversal code. It never switches page, timeline or folder.

- `include` picks sections. Unknown names and an empty list are refused.
- `item_limit` (default 200) caps returned items. Counts and gaps still cover the whole timeline.
- A failing section reports `{error}` in its place. The others still return.
- It saves turns and response size, not read time. The timeline sections cost what `probe_timeline_structure` costs, and `media_pool` walks every clip. The docs say so.

**Testing.** 16 new tests. Full suite: 3780 passed, 22 skipped. Parity audit and drift guards pass. Tested against stubs only. Not yet run against a live Resolve.

**Open to your preference.** The name (`snapshot` vs `probe_project_state`, which the read-only classifier would recognise), and a CHANGELOG entry.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BdwSFTKitGV3ncsmBqpN8w

PR #255, opened 2026-09-19
fix(tests): keep the offline suite away from Resolve through the granular server

denvital:
## Summary

With DaVinci Resolve open, the offline suite connected to it, and the offline guard did not stop it.

- `src/granular/common.py` ran `import DaVinciResolveScript` and `connect_resolve()` at **import time**. Any test that imported a granular module therefore called `scriptapp("Resolve")` on the real `fusionscript.so`.
- `tests/offline_guard.py` only swapped `_launch_resolve`, `get_resolve` and `resolve_is_running` on `src.server`. The granular `get_resolve()` still fell through to its own `_launch_resolve()`.
- The compound server leaked the same way. Before tool calls, its execution-lifecycle state provider calls `src.server._try_connect()` directly, and the guard never swapped that function.
- Whether the real module loaded at all depended on import order. A few test modules `sys.modules.setdefault()` a stub, which only wins when they run first. `tests/test_live_api.py`, which pytest collects, calls `scriptapp` whenever the real module wins.

Measured with a full `python -m unittest discover -s tests -t .` run of v4.8.4, behind a tripwire standing in for the native library:

- The test process called `scriptapp("Resolve")` **1,165 times**. 1,162 of those came from `_get_resolve_lifecycle_state`.
- It made **one attempt to `open` the application**, through the granular `ResolveProxy` → `get_resolve()` → `_launch_resolve()`, from `test_granular_destructive_op`.

**On this branch the same run records 0 and 0.**

## Change

**Guard** (`tests/offline_guard.py`):

- Before it imports `src.server`, the guard installs a `sys.meta_path` finder that answers `DaVinciResolveScript` and `fusionscript` with an empty stub.
  - It is a finder rather than a `sys.modules` entry, so a test that pops the module cannot let the next import reach the real library.
  - The stub has no `scriptapp`. `connect_resolve()` therefore raises before its bridge fallback instead of falling through to it.
- `_try_connect`, `_launch_resolve` and `get_resolve` in `src.granular.common` are swapped in the same way as the compound server's.
  - The swap covers **every** granular module that holds these functions. `src/granular/__init__.py` imports each tool module, and each one binds its own copy through `from src.granular.common import *` before the guard can swap `common`.
  - The originals stay reachable as `_*_unpatched`.
- For the run, the in-app bridge client points at a config file that does not exist. `DAVINCI_RESOLVE_BRIDGE=1` in a developer's shell therefore cannot open a socket to a bridge running inside Resolve.
- Installation stays idempotent under unittest (`tests/__init__.py`, `tests/test_0000_offline_bootstrap.py`) and under pytest (`conftest.py`).

**Runtime** (`src/granular/common.py`):

- The module no longer connects at import. It still imports DaVinciResolveScript, with the same diagnostics when that fails.
- The connection moved to `connect_at_startup()`. `src/resolve_mcp_server.py` and `src/server.py --full` call it right after importing the package.
- Starting the granular server behaves as before: it connects to a Resolve that is already open, logs it, and never launches one. Launching stays with `get_resolve()` on the first tool call.
- Code that only imports the package no longer talks to Resolve.

## Tests

- New `tests/test_offline_guard_granular.py` puts a module shaped like Blackmagic's loader first on `sys.path`, and first checks that a path import would load it. It then asserts:
  - neither a fresh import of `src.granular.common` nor `connect_at_startup()` ever executes that module
  - both module names resolve to the stub
  - `connect_resolve(stub)` never reaches the bridge fallback
  - every granular module holds the stand-ins, and the originals stay reachable
  - the bridge has no config to connect with
- **With the finder disabled, it fails with `['imported DaVinciResolveScript', "scriptapp ('Resolve',)"]`.** In that same run, installing the guard alone called `scriptapp('Resolve',)` through the old import-time connect.
- `tests/test_0000_offline_bootstrap.py` now also asserts that the finder was in place before `src.server` imported, and that the granular swap is installed.
- Full suite (`discover -s tests -t .` and `discover -s tests`): 3,749 run, 84 skipped, 11 errors. All 11 also error on v4.8.4 on this machine: `numpy` and `requests` are missing from the venv, plus `test_offline_fallback` / `test_lut_file_controls`. pytest is not installed locally, so the pytest path was not run here. It calls the same idempotent `install()`.
- Drift guards, `test_import`, `audit_api_parity`, `gen_api_limitations --check`, `audit_readwrite_symmetry --check`, `agent-rules/generate.mjs --check`, `npm pack --dry-run` and `git diff --check` are clean.

## Not covered (follow-ups)

- **Control-panel children:** `test_control_panel_ipv6_loopback` and one `test_open_control_panel` case start the real `src/analysis_dashboard.py` as a child process. The in-process guard cannot reach it, and the child still calls `scriptapp("Resolve")` read-only: 5 calls from 3 children per full run, the same on v4.8.4 and on this branch.
- **Trace log:** a full run writes about 5,500 synthetic lines into the repo's `logs/execution-traces.jsonl`.

Both launchers were started from this branch with Resolve already open. Each logs `Connected to DaVinci Resolve: DaVinci Resolve Studio 21.1.0.14`, and neither started anything:

- `src/resolve_mcp_server.py` logs the line before `Starting DaVinci Resolve MCP Server v4.8.5 (389 granular tools)`, where the import-time connect used to log it.
- `src/server.py --full` logs it right after the granular import, before `Threaded tool dispatch installed for 389 tools`.

Not measured: startup with Resolve closed. That path logs `Failed to get Resolve object` and leaves launching to the first tool call. The offline suite was never run against the live Resolve. Recorded in the CHANGELOG by 587090e.

## Release

Per `docs/process/release-process.md`, this is a separate release commit that bumps every version surface to **4.8.5** and regenerates `package-lock.json` with `npm install --package-lock-only`. Renumber as you see fit when landing. No tag or GitHub Release was created.


samuelgursky:
Merged as-is and released as [v4.8.5](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.5). Thank you — this is the most important of the three, and it corrects something this repo believed about itself. An August note in my own records said the offline runner had to keep `_try_connect` real or two bridge tests went red, which is exactly the hole your tripwire counted: 1,162 of the 1,165 `scriptapp` calls came through that function. And a suite that can *launch* the application from a `hasattr` scan is the mechanism behind a 23-minute hang recorded here in August that nobody had traced.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3835, only the known worktree path artifacts, zero disk errors), and the bridge tests that note worried about are green under your guard. Moving the connection out of import and into an explicit `connect_at_startup()` that both launchers call is the right runtime shape: starting the server behaves as before, and merely importing the package no longer talks to whatever Resolve is open. Your live check of both launchers on 21.1.0.14 lands as your measurement.

Nothing changed on landing beyond the release notes, which also record @tpellet's live run of `snapshot` from #251. The two follow-ups you list (control-panel children, the trace log a full run writes) are noted in the notes as not covered. #256 and #257 are next, in that order.

Issue #253, opened 2026-09-19
Follower follow-up: TransformSize and Softness1 refuse spline attach under add_keyframe


artpavelalex-ux:
### Steps to reproduce

1.

### Environment

| | |
|---|---|
| MCP server | |
| DaVinci Resolve |  |
| Connection |  |
| OS | |
| MCP client |  |

samuelgursky:
Fixed and released as [v4.8.4](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.4). Thank you for the follow-up. The report came through as the empty template with only its title, so I worked from the title — and the title was enough to reproduce it exactly.

**What is going on.** `TransformSize` and `Softness1` are not inputs you can animate; they are *nest controls*, the fold-down group headers the Fusion UI draws. Measured today on Studio 19.1.3.7 on the Follower your #250 fix creates:

| input | `INPID_InputControl` | `AddModifier(…, BezierSpline)` |
|---|---|---|
| `TransformSize` (shown as "Size") | `NestControl`, passive | False, also for Path and TextScramble |
| `Softness1` | `NestControl`, passive | False |
| `Size1` | `NestControl`, passive | False |
| `Size`, `Opacity1`, `Delay` | `SliderControl` | attaches |

TextPlus's own `Softness1` refuses the same way, so this is not specific to the Follower. The controls a header folds are the next entries in `GetInputList()` order (the header's `INPI_LabelControl_NumInputs` says how many), and *those* keyframe normally:

- `Softness1` → `SoftnessX1`, `SoftnessY1`, `SoftnessOnFillColorToo1`, `SoftnessGlow1`, `SoftnessBlend1`
- `TransformSize` → `LineSizeX`, `LineSizeY`, `WordSizeX`, `WordSizeY`, `CharacterSizeX`, `CharacterSizeY`
- `Size1` → `SizeX1`, `SizeY1`

**What changed.** `add_keyframe` and `add_modifier` now detect a nest control before touching Fusion and refuse it with **`FUSION_INPUT_IS_NEST_CONTROL`**, naming the folded controls in the remediation and in `error.state.nest_members`, so the next call can target the right input without guessing. Verified live through the real actions: the refusal listed exactly the members above on the Follower and on TextPlus, and `SoftnessX1`, `CharacterSizeX` and `Delay` keyframed and read back.

[...]

PR #256, opened 2026-09-19
fix(media_pool): every clip_ids batch is all-or-nothing; a refused append fails

denvital:
> **Two releases in one branch, on top of `main` (v4.8.4, #253).** `db38964` is v4.8.6 and `e49941c` + `4748575` are v4.8.7. **v4.8.5 is taken by the open #255** (offline guard), so whichever of the two lands second will conflict on the version surfaces and `CHANGELOG.md`. The code merges cleanly. I will rebase this branch if needed. Renumber as you see fit when landing. No tag or GitHub Release was created.

## Summary

1. **Every `clip_ids` batch is all-or-nothing (v4.8.6).** #252 fixed `delete_clips`, `move_clips`, `relink` and `unlink`. Four more `media_pool` actions still resolved `clip_ids` with `[_find_clip(...)]`, dropped every miss and handed Resolve the rest:
   - `create_timeline_from_clips` (simple `clip_ids` mode) built the timeline from the subset that resolved.
   - `append_to_timeline` (legacy `clip_ids` mode) appended the subset and verified it against the *resolved* count, so a partial append read back as `readback_verified`.
   - `export_metadata` exported the subset. An all-missing batch reached `ExportMetadata(path, [])`, whose behaviour on an empty list has not been measured.
   - `auto_sync_audio` synced the subset.
2. **An append Resolve refuses is reported as a failure (v4.8.7).** In legacy `clip_ids` mode, `append_to_timeline` answered `{"success": true, "count": 0}` when `MediaPool.AppendToTimeline` returned `None`, `False` or `[]`. Only `verified_operation.verification_status: "api_failed"` said otherwise. The granular `append_to_timeline` had the same answer.

## Change

**v4.8.6.** The four actions now use #252's `_clips_from_ids`. An unresolved id fails with `CLIP_NOT_FOUND` / `invalid_input` before anything reaches Resolve, with the unresolved and resolved ids in `error.state`. A bare string is `INVALID_CLIP_IDS`. `export_metadata` without `clip_ids` still exports every clip. An explicit `clip_ids: []` is now `INVALID_CLIP_IDS` instead of widening to a whole-pool export. The granular `append_to_timeline`, `auto_sync_audio`, `delete_media_pool_clips` and `move_clips_to_folder` get a granular `_clips_from_ids` with the same rule, in the granular server's flat error shape.

**v4.8.7.** Both `append_to_timeline` modes go through `_append_to_timeline_failed`:

- `APPEND_TO_TIMELINE_FAILED` / `resolve_api_failed`
- `error.reason` names what Resolve returned
- `error.state` carries `expected_item_count_delta` and `item_count_delta`
- `verified_operation` stays on the error, so the before/after timeline readback is not lost
- **`retryable` is true only when that readback shows the item count unchanged.** The category default would say retryable. But Resolve's answer does not prove nothing landed, and a retry after an append that did land puts the clips on the timeline twice. With no current timeline to read, it is not retryable.

`clip_infos` mode already errored, but as `UNSPECIFIED` with the readback dropped. It now shares the code and the rule, and its message is unchanged. The granular `clip_ids` form returns `{"success": false, "error": "Failed to append clip_ids to timeline"}`, like its own `clip_infos` form. The tool docstring and `docs/SKILL.md` are updated.

## Tests

- `tests/test_media_pool_clip_ids.py` now runs its cases over all eight compound actions. **Against the v4.8.2 code, 14 cases fail; all pass after.**
- The new `tests/test_granular_media_pool_clip_ids.py` covers the four granular tools. **8 of its 9 original tests fail on the old code; all pass after.**
- Five `FalsyAppendTest` cases (4 compound, 1 granular) use a fake Media Pool whose `AppendToTimeline` returns `None` / `False` / `[]`. They cover the error, the kept readback, the `retryable` rule (including items that land anyway and no current timeline) and `clip_infos`. **All 5 fail on the pre-fix code (7 subtest failures, 2 errors); all pass after.**
- Offline suite on this branch: 3,762 run, 0 failures, 84 skipped, 13 errors. The errors are environment gaps: `numpy`/`requests` missing from the venv, no `node_modules`, plus `test_live_api` / `test_resolve20_api`. Resolve was open on the machine, so the suite ran with `DaVinciResolveScript` blocked, the bridge config pointed at a missing file, and the granular launcher refusing. That is the gap #255 closes. The granular launcher was reached 10 times. Drift guards, api-parity, api-limitations, read/write symmetry and `generate.mjs --check` are clean. `test_static_undefined_names` was skipped because pyflakes is not installed.

`tests/live_clip_ids_check.py` ran against a real project with 144 folders, 2,600 clips and 24 timelines, probing a clip seven folders deep. The MediaPool handed to the actions forwards only `GetRootFolder` / `GetCurrentFolder`. Every other method is recorded, answered `False`, and never reaches Resolve. Token gating is forced on. Logs go to a temp directory.

- **v4.8.6: PASS.** Partial and all-missing batches returned `CLIP_NOT_FOUND` for all eight actions, with no MediaPool call attempted. Malformed input returned `INVALID_CLIP_IDS` / `MISSING_CLIP_IDS`. A whole batch handed the tripwire exactly the probe clip. That run was on the change applied to v4.8.2. The same run flagged the append answering `success=True` to the tripwire's `False`, which is what v4.8.7 fixes.
- **v4.8.7: PASS.** The whole-batch `append_to_timeline([probe])` now returns `APPEND_TO_TIMELINE_FAILED`, with `verification_status: "api_failed"` and `retryable` matching the readback. Its operation envelope is logged as `status: failed`. That run was on this stack on top of v4.8.3. v4.8.4 underneath only touches `fusion_comp`, and apart from #253's hunks, `src/server.py` is identical to what ran.
- After both runs, the folder tree, clip placement, the probe's File Path, the timeline count and the current timeline's items read back identical. No metadata file was written.
- Driven offline against a fake Resolve, the check reports 20 findings on the v4.8.2 code and 3 on the pre-v4.8.7 code (reproducing `success=True`).

## Release

`db38964` bumps every version surface to **4.8.6** and `4748575` bumps them to **4.8.7**. Each regenerates `package-lock.json` with `npm install --package-lock-only`. Per the release process, push the two tags one at a time.


samuelgursky:
Merged and released as [v4.8.7](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.7), carrying both of your entries; there is no separate v4.8.6 tag, and the release notes say so. Thank you — the append fix is the sharper of the two: a `success: true, count: 0` that only the verification block contradicted is the exact shape of lie this repo keeps hunting, and making `retryable` depend on the readback rather than on the error category is the right call for a call that can double-append.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3846, only the known worktree path artifacts), and the merged main after resolution ran clean again (3857). The resolution was exactly what you predicted: your code merged untouched; only the version surfaces and the changelog conflicted with #255's v4.8.5, and I set them to your v4.8.7 with both of your entries kept above the v4.8.5 entry. (I initially took two whole source files from your side, which would have dropped #255's runtime hunks — caught by a `connect_at_startup` grep, redone as a proper three-way merge, and verified: relative to your branch the only remaining source difference is #255's.) Your live check on 21.1.0.14 lands as your measurement.

`ExportMetadata(path, [])` stays unmeasured, as your notes say; the explicit-empty-list refusal means nothing reaches it now, which is the safer side to be on until someone measures it. #257 is next.

PR #257, opened 2026-09-19
fix(tests): fail a test that reaches a Resolve launcher

denvital:
> **Stacked on #255.** This branch contains #255's three commits. The two commits new here are 5d48670 (fix) and 19a0d29 (release v4.8.8). Merge #255 first; this diff then shrinks to those two.

## Summary

#255's guard answers both servers' launchers with a stand-in, but a test that reached one still passed.

- The stand-in appended to `LAUNCH_ATTEMPTS`. pytest printed that list in its summary without failing the run. `unittest`, the runner the release checklist uses, never read it.
- Every entry recorded under `unittest` read `<unknown test>`.

Measured on v4.8.4 with the granular `_launch_resolve` replaced by a counting stub, in one full `python -m unittest discover -s tests -t .` run:

- The launcher was reached **10 times**, all from **one test**: `test_granular_destructive_op.McpSchema.test_every_hooked_tool_advertises_the_override_and_nothing_else_new`.
- Its scan called `hasattr(value, "__granular_destructive__")` on every global of every granular module.
- Ten of those modules hold `resolve = ResolveProxy()` (timeline_item, timeline, folder, resolve_control, media_storage, graph, media_pool, gallery, media_pool_item, project). The proxy's `__getattr__` calls `get_resolve()`, which falls through to `_launch_resolve()` when nothing answers. That is one reach per module.
- With Resolve closed, each reach would have run `open` on the application; with Resolve open, the scan connects to it instead.
- No other test reached the granular launcher.

## Change

**A test that reaches a launcher fails** (`tests/offline_guard.py`, `tests/conftest.py`):

- The guard wraps `unittest.TestCase.run`, installed and uninstalled together with #255's swaps. Every test gets a cleanup that fails it when an attempt was recorded during the test, or since the previous test finished, which covers imports and class fixtures.
  - A cleanup, because unittest reports an exception in a cleanup as a failure of that test, and pytest's unittest integration drives the same `run`.
  - Each attempt fails exactly one test.
- Plain pytest functions get the same check from an autouse fixture in `conftest.py`.
- Each entry now names the test and the module that called the launcher, e.g. `tests.test_x.Case.test_y (called from src.granular.common)`.
- A test that calls the stand-in on purpose deletes its own entry and passes. `test_offline_guard_granular` already did this; only its comment changed.

**The test that reached it** (`tests/test_granular_destructive_op.py`):

- The scan reads the marker with `inspect.getattr_static`, so it never touches a proxy.
- New `test_the_module_scan_asks_for_no_connection` runs the scan with `get_resolve` raising. Under #255's guard, `get_resolve` answers `None` before any launch, so without this test a return of `hasattr` would go unnoticed.

## Tests

- New `tests/test_offline_launch_check.py` runs small inner tests and reads their results:
  - reaching the granular launcher through a `ResolveProxy` (the v4.8.4 path, with the guard's `get_resolve` put back to the real one) fails, labelled `… (called from src.granular.common)`
  - reaching the compound launcher through the real `get_resolve` fails, labelled `… (called from src.server)`
  - a reach from `setUpClass` fails the class's first test and not its second
  - a deliberate call that deletes its entry passes, and so does a test that reaches nothing
- **Negative controls**, run from a scratch driver without changing the repo:
  - With the `TestCase.run` wrapper removed, 4 of the 6 new tests fail.
  - The v4.8.4 `McpSchema` test, run under the check with the guard's `get_resolve` swapped back to the real one, fails with **10 recorded attempts**, each `…McpSchema.test_every_hooked_tool_advertises_the_override_and_nothing_else_new (called from src.granular.common)`.
  - With `hasattr` restored in the scan, the new pin test fails with `asked for a connection`.
- Full suite, `discover -s tests -t .` and `discover -s tests`: 3,756 run, 84 skipped, 11 errors. They are the same 11 errors as on #255 on this machine (`numpy` and `requests` missing from the venv, plus `test_offline_fallback` / `test_lut_file_controls`). No test failed the launch check, and `LAUNCH_ATTEMPTS` was empty at exit.
- Resolve was open on the machine, so every run went through a scratch `sitecustomize` tripwire:
  - `sys.modules["DaVinciResolveScript"] = None`
  - a recording finder for any later import
  - `DAVINCI_RESOLVE_BRIDGE_CONFIG` pointed at a file that does not exist
  - `Popen` of the application blocked

  It recorded no `scriptapp` call and no launch, in the test process or in any of its 59 children. The children were kept away by the tripwire's `None` entry, not by this change.
- Drift guards, `test_import`, `audit_api_parity`, `gen_api_limitations --check`, `audit_readwrite_symmetry --check`, `agent-rules/generate.mjs --check`, `npm pack --dry-run` and `git diff --check` are clean. `test_static_undefined_names` skipped: pyflakes is not installed locally.

## Not covered

- **pytest was not run locally** (not installed here), so the `conftest.py` fixture has only CI's `python -m pytest tests -q` to exercise it.
- **Control-panel children**, as noted in #255: the in-process guard cannot reach a child process.

No Resolve behavior changed, so no live Resolve run is required, and none was made.


samuelgursky:
Merged and released as [v4.8.8](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.8). Thank you — this is the piece that turns #255 from "the guard is in place" into "the guard is enforced": a launch attempt that only lands in a list nobody reads is exactly the kind of promise this repo's ratchet tests were written against, and failing the test that caused it, named, is the right instrument. Tracing all ten reaches to one `hasattr` scan over the proxies, then pinning the scan with a `get_resolve` that raises, is careful work.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3842, only the known worktree path artifacts), and the merged main after resolution ran clean again (3864) with no test failing the launch check. As with #256, your code merged untouched; only the version surfaces and the changelog conflicted with #256's v4.8.7 and were set to your v4.8.8, this time with the two source files resolved as proper three-way merges from the start. Relative to your branch, the only source difference on `main` is #256's hunks. Your negative controls land as your measurement in the release notes.

That is three releases from you today after two yesterday — thank you for taking the offline guard seriously enough to measure it rather than assume it.

PR #258, opened 2026-09-19
fix(tests): keep the offline suite's child processes away from Resolve

denvital:
> **Stacked on #257.** #257 is itself stacked on #255. This branch contains #255's three commits and #257's two. The two commits new here are dab0dfe (fix) and 2ceeecb (release v4.8.9). Merge #255 and #257 first; this diff then shrinks to those two.

## Summary

#255's guard works inside the test process: a `sys.meta_path` stand-in for Blackmagic's modules, plus swapped launchers. None of that reaches a child process. The control panels the suite starts are real children.

- `server._open_control_panel` starts `src/analysis_dashboard.py` with `subprocess.Popen`. The child imported Blackmagic's module and called `scriptapp("Resolve")` through `_connect_resolve_read_only`: at startup, in the inventory warm-up, and for `/api/boot`.
- Measured on v4.8.4 with a tripwire standing in for the library, one full `python -m unittest discover -s tests -t .` run made **6 calls from 3 panel children**.
- Two children came from `test_control_panel_ipv6_loopback`, whose docstring says "No Resolve". The third came from `test_tool_argument_validation`, which calls every action with no arguments, `open_control_panel` included. #255's notes attribute it to `test_open_control_panel`, but every test there stubs the port probe, and none of them starts a panel.

## Change

Test-only. No tool, action or runtime code changed.

- `offline_guard.install()` now puts `tests/offline_child_site` first on `PYTHONPATH`, on every install path. Every Python child that inherits the environment runs its `sitecustomize.py` at startup, before any code of its own. That file:
  - **runs the `sitecustomize` it shadows** first. Python loads only one, and Homebrew's Python ships one that rewrites `sys.prefix` and `sys.path`.
  - **answers `DaVinciResolveScript` and `fusionscript` with an empty stand-in** from a finder at `sys.meta_path[0]`, like the test process. It has no `scriptapp`, so `connect_resolve()` raises before its bridge fallback. The error names the guard, so a test-started panel now reports `Resolve connection failed: DaVinciResolveScript is the offline test guard's stand-in …` from `/api/boot`.
  - **refuses to launch the application through `subprocess`.** A child that imports `src.server` and calls a tool falls through `get_resolve()` to `_launch_resolve()` whenever Resolve is closed, and the in-process swap doesn't exist in a child. The check matches what `resolve_runtime.launch_command` builds on each platform, any other program inside an installation (`fuscript`), and `open`/`osascript` naming the app. For any other command it looks only at the program being run, so a Python child whose arguments mention the bundle path still runs.
- A developer's tripwire is usually a `sitecustomize` that chains to "the next one that is not me", which is now the guard. Before the guard tracked what had run, the two recursed into each other until `RecursionError`, the tripwire ended up answering the import, and the guard was never installed. The chain now continues past the guard, and the guard's stand-in goes in front last.
- #255's bridge-config redirect reaches children through the environment. That includes the doctor probe, which replaces `PYTHONPATH` and calls `connect_resolve(None)`.

## Validation

- New `tests/test_offline_guard_child_process.py`, 14 tests. The main one starts the real panel the way the suite does. Fakes shaped like Blackmagic's pair sit on `PYTHONPATH` behind the guard: a `DaVinciResolveScript.py` loader that hands over `fusionscript`, whose `scriptapp` records to a file. The test asserts that `/api/boot` names the stand-in and that neither fake loaded. A control child without the guard does reach the fake `scriptapp`. The other tests cover:
  - the stand-in again after a `sys.modules` pop
  - a shadowed `sitecustomize` still running
  - a chaining one reaching the file after the guard
  - a second checkout's copy not chaining in
  - the launch check against `launch_command` for three platforms and both modes
  - other launch forms, and ordinary children still running
  - a guarded child refusing a bundle-shaped program that does not exist
  - `PYTHONPATH` install and restore
- Mutations: with the `PYTHONPATH` export removed, the panel test fails ("DaVinci Resolve is not connected"), and the tripwire records 4 `scriptapp` calls from the panel child. With the chain tracking removed, both chaining tests fail.
- Full suite, `discover -s tests -t .` and `discover -s tests`: 3,770 tests each. The 11 errors are the same as on v4.8.8 in this environment (no `numpy`/`requests` in the venv, plus `test_offline_fallback` and `test_lut_file_controls`). Behind a scratch tripwire, no `scriptapp` call, native load or launch was recorded in the test process or in any child the tripwire could see, including 4 panel children per run.
- Release static checks from `docs/process/release-process.md` pass: `test_import` including the lockfile check, API parity, api-limitations and read/write symmetry `--check`, agent-rules `--check`, the drift guards incl. `test_release_surface_drift`, the Node CLI smoke tests, `npm pack --dry-run` and `git diff --check`.
- pytest is not installed locally, so the pytest path was not run here. CI runs it.
- No Resolve behavior changed, so no live Resolve run is required. None was made on purpose. The next section covers the one real connection these runs did make.

## Not covered, and one real connection during validation

A child that sets its own `PYTHONPATH`, uses `-E`/`-I`, or builds its environment from scratch never loads the guard. `os.system`/`os.exec*` launches are not intercepted.

One such child reached Resolve during this PR's validation:

- `install.verify_resolve_connection` replaces `PYTHONPATH` with the Blackmagic Modules directory.
- `test_scripting_lib_discovery.test_the_live_probe_agrees_with_the_summary` runs that probe for real whenever Resolve is installed.
- Resolve was open. So each of the four full runs most likely made one read-only connection through it (`scriptapp`, `GetProductName`, `GetVersionString`). The tripwire could not see that child.

This PR does not close that path. It needs a follow-up.


samuelgursky:
Merged as-is and released as [v4.8.9](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.9). Thank you — the `sitecustomize` chain is the detail that makes this robust rather than clever: Homebrew's Python really does ship one, and a guard that silently displaced it would have broken every child in a way nobody would trace back to the tests. Chaining to it first, and refusing to recurse into a developer's own tripwire, is the right shape.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3878, only the known worktree path artifacts), and the merged main again after the merge. Your correction of #255's attribution (the third child came from the argument-validation test, not from `test_open_control_panel`) is in the release notes as you stated it.

Nothing changed on landing. #259, #260 and #261 follow, in your numbering: #259's and #261's release surfaces conflict with the stack, as you said, and I will resolve those on `main` rather than ask you to rebase.

PR #259, opened 2026-09-19
fix(tests): refuse outbound requests from the offline suite

denvital:
> **Based on `main`, independent of #255, #257 and #258.** Released as v4.8.10 because 4.8.5–4.8.9 are claimed by #255–#258. Merging after those will conflict textually in `tests/offline_guard.py` (`install()` / `uninstall()`), `CHANGELOG.md` and the version files. Keep both sides; the network guard does not depend on the Resolve swaps.

## Summary

Every full offline run sent a live request to GitHub.

- `install.py` checks `https://api.github.com/repos/samuelgursky/davinci-resolve-mcp/releases/latest` for a newer version. `tests/test_scripting_lib_discovery.py` runs its `main()` in-process (`install.py --clients manual --dry-run`) through `_run_main`.
- So each `python -m unittest discover -s tests -t .` made that request. The run depended on the machine being online, and the release notes that came back were written into `logs/update-check.json`. Redirecting that file moves the answer, not the request.

Measured on v4.8.4 with a `sitecustomize` tripwire around `urllib.request.urlopen`, DNS lookups and socket connects: exactly **one** off-machine request in the whole run, from
`test_only_the_designated_live_test_touches_a_real_resolve` → `test_a_failed_verification_exits_non_zero` → `_run_main` → `update_check.check_for_updates` → `_fetch_latest_release`. The same run on this branch makes **none**.

## Change

**The offline guard refuses off-machine `urlopen`** (`tests/offline_guard.py`, `tests/conftest.py`):

- `urllib.request.urlopen` is wrapped. Loopback URLs (`127.0.0.0/8`, `::1`, `localhost`) and `file:`/`data:` URLs pass through, so the control-panel tests still serve and probe a real panel (`_control_panel_probe` goes through this same function).
- Anything else raises `NetworkRefused` before a socket is opened. That is a `URLError`, so callers take their no-network path; the update check reports `status: "error"`.
- Each refusal is appended to `NETWORK_ATTEMPTS` as `{url, caller, test}`, e.g. `src/utils/update_check.py:526 in _fetch_latest_release`. pytest lists them in its terminal summary, next to `LAUNCH_ATTEMPTS`.
- It is installed before `src.server` is imported and does not need it, so a checkout without the runtime stack is guarded too.
- Installing twice is a no-op. The marker lives on the wrapper in `urllib.request`, not in module state, so a second copy of `offline_guard` also sees it. `uninstall()` puts the original back.
- `test_offline_guard`'s test that calls `uninstall()` in the middle of a run now reinstalls the guard in a cleanup. Otherwise every later test would run without it.

**The test that made the request stops asking** (`tests/test_scripting_lib_discovery.py`): `_run_main` sets `DAVINCI_RESOLVE_MCP_UPDATE_CHECK=0`, so a clean run records no attempt at all.

## Tests

- New `tests/test_offline_network_isolation.py` (8 tests) is the tripwire. It fails if:
  - `urlopen` is not the guard (checked in `setUp`, before any URL is opened)
  - a remote URL gets through, as a string or a `Request`, or the refusal stops naming the calling line and the test
  - a loopback request to a real local server is stopped: `127.0.0.1`, `localhost`, `[::1]`
  - `check_for_updates` with its own defaults no longer ends at the guard with `caller` = `update_check.py … _fetch_latest_release`
  - `_run_main` reaches for the network again (both `--clients manual` and no clients)
  - installing again wraps the guard in itself, or `uninstall()` does not restore the original

  The refusal tests put a spy downstream of the guard, so a wrong verdict fails the test instead of sending a request.
- `tests/test_offline_guard.py`: the network guard installs even when `src` cannot be imported.
- **Negative controls**, by temporarily editing the working tree:
  - With the `_run_main` switch removed, `test_the_installer_tests_do_not_ask_for_the_network` fails. It names the URL, `src/utils/update_check.py:526 in _fetch_latest_release` and the test. The outer tripwire still saw 0 requests, because the guard refused it.
  - With the guard disabled, all 8 new tests fail in `setUp` before opening a URL.
- **Full suite**, `python -m unittest discover -s tests -t .`: 3,747 run, 85 skipped, 14 errors. They are the same 14 tests as on v4.8.4 on this machine, all environmental:
  - 6 need `numpy` or `requests`, which are missing from the venv
  - 5 in `test_offline_fallback` need `jszip`, since this worktree has no `node_modules`
  - 3 are refusals by the tripwire itself: two `test_doctor_paths` probe children, and the granular launch in `test_granular_destructive_op` that #255/#257 address
- Resolve was open on the machine, so every run went through a scratch `sitecustomize` tripwire, chained to Homebrew's own `sitecustomize`. It:
  - answers `DaVinciResolveScript`/`fusionscript` from `sys.meta_path[0]` with a stub whose `scriptapp` records and returns `None`
  - blocks `Popen` of the application, and of any child whose command line imports the scripting module
  - puts itself back on the `PYTHONPATH` of children given an explicit env
  - points `DAVINCI_RESOLVE_BRIDGE_CONFIG` at a file that does not exist
  - refuses off-machine `urlopen`, DNS and connects

  Off-machine requests went from 1 to 0. The `scriptapp` counts are unchanged, since this PR does not touch those paths: 1,165 from the test process in both runs, plus 5–6 from control-panel children depending on timing.
- `test_import` (including the lockfile sync), `audit_api_parity`, `gen_api_limitations --check`, `audit_readwrite_symmetry --check`, `agent-rules/generate.mjs --check`, the seven drift guards, `bin/davinci-resolve-mcp.mjs --help/--version`, `npm pack --dry-run` and `git diff --check` are clean. `test_static_undefined_names` skipped: pyflakes is not installed locally.

## Not covered

- **pytest was not run locally** (not installed here). The `conftest.py` summary lines are exercised only by CI's pytest run.
- **Only `urllib.request.urlopen` is guarded.** `http.client` or raw sockets to another host would not be refused. The outer tripwire saw none in this suite, and every `src/` caller uses `urlopen`.
- **`install.verify_resolve_connection`'s child.** It replaces `PYTHONPATH` with Resolve's `Modules` dir. On a machine with Resolve installed, `test_the_live_probe_agrees_with_the_summary` therefore starts a child that no `PYTHONPATH`-based guard reaches, and that calls the real `scriptapp("Resolve")`. Here the tripwire refused it and the test skipped. This PR does not change that; it's noted so it can be picked up separately.

No Resolve behavior changed, so no live Resolve run is required, and none was made.


samuelgursky:
Merged and released as [v4.8.10](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.10). Thank you — one request per run is small, but a test suite whose result depends on whether the machine is online, and which writes a real release note into the operator's logs, is the kind of thing that is invisible until a plane or a firewall makes it visible. Passing loopback through so the panel tests keep probing a real server, and raising a `URLError` subclass so callers take their genuine offline path, are both exactly right.

Checked before merging: your code was run against `main` in a scratch worktree, and the merged `main` after resolution ran the full suite clean (3887, only the known worktree path artifacts). The merge was the one you predicted: `tests/offline_guard.py` conflicted with the #255/#257/#258 stack in `install()` and `uninstall()`, and I kept both sides — the network guard goes in first on every install path, before the scripting stub and the child site, and comes out first on uninstall; the return value carries your `network_installed` when `src.server` is absent. All 79 guard-related tests pass together. (On first attempt I took the guard file whole from your side, which would have dropped the stack's work; a symbol grep caught it before anything was pushed, and it was redone as a proper three-way merge.)

Your code is unchanged. #260 and #261 follow.

PR #260, opened 2026-09-19
fix(tests): run the installer's live probe only with RESOLVE_VERIFY=1

denvital:
> **Stacked on #258** (which is stacked on #257 → #255). The two commits new here are 1852607 (fix) and dc54e57 (release v4.8.11); everything below them is #255, #257 and #258. Merge those first; this diff then shrinks to those two. v4.8.10 is #259.

## Summary

The follow-up #258 named and left open. `test_scripting_lib_discovery.test_the_live_probe_agrees_with_the_summary` runs the real `install.verify_resolve_connection` whenever Resolve is installed:

- The probe's child sets `PYTHONPATH` to Blackmagic's Modules directory. So neither #258's child site nor a `PYTHONPATH` tripwire loads in it.
- The child imports the real module and calls `scriptapp("Resolve")`, `GetProductName` and `GetVersionString`.
- With Resolve open, every run of the offline suite connected to it, read-only. In the four full runs behind #258 the test passed rather than skipped. It skips when the probe doesn't answer, so each of those runs called `scriptapp("Resolve")` on the real module.

## Change

Test-only. No tool, action or runtime code changed.

- The live test now runs only with **`RESOLVE_VERIFY=1`**, the switch the live harnesses already use. The check happens when the test runs instead of in a `skipUnless` at import, so it is testable. The installed-Resolve check still applies after it.
- `docs/process/release-process.md` names the command for changes to the installer's verification or summary. The automatic coverage on machines with Resolve is gone, so the release process has to ask for it.
- #258's guard docstring no longer says nothing stops this probe.

## Validation

- New `test_the_live_test_runs_only_when_asked` runs the live test with the host made to look installed, discovery pinned and the probe booby-trapped:
  - With `RESOLVE_VERIFY=1`, the trap goes off once, which shows it is armed.
  - Without it, the test skips, naming the switch, and never reaches the trap.
  - With the opt-in check removed, the new test fails.
- Full suite, `discover -s tests -t .` and `discover -s tests`: 3,771 tests each, 85 skipped (84 of 3,770 on #258; the new skip is the live test). The 11 errors are the same environment ones as on v4.8.8. Release static checks pass.
- The validation tripwire now also refuses any child that names Blackmagic's module and points at the real install on its command line or `PYTHONPATH`, the shape of the installer and doctor probes. It recorded nothing.
- pytest is not installed locally; CI runs it.
- No Resolve behavior changed, so no live Resolve run is required. None was made.


samuelgursky:
Merged and released as [v4.8.11](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.11). Thank you — this is the follow-up #258 was honest enough to name, and the fix is the right trade: a live probe that ran automatically on any machine with Resolve installed was coverage nobody could rely on, because it depended on what happened to be open; an opt-in that the release process names is coverage someone actually runs. Reading the switch at test time so the opt-in itself has a regression test is the detail I would have asked for.

Checked before merging: your branch was run against `main` in a scratch worktree, and the merged `main` after resolution ran the full suite clean (3888, two expected skips, only the known worktree path artifacts). Only the release surfaces and the changelog conflicted with #259's v4.8.10 this time; the guard file merged on its own. Your code is unchanged. The release-process addition is in, with the exact command.

#261 is last.

PR #261, opened 2026-09-19
fix(tests): keep the offline suite's traces, reports and update state out of logs/

denvital:
> **Based on `main`, independent of #255–#260.** Released as v4.8.12 because 4.8.5–4.8.11 are claimed by #255–#260. Merging after #259 conflicts textually in `tests/offline_guard.py`: both add a step to `install()` and `uninstall()`. Keep both. The same goes for `CHANGELOG.md` and the version files after any of the others.

## Summary

The offline suite wrote synthetic records into the operator's `logs/`.

- `logs/execution-traces.jsonl`, `logs/execution-reports/` and `logs/update-check.json` are the server's defaults.
- `tests/offline_guard.py` did not move them the way it already moves the audit log, the operation log and the media-analysis preferences.

Measured with `python -m unittest discover -s tests -t .` from a checkout with an empty `logs/`, one run:

- appended **1,865 tool-call records (672 KB)** to the trace log
- wrote an audit report for a synthetic execution
- created `update-check.json`

In a main checkout, the trace records share a file with the real ones, and that file is where `list_recent_executions` sends a reviewer.

## Change

**The guard redirects the run state** (`tests/offline_guard.py`):

- `RESOLVE_MCP_TRACE_FILE`, `RESOLVE_MCP_TRACE_REPORT_DIR` and `DAVINCI_RESOLVE_MCP_UPDATE_STATE` point at a temp directory for the run. A value the caller already set is left alone, and `uninstall()` restores the environment.
- It sets environment variables rather than swapping functions, for three reasons:
  - the server already reads these variables
  - `src.server` binds its own `update_state_path`
  - a child process inherits the environment

**A test that removed the redirect** (`tests/test_execution_trace.py`): two classes unset these variables with a bare `os.environ.pop`, which would have removed the redirect for every later test. Both now scope the change with `mock.patch.dict`.

## Tests

- New `tests/test_repo_logs_isolation.py` is the tripwire. It fails if:
  - any of the three variables is unset or points into the repo's `logs/`
  - the server resolves a path other than the one the guard set
  - a traced tool call or its exported report reaches the real `logs/`

  It looks for a unique execution id, so a live server appending to the same log cannot make it pass or fail.
- Against the previous guard, all three new tests fail: 4 failures (one per variable, plus the traced call), and 1 error because the old guard has no `RUN_STATE_ENVS`. Re-checked for this PR by restoring `main`'s `tests/offline_guard.py` in a scratch clone. With the previous `test_execution_trace.py` run first in the same process, the tripwire fails on the two variables that file popped.
- **Full suite**, `python -m unittest discover -s tests -t .`, with an empty `logs/`: before the change it left 1,865 trace lines, one report and `update-check.json`; afterwards `logs/` holds only `.gitkeep`.
- **Re-run for this release**, behind a stricter tripwire: 3,741 tests, 85 skipped, 14 errors, the same set as v4.8.4 under that tripwire, all environmental:
  - 6 need `numpy` or `requests`, which are missing from the venv
  - 5 in `test_offline_fallback` need `jszip`, since there is no `node_modules`
  - 3 are the tripwire's own refusals: two `test_doctor_paths` probe children, and the granular launch #255 addresses

  That tripwire answers `DaVinciResolveScript` with a stub. It refuses Resolve launches, off-machine network access, and any child whose command line imports the scripting module.
- Release checks are clean: `test_import` (lockfile sync), `audit_api_parity`, `gen_api_limitations --check`, `audit_readwrite_symmetry --check`, `agent-rules/generate.mjs --check`, the drift guards, `npm pack --dry-run`, `git diff --check`. `test_static_undefined_names` skipped: pyflakes is not installed locally.

## Correction to this change's original validation

The first before/after runs used a tripwire that did not refuse `install.verify_resolve_connection`'s probe child. That child replaces `PYTHONPATH` with Resolve's `Modules` directory, so it never loaded the tripwire.

In both runs `test_the_live_probe_agrees_with_the_summary` passed instead of skipping (84 skips instead of 85). The child therefore called `scriptapp("Resolve")` on the Resolve that was open, a read-only probe, once per run. The CHANGELOG entry says so, rather than claiming no live Resolve was reached. This change does not touch that path; #260 puts the test behind `RESOLVE_VERIFY=1`.

## Not covered

- **pytest was not run locally** (not installed here). `conftest.py` calls the same idempotent `install()`/`uninstall()`, and that lifecycle was checked directly.
- **The outbound GitHub request** from `install.py`'s update check still happens on this branch; its state now lands in the temp directory. #259 refuses the request itself.

No Resolve behavior changed, so no live Resolve run is required. None was made for this release; the earlier probe connection is described above.


samuelgursky:
Merged and released as [v4.8.12](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.12). Thank you — and that closes the set: 1,865 synthetic records per run into the file a reviewer is sent to read is the kind of pollution that quietly erodes trust in the audit trail, and redirecting by environment so children inherit it is the same reasoning #258 used, applied consistently.

Checked before merging: your branch was run against `main` in a scratch worktree, the merged `main` after resolution ran the full suite clean (3891, two expected skips, only the known worktree path artifacts), and the repo's trace log had the same line count before and after that full run — the direct proof of the fix. The merge was the one you predicted: `tests/offline_guard.py` conflicted with the #255/#257/#258/#259 stack in `install()` and `uninstall()`, and I kept both sides — your `_redirect_run_state()` goes in after the bridge redirect and `_restore_run_state()` comes out right after it, before the stub and child-site teardown. All 128 guard-related tests pass together. Your code is unchanged.

Seven releases from you today, v4.8.5 through v4.8.12 with v4.8.6 folded into v4.8.7, each with a measurement and a test that fails on the old code. The offline suite is now something I can trust with Resolve open, which it was not a week ago.

PR #262, opened 2026-09-19
fix(timeline): read include_linked="false" as false

Dev-next-gen:
## Summary

`_normalize_include_linked` handled the boolean `False` but not the string `"false"` (nor `"no"`, `"0"`, `"off"`, `"none"`). A caller sending `include_linked="false"` got a set containing the literal string `"false"` instead of an empty set, which made `bool(include_types)` true and defaulted the `relink` flag to `True` — the opposite of what the caller asked for.

The same "read string false as false" pattern was fixed on six permission flags (1089182), on `overwrite` (9618f75), on `ripple` (84a8a49), on `allow_partial_item_delete` (6bf65bd) and on `allow_non_mcp_name` (eaa6e94). This is its twin in the linked-items normalizer.

Before:

```
include_linked="false" → {"false"}  (bug — bool({"false"}) is True)
include_linked="no"    → {"no"}
include_linked="0"     → {"0"}
```

After:

```
include_linked="false" → set()      (correct — bool(set()) is False)
include_linked="no"    → set()
include_linked="0"     → set()
```

Existing behavior (True, False, "all", "audio", "video,audio", lists) is unchanged.

## Test plan

- [x] `test_normalize_include_linked_string_false_returns_empty` — 7 subtests covering every spelling ("false", "False", "FALSE", "no", "0", "off", "none"); all 7 fail on the previous code, all 7 pass after
- [x] The existing 29 tests in the file pass unchanged


samuelgursky:
Merged as-is and released as [v4.8.13](https://github.com/samuelgursky/davinci-resolve-mcp/releases/tag/v4.8.13). Thank you — the one normaliser in the tree that parses its own strings was exactly where the earlier `coerce_bool` sweep could not reach, and `bool({"false"})` being true is the kind of thing that only a pipeline looking for the pattern rather than the helper would find.

Checked before merging: `main` merged into your branch in a scratch worktree ran the full suite clean (3892, only the known worktree path artifacts). Nothing changed on landing beyond the changelog entry.

For your list, since this class is now closed on every destructive and opt-in read: the bare reads that remain in `src/server.py` are `include_*` flags that shape a read rather than gate a write (`include_flags`, `include_clip_color`, `include_markers`, `include_clip_properties`, `include_third_party`, and a few in media analysis behind their own `_media_analysis_bool`). A stringified `"false"` there widens a report instead of destroying anything, so they are lower stakes; if the pipeline wants them, one PR for the family is fine.

PR #263, opened 2026-09-19
feat(codex): migrate project defaults to GPT-6 Astra

panda8413:
## Summary

- default trusted Codex project sessions to `gpt-6-astra` with `high` reasoning
- default generic subagents to `gpt-6-astra` with `medium` reasoning
- pin cut and grade-match reviewers to `high`, and drift guard to `low`
- document trust, restart, override precedence, unavailable-model behavior, and rollback
- add parity tests for project, specialist, Claude, and documentation contracts

## Verification

- `python -m unittest tests.test_portable_asset_parity -v` — 10/10 passed
- `git diff --check 4a643b65..HEAD` — passed
- expanded drift suite — 26/27 passed; the sole failure is the pre-existing set of 32 stale generated agent-rule files, unchanged by this PR

## Follow-up

- confirm effective parent and subagent model selection after reopening the repository as a trusted Codex session


samuelgursky:
Thank you for the PR, but I am closing it without merging.

Which model, and at what reasoning effort, every Codex session opened in this repository should default to is a maintainer decision, and this project deliberately pins nothing for Codex today: the Codex specialist files carry instructions and tools only, and the Claude adapters carry their own pins for their own reasons. A repo-wide default, plus developer instructions applied to everyone's sessions, would need a reason grounded in this project — a measured problem it solves here — and the PR does not give one.

Two smaller things, for completeness. The verification section says the drift suite fails on 32 stale generated agent-rule files that predate this change; that guard passes on `main` and has on every release this week, so I could not reproduce that claim. And I have no way to verify the model id from here; the PR's own docs note that Codex fails rather than falls back when the account cannot use it, which would make a wrong id break every trusted session on the repo.

If a Codex default is ever wanted here it will be a small, deliberate change made with a model that has been verified against this workflow. Contributions that fix a measured behaviour of the server itself are very welcome — the recent history of this repo shows what those look like.

